PR #3, opened 2018-09-17
Fixes bug - datasets.load was not implemented

piyush-kgp:
The function is still unimplemented but at least the required files are there in the library and raises a decent NotImplementedError rather than the ugly 'attribute not found'
Suggestions welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,13 @@ _pycache__/
 /*.egg-info
 .eggs/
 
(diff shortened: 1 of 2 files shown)


rsepassi:
Hi @piyush-kgp, as the README states, tensorflow-datasets is not yet functional and is under active development. Happy to have further contributions once we have some of the initial stuff out.

PR #14, opened 2019-01-14
Fix docs for ImageLabelFolder

RicCu:
The example provided in the docstring for ImageLabelFolder doesn't work as-is. I may have overthought it, but I spent some time trying to fix a (non-existent) bug.


Conchylicultor:
Sorry about that. Thanks for your contribution. The doc will be re-generated soon to take this change into account.

PR #28, opened 2019-01-28
Add MovingMNIST dataset

jackd:
`moving_sequence` doesn't actually provide a dataset per-se, but it's obviously strongly linked to `moving_mnist`. Not sure if that's precisely what this repo is intended for, but had fun writing it (and watching bouncing shoes/coats from `fashion_mnist` was interesting).


tensorflow_datasets/video/moving_mnist.py
```diff
@@ -0,0 +1,95 @@
...
+  VERSION = tfds.core.Version("0.1.0")
+
+  def _info(self):
+    shape = (_SEQUENCE_LENGTH,) + _OUT_RESOLUTION + (1,)
+
+    # as Image - doesn't work with 1 as final dim?
+    # sequence = tfds.features.Image(shape=shape)
+
+    # as video - doesn't work with 1 as final dim?
+    # sequence = tfds.features.Video(shape=shape)
```
Conchylicultor: Video should works.
tfds.features.Video(shape=(seq_lenqth, height, width, 1))

If not, this is a bug from our end. Which error are you seeing ?

tensorflow_datasets/video/__init__.py
```diff
@@ -18,3 +18,4 @@
 from tensorflow_datasets.video.bair_robot_pushing import BairRobotPushingSmall
 from tensorflow_datasets.video.starcraft import StarcraftVideo
 from tensorflow_datasets.video.starcraft import StarcraftVideoConfig
+from tensorflow_datasets.video.moving_mnist import MovingMnist
```
rsepassi: please keep alphabetic order

tensorflow_datasets/video/moving_mnist.py
```diff
@@ -0,0 +1,95 @@
...
+  def _generate_examples(self, data_path):
+    """Generate MOVING_MNIST sequences as a single.
+
+    Args:
+      data_path (str): Path to the data file
+
+    Returns:
+      10000 x 20 x 64 x 64 x 1 uint8 numpy array
+    """
+    with tf.io.gfile.GFile(data_path, "r") as fp:
```
rsepassi: "rb"

jackd:
Well that was a fun rabbit hole to dive down. Found/[fixed](https://github.com/numpy/numpy/pull/12873) a bug in numpy related to `squeeze` on lists - it was ignoring the `axis` argument and being a little overzealous and squeezing out the final rank of the `20, 64, 64, 1` sequence, resulting in shape errors down the line.

That being said, I feel the fixed implementation of `squeeze` should make the existing implementation of `np_to_list` (this repo) raise an error. As far as I can tell, the old implementation of `np_to_list` in this repo and bugged `squeeze` was effectively the identity function. Fixed `np_to_list` here, but maybe this isn't the best place to issue such a change?

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -61,6 +62,8 @@ def __init__(self, shape):
       raise ValueError('Video shape should be of rank 4')
     if shape.count(None) > 1:
       raise ValueError('Video shape cannot have more than 1 unknown dim')
+    if shape[-1] not in (1, 3):
```
Conchylicultor: Please remove. The test is already in Image() called bellow: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/features/image_feature.py#L108

tensorflow_datasets/video/moving_mnist.py
```diff
@@ -0,0 +1,80 @@
...
+          "Moving variant of MNIST database of handwritten digits. This is the "
+          "data used by the authors for reporting model performance. See "
+          "`tfds.video.moving_sequence` for functions to generate training/"
+          "validation data."),
+        features=tfds.features.FeaturesDict(
+          dict(image_sequence=tfds.features.Video(
+              shape=(_SEQUENCE_LENGTH,) + _OUT_RESOLUTION + (1,)))),
+        urls=["http://www.cs.toronto.edu/~nitish/unsupervised_video/"],
+        citation=_citation,
+        splits=[tfds.Split.TEST]
```
Conchylicultor: You shouldn't put the `splits=[tfds.Split.TEST]` as this will automatically be pre-computed from the `_split_generators` function.

tensorflow_datasets/core/features/sequence_feature.py
```diff
@@ -260,7 +260,7 @@ def np_to_list(elem):
     return elem
   elif isinstance(elem, np.ndarray):
     elem = np.split(elem, elem.shape[0])
-    elem = np.squeeze(elem, axis=0)
+    elem = [np.squeeze(e, axis=0) for e in elem]
```
Conchylicultor: Nice catch. Thanks for fixing this.

tensorflow_datasets/core/features/video_feature.py
jackd: done, but does this mean videos with 0 or 2 channels are also accepted? In the interest of keeping documentation up to date (had me confused initially when it said channels had to be 3).

jackd:
The `moving_mnist` dataset is the testing dataset used/provided by the authors. They also provided code to dynamically produce their training/validation data. `moving_sequence` is a port of that to tensorflow with some generalizations (customizable number of images, speed, base dataset, output size) and a simplified bouncing mechanism (visually indistinguishable - original code clipped movement passed boundary to the boundary before reversing direction, this implementation reflects is directly resulting in a much simpler implementation that doesn't involve looping over all time steps. Happy to ellaborate more).

In the interest of reproducibility I think it's appropriate to package them together. Anyone wanting to test on this dataset will, presumably, want a similar dataset to train with, and accessing it from the same source as the test data makes sense to me. How one packages it is another question, and I'm open to suggestions.

My original implementation of `moving_mnist` overrode `as_dataset` for `train`/`validation` splits to return a `moving_sequence` mapped `mnist` dataset. This makes the interface uniform with other datasets, but it also raised other questions.

* What's an epoch?
* Should the base mnist dataset be shuffled? If so, how do we discourage users from shuffling the output dataset?
* Should both `train` and `test` sets from `mnist` be used in the `train` moving_mnist dataset?

Someone could just make those decisions (or `moving_mnist` could just have a very large set of `builder_kwargs` for the train/val case) but I feel if users already understand the `Dataset` interface it would be easier to give them the lower level tools and examples on how to use them. We could always do both.

A lot of these questions above are related to dynamically generated datasets in general (or those involving non-trivial mapping operations at least). Does `tfds` have a policy on these? Most of the infrastructure seems tailored to datasets on file, but in terms of my understanding of the projects goals and interface I'm not sure this is the limitation of its scope.

tensorflow_datasets/video/moving_mnist.py
```diff
@@ -0,0 +1,79 @@
...
+
+  VERSION = tfds.core.Version("0.1.0")
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=(
+          "Moving variant of MNIST database of handwritten digits. This is the "
+          "data used by the authors for reporting model performance. See "
+          "`tfds.video.moving_sequence` for functions to generate training/"
```
rsepassi: Let’s make this `tfds.video.moving_mnist.moving_sequence` for now so that it’s clear that it’s closely related to this dataset. You can keep the other file for the definition, but add the import for that function in this file. I’ll later work on exposing things properly and getting them to show up in the documentation.

tensorflow_datasets/video/moving_mnist.py
```diff
@@ -0,0 +1,79 @@
...
+          "validation data."),
```
rsepassi: validation data from the MNIST dataset.

tensorflow_datasets/video/moving_mnist.py
```diff
@@ -0,0 +1,79 @@
...
+        urls=["http://www.cs.toronto.edu/~nitish/unsupervised_video/"],
+        citation=_citation
+    )
+
+  def _split_generators(self, dl_manager):
+    data_path = dl_manager.download(
+        "http://www.cs.toronto.edu/~nitish/unsupervised_video/"
+        "mnist_test_seq.npy")
+
+    # authors only provide test data. See `tfds.video.moving_sequence` for
```
rsepassi: Update name

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
...
+  sequence = create_merged_moving_sequence(
+    images, sequence_pad_lefts, total_padding, **kwargs)
+  return MovingSequence(
+    image_sequence=sequence,
+    trajectories=trajectories,
+    start_positions=x0,
+    velocities=velocity)
+
+
+def as_moving_sequence_dataset(
```
rsepassi: Let’s rm this function and instead have users directly use `images_to_moving_sequence` in their own map functions.

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
...
+  if pad_lefts.shape.ndims != 2:
+    raise ValueError("`sequence_pad_lefts` must be a rank 2 tensor")
+  if len(total_padding) != 2:
+    raise ValueError(
+      "`total_padding` must have 2 entres, got %s"
+      % str(total_padding.as_list()))
+  seq = _create_moving_sequence(
+      image, pad_lefts, tf.convert_to_tensor(total_padding))
+  ph, pw = total_padding
+  h, w, n_channels = image.shape
```
rsepassi: Does this fn only work if the image shape is fully known statically?

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
...
+    [sequence_length, out_h, out_w, n_channels] overlayed padded sequence.
+  """
+  if isinstance(merge_fn, six.string_types):
+      merge_fn = _merge_fns[merge_fn]
+  images = tf.convert_to_tensor(images, dtype=tf.uint8)
+  sequence_pad_lefts = tf.convert_to_tensor(sequence_pad_lefts, dtype=tf.int32)
+  if images.shape.ndims != 4:
+    raise ValueError("`images` must be a rank 4 tensor")
+  if sequence_pad_lefts.shape.ndims != 3:
+    raise ValueError("`sequence_pad_lefts` must be a rank 4 tensor")
```
rsepassi: rank 3 tensor

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
...
+  if len(total_padding) != 2:
+    raise ValueError(
+      "`total_padding` must be len 2, got %s"
+      % str(total_padding.as_list()))
+
+  image_res = [i + t for i, t in zip(images.shape[1:3], total_padding)]
```
rsepassi: Image shape required to be fully known statically? If so, check and raise an error up front. Or make it so that it can be dynamic.

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
+"""
```
rsepassi: Let’s move this documentation into images_as_moving_sequence. Use MNIST instead of FashionMNIST as the example.

For the docstring here, you can put something simpler:

```
“””Turn simple static images into sequences of the originals bouncing around.

Adapted from <url to MovingMNIST page>.
”””
```

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
...
+  points = points % 2
+  return tf.math.minimum(2 - points, points)
+
+
+MovingSequence = collections.namedtuple(
+  "MovingSequence",
+  ["image_sequence", "trajectories", "start_positions", "velocities"])
+
+
+def images_to_moving_sequence(
```
rsepassi: I’m thinking that this should be the only publicly exposed function and the others should remain private for now. What do you think?

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
...
+    images, sequence_length=20, speed=0.1, total_padding=(36, 36),
+    **kwargs):
```
rsepassi: Since this is the main function, let’s not have kwargs and instead list out all the arguments.

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
...
+    images: [?, in_h, in_w, n_channels] uint8 tensor of images.
+    sequence_length: int, length of sequence.
+    speed: float, length of each step. Scalar, or rank 1 tensor with length
+        the same as images.shape[0].
+    total_padding: (pad_y, pad_x) total padding to be applied in each dimension.
+    kwargs: passed to `create_merged_moving_sequence`
+
+  Returns:
+    `MovingSequence` namedtuple containing:
+        `image_sequence`:
```
rsepassi: Could you please make these descriptions of the return values a bit more descriptive?

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,419 @@
+"""
+Contains functions for creating moving sequences of smaller bouncing images.
+
```
rsepassi: Could you please add a test for images_as_moving_sequence?

rsepassi:
Thank you for explaining!

I think you made the right call to only include the test data and to include this function here so that users can create the moving sequences themselves from the MNIST dataset.

My main request is that we limit the surface area for the new module to just the 1 key method and to add a test for `images_as_moving_sequence` on some dummy data and use the `test_utils.run_in_graph_and_eager_modes` decorator.

jackd:
No disagreement here. I've added a few more optional kwargs to the main function in place of other publicly visible functions and adjusted for dynamic sizing, but the implementation remains very much the same.

jackd:
Finally, an excuse to play with tf 2.0 :). Merged in master changes and updated to single-image implementation.

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,216 @@
...
+    velocity = _get_random_unit_vector(ndims, tf.float32) * velocity
+  elif velocity.shape.ndims != 1:
+    raise ValueError("velocity must be rank 0 or rank 1, got %s" % velocity)
+  t = tf.range(sequence_length, dtype=tf.float32)
+  trajectory = _get_linear_trajectory(start_position, velocity, t)
+  trajectory = _bounce_to_bbox(trajectory)
+
+  total_padding = output_size - image_shape[:2]
+
+  # cond = tf.assert_greater(total_padding, -1)
```
rsepassi: clean up?

tensorflow_datasets/video/moving_sequence_test.py
```diff
@@ -0,0 +1,44 @@
+"""Tests for moving_sequence."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+import tensorflow as tf
+from tensorflow_datasets.core import test_utils
+import tensorflow_datasets.video.moving_sequence as ms
+
```
rsepassi: Add here:
tf.compat.v1.enable_eager_execution()

tensorflow_datasets/video/moving_sequence_test.py
```diff
@@ -0,0 +1,44 @@
...
+    velocity = tf.constant([vh, vw], dtype=tf.float32)
+    out_size = (h + sequence_length, w + sequence_length)
+    start_position = tf.constant([0, 0], dtype=tf.float32)
+
+    sequence = ms.image_as_moving_sequence(
+        image, start_position=start_position, velocity=velocity,
+        output_size=out_size, sequence_length=sequence_length)
+    sequence = tf.cast(sequence.image_sequence, tf.float32)
+
+    self.assertAllEqual(
```
rsepassi: use `self.evaluate` to actually evaluate these Tensors into values in a way that's graph/eager compatible.
Or you can use `as_numpy`

tensorflow_datasets/video/moving_sequence_test.py
```diff
@@ -0,0 +1,44 @@
...
+        tf.reduce_sum(sequence, axis=(1, 2, 3)),
+        tf.fill((sequence_length,), tf.reduce_sum(tf.cast(image, tf.float32))))
+
+    for i, full_image in enumerate(tf.unstack(sequence, axis=0)):
+      j = i // 2
+      subimage = full_image[i:i+h, j:j+w]
+      n_true = tf.reduce_sum(subimage)
+      # allow for pixel rounding errors in each dimension
+      self.assertAllEqual(n_true >= (h-1)*(w-1), True)
```
rsepassi: same here, this has to be evaluated to work in both graph/eager modes

tensorflow_datasets/video/moving_sequence.py
jackd: Can't get this assert working with `tf.assert_greater` in 2.0 - in graph mode it returns `None`. Temporarily fixed using `tf.compat.v1.assert_greater`, but that's a little unsatisfactory...

tensorflow_datasets/video/moving_sequence.py
```diff
@@ -0,0 +1,218 @@
...
+    points: float array
+
+  Returns:
+    tensor with same shape/dtype but values in [0, 1].
+  """
+  points = points % 2
+  return tf.math.minimum(2 - points, points)
+
+
+def _get_random_unit_vector(ndims=2, dtype=tf.float32):
```
MechCoder: The current code generates the direction from a normal distribution. I believe this is not what the original code does which generates the direction uniformly.

Please see:

line 245 in data_handlers.py in http://www.cs.toronto.edu/~nitish/unsupervised_video/unsup_video_lstm.tar.gz
where the direction is sampled uniformly from 0 - 2*pi

Is that correct or am I missing something?

jackd: Glad to see someone's actually reading the code! The original takes a unit vector with *angle* sampled uniformly. Here we take random normal *coordinates* and normalize them. The resulting distributions are equivalent. See e.g. [alternative method 1](http://corysimon.github.io/articles/uniformdistn-on-sphere/) or convince yourself with the below code.

```python
import numpy as np
import matplotlib.pyplot as plt

n = 100000
x, y = np.random.normal(size=(2, n))
angle = np.arctan2(y, x)
plt.hist(angle)
plt.show()
```

MechCoder: I see, cool thanks!

PR #32, opened 2019-02-01
Run length encoding + utils for shapenet voxel dataset

jackd:
This ended up being way bigger than I initially planned. Apologies.

The main motivation was to make a common subset of shapenet available that looks at single/multi view object reconstruction (`shapenet_r2n2`). The voxel data is available in binvox files which use integer run length encoding (RLE), but the binary version of RLE (BRLE) is more efficient for binary data like voxels. While `shapenet_r2n2` only requires the `BrleFeature`, I figured I might as well write an `RleFeature` as well.

Note (B)RLE could also be used to efficiently store image segmentation data. I haven't profiled it or compared compression rates, and `rle_to_dense` will need some work before that's competitive in terms of speed (currently uses `tf.py_function`, but that's just to replicate `np.repeat`) - but use cases aren't necessarily restricted to voxels.

I'm not sure if I'm going completely the wrong way with `ShapedFeature`s and `SkipEncodingFeature`s, but they were the result after I abstracted the behaviour I wanted from creating `shapenet_r2n2`. Happy to heavily refactor if I'm missing something obvious.


tensorflow_datasets/core/features/skip_encoding_feature.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow_datasets as tfds
+from tensorflow_datasets.core.features import feature
+
+
+class SkipEncodingFeature(feature.FeatureConnector):
```
rsepassi: This is an interesting approach to skipping encoding. Is it possible to instead check in the base feature whether the data looks encoded? For example, in the `Text` `FeatureConnector`, if the user passes in a list of integers, it could simply skip encoding by checking the input shape/type.

rsepassi: Also, it's important for documentation to be correct and so if we stick with this approach, we'll want to update how [`FeatureConnector`s display](https://github.com/tensorflow/datasets/blob/d2123d21ec68607a0de830ab3afcde36162190f0/tensorflow_datasets/core/features/feature.py#L283) so that `SkipEncodingFeature` could forward the base connector and not show up in documentation.

tensorflow_datasets/core/features/shaped_feature.py
```diff
@@ -0,0 +1,190 @@
...
+      raise ValueError(
+          "%s must be a `FeatureConnector` but got %s"
+          % (name, connector))
+
+
+def _num_unknown_dims(shape):
+  return shape.count(None)
+
+
+class DynamicShapedTensor(feature.FeaturesDict):
```
rsepassi: Are these 2 new connectors maybe better to add to the `Tensor` `FeatureConnector`. There's a [todo](https://github.com/tensorflow/datasets/blob/d2123d21ec68607a0de830ab3afcde36162190f0/tensorflow_datasets/core/features/feature.py#L566) to support dynamic shapes.

If the `shape` passed to the `Tensor` constructor was not fully known, then you could write out to disk the feature value as well as the shape and reshape on the way out in decode.

Also note that you should not subclass `FeaturesDict` here. `FeatureConnector`s that are logically just a single features can still write multiple features to disk. `get_serialized_info` would reflect the fact that on disk there are multiple fields connected to this feature.

tensorflow_datasets/core/features/rle_feature/rle_feature_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+import functools
+import numpy as np
+import tensorflow as tf
+from tensorflow_datasets.testing import test_utils
+import tensorflow_datasets.public_api as tfds
+from tensorflow_datasets.core.features.rle_feature.rle import tf_impl
+tf.compat.v1.enable_eager_execution()
+
+b = functools.partial(np.array, dtype=np.bool)
```
rsepassi: See other comment about single-letter names

tensorflow_datasets/core/features/rle_feature/rle_feature.py
```diff
@@ -0,0 +1,54 @@
...
+  def decode_example(self, tfexample_data):
+    return tf_impl.rle_to_dense(tfexample_data)
+
+  def encode_example(self, example_data):
+    return np_impl.dense_to_rle(
+      example_data, dtype=self._encoded_dtype.as_numpy_dtype)
+
+
+class BrleFeature(RleFeatureBase):
+  """`FeatureConnector` for brinary run length encoded 1D tensors."""
```
rsepassi: spelling

tensorflow_datasets/core/features/rle_feature/rle_feature.py
```diff
@@ -0,0 +1,54 @@
...
+  def __init__(self):
+    tf_impl.ensure_supported_version('`%s`' % self.__class__.__name__)
+  _encoded_dtype = tf.int64
+
+  def get_serialized_info(self):
+    return feature.to_serialized_field(
+        tfds.features.TensorInfo(shape=(None,), dtype=self._encoded_dtype))
+
+
+class RleFeature(RleFeatureBase):
```
rsepassi: let's have the names be longer and more clear. RunLengthEncodedFeature

tensorflow_datasets/core/features/rle_feature/rle/tf_impl.py
```diff
@@ -0,0 +1,101 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow as tf
+
+major, minor = (int(v) for v in tf.__version__.split('.')[:2])
+is_over_1_12 = major >= 1 and minor > 12
+
+
+def ensure_supported_version(error_identifier):
```
rsepassi: Let's move this logic to `tf_compat.py` and add a function that can be called as `tf_compat.assert_tensorflow_version_gt("1.12.0", err_msg)`.

tensorflow_datasets/core/features/rle_feature/rle/tf_impl.py
```diff
@@ -0,0 +1,101 @@
...
+    true_vals = tf.fill([n_true], true_val)
+
+  ragged_false = tf.RaggedTensor.from_row_lengths(false_vals, enc_false)
+  ragged_true = tf.RaggedTensor.from_row_lengths(true_vals, enc_true)
+
+  ragged_all = tf.concat([ragged_false, ragged_true], axis=1)
+  return ragged_all.values
+
+
+def tf_repeat(values, counts):
```
rsepassi: This is possible to do in TF and we, for now, are not allowing `py_function` usage in the `tf.data` pipeline. This ensures that the input pipelines are high-performance and work everywhere.

tensorflow_datasets/core/features/rle_feature/rle/shared_tests.py
```diff
@@ -0,0 +1,200 @@
...
+
+Note these tests aren't run here directly. See `np_test` and `tf_test`.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import numpy as np
+
+
+class RleNpTest(object):
```
rsepassi: I don't see this used anywhere. Rm?

tensorflow_datasets/core/features/rle_feature/rle_feature_test.py
```diff
@@ -0,0 +1,70 @@
+"""Tests for tensorflow_datasets.core.features.rle_feature."""
```
rsepassi: For tests, please follow these [instructions](https://github.com/tensorflow/datasets/blob/master/CONTRIBUTING.md#tests) (some of which you've already done).

tensorflow_datasets/core/features/rle_feature/rle/np_impl.py
```diff
@@ -0,0 +1,276 @@
+"""Numpy encode/decode/utility implementations."""
```
rsepassi: What's the motivation to having both a NumPy implementation and a TF implementation? It seems that with Eager mode enabled, the usage of the TF version would be nearly the same.

tensorflow_datasets/core/features/skip_encoding_feature.py
jackd: My original implementation had this plus reshaping bundled into the base `FeatureConnector`s. Things just got a bit ugly, especially when allowing for potentially dynamic shapes. At one stage it could accept dense arrays or dicts with shape/encoding/encoding_type keys so the BRLE feature could encode directly from RLE data (that's the use case for `shapenet_r2n2` after all - everything else is just trying to optimize for reusability). Replicating that with the RLE feature connector (not used as-yet, but could be in the future?) and there's suddenly 4 classes (RLE/BRLE connector with static/dynamic shapes) that each need to support multiple different input formats.

I'd be worried about inferring the encoding from the shape/data type alone. What if someone accidentally passes in a dense array with `0`s and `1`s rather than `True`s and `False`s? What if they want to pass in `RLE` data and have it encoded as `BRLE`?

Presumably the creator of the feature connector knows the format they'll be feeding the data. At one stage I had the constructor accept an encoder in the constructor... but if they've created the encoder, why not just have them encode the data directly and tell the `FeatureConnector` not to worry?

I guess I don't love the way `FeatureConnector`s couple encoding with decoding. This was my way to optionally decouple those. As it stands I feel composition is preferable to convoluting an otherwise straight-forward `encode` method. If you can suggest a way of dealing with the shape issue more elegantly the complication would be less of an issue, but I still feel there would be a compelling argument for composition.

Happy to be over-ruled :)

jackd: Happy to make this change - possibly via a `DelegatingFeatureConnector` that redirects all base methods? I've done a lot of work in dart and really like the `DelegatingXXX` pattern... but I'm wary of throwing in too many micro-classes in a single request.

tensorflow_datasets/core/features/shaped_feature.py
jackd: This is what I've attempted to do here (originally there was even a `ShapedTensor` function that directed appropriately depending on the `shape`).

I'm happy to move away from `FeaturesDict` (didn't realize similar behaviour was supported by `FeatureConnector`s), but I'm unsure how to implement this directly with `Tensor`. I could subclass `Tensor` and redict to the super method purely for the reshape, but that feels awfully convoluted and conceptually wrong. (B)RLE encodings are intrinsically 1D, and my gut is much happier composing ND features out of an encoding feature and a shape feature than duplicating classes or writing code with lots of `if else` clauses.

tensorflow_datasets/core/features/rle_feature/rle/tf_impl.py
jackd: I suspected this might be the case (but given I had all the other pieces for an `RleFeature` I felt it was worth putting in with disclaimers in the README).

Can you direct me to a working implementation of `tf.repeat`? [This issue](https://github.com/tensorflow/tensorflow/issues/8246) is still open, and while there was a pull request with a custom kernel implementation it was abandoned (the reviewer seemed to think it could be implemented using existing ops... but I never worked out how). The implementation given in the issue doesn't seem to work when there are counts of 0 - this is unlikely to be an issue for `int64` encodings, but is certainly an issue with, say, `uint8` encodings (which aren't included here, so we might be able to get away with it... but I'd feel more comfortable making it more general).

tensorflow_datasets/core/features/rle_feature/rle/np_impl.py
jackd: ... ... ... I'm not actually sure. Historically I've always developed numpy versions then tried to optimize for tensorflow, and eager mode still isn't part of my standard work flow.

That said, how would one use `tf` implementations in the encoding phase and maintain compatibility with tensorflow 1.x? My understanding is eager mode has to be enabled at the beginning of a script and there's no guarantee when encoding will be run - or is there some way to turn it on and off dynamically?

Also, without an `np.repeat` tensorflow equivalent, there are some ops I'm not equipped to deal with in tensorflow (never written a custom op, and while I'd be keen to get stuck in and learn how this rabbit hole is deep enough as it is).

I can definitely appreciate that numpy encoding implementations and tf decoding implementations are sufficient for most use cases. Adding numpy decoding isn't much trouble (easier than tensorflow until eager mode becomes the norm) and given I already had it I thought I'd throw it in. I accept that's not the most compelling argument though :).

jackd:
I agree with everything I haven't commented on above (will get around to changes once I've got a clear vision of the greater picture).

A few general notes:

* For the use case I originally had in mind (`shapenet_r2n2`) there's a lot more in here than is strictly necessary. In particular, I could easily hard-code the shape/encoding and do away with the `ShapedFeature` and `SkipEncodingFeature` entirely. Most voxel stuff I do is on regular grids so that's unlikely to ever cause me issues. This would make things much simpler but at the cost of flexibility. I was mainly thinking about image masks in things like coco where each image is a different shape.
* Including RLE stuff (other than `rle_to_brle`) was largely an after-thought. I haven't done the analysis, but I suspect even multi-class segmentation tasks would be better encoded as per-class BRLEs, rather than a single RLE (though this might not be the case when there are a large number of classes), and normal image compression probably works pretty well in most cases anyway. I'm happy to strip this right back.
* A lot of the encoding functions are complicated by the possibility of exceeding the maximum integer value of the corresponding dtype. For voxel data, the binvox format seems quite popular in the research community which is `uint8` - so e.g. a string of 300 zeros would have to be BRL encoded as `255 0 45 0`. The `FeatureConnector`s here are limited to only encoding in `int64` though, and the likelihood of exceeding the corresponding maximum value is small. If we wanted to remove support for different dtypes this could be written more elegantly and run faster. Note decoding speed is not affected by this (though smaller dtypes may require longer decoding time under certain circumstances).
* Sorry about the lazy tests.

tensorflow_datasets/core/features/skip_encoding_feature.py
jackd: After some more thinking I've come to agree `SkipEncoding` isn't a great idea. Even if the input is an encoding, it should be checked for consistent lengths. I'll change this to accept either a fully encoded input of the relevant type or dense inputs.

I still feel a separate feature for transforming/possibly saving shape is appropriate...

jackd:
Summary of changes/open questions:

## SkipEncodingFeature:
Removed.

## RunLengthEncodedFeatures:
`tf_impl.rle_to_dense` currently relies on `tf.py_function`. I can see this isn't ideal, but without a `tf.repeat` or custom kernel I don't see an easy way around this other than removing it entirely (or just from imports). I've left it in for the moment but it's not used in `shapenet_r2n2` so taking it out isn't a big deal.

## ShapedTensor:
I can see there's a duplication between `ShapedTensor` and `Tensor` in the reshaping, and the problem is essentially the same (shape cannot be saved to disk). I'm happy to be directed towards the appropriate implementation, but I'd go with something like:
```python
class Tensor(ShapedTensorBase):
  def __init__(self, shape=None, dtype=None, base_feature=None):
    if base_feature is None:
      assert(shape is not None)
      n_none = shape.count(None)
      assert(n_none <= 1)
      base_feature = FlatTensor(
          dtype=dtype, size=np.prod(shape) if n_none == 0 else None)
    else:
      assert(dtype is None or dtype == base_feature.dtype)
    self._base_feature = base_feature
    self._shape = shape
  # rest like StaticShapedTensor
```
i.e. replace `Tensor` implementation with `StaticShapedTensor` and (for backward compatibility/convenience) allow a default `FlatTensor` (which would be almost identical to the current `Tensor` implementation but only accepts a `size`/`num_elements`). Dynamic shaped tensors could be handled similarly but in a separate class (much like the current `DynamicShapedTensor`).

I can't see that being a breaking change... but I haven't implemented it properly, so maybe there's something I'm not appreciating. I'm wary of touching too many different elements in the one PR... but without RLE features the motivation for changing `Tensor` might not be clear. Would it be easier if I opened a different PR for that?

## File structure:
Should `features/run_length_encoded_feature/rle` be moved to `util/rle`? Or `features/run_length_encoded_feature/impl`?

tensorflow_datasets/core/features/rle_feature/rle/tf_impl.py
rsepassi: So here's something that may work for this instance though doesn't capture the full generality of `np.repeat`. Does this work for you?

https://gist.github.com/rsepassi/d2367b4355828575b57fe86c28c3f1cf

tensorflow_datasets/core/features/rle_feature/rle/np_impl.py
rsepassi: I think if we can get the TF implementation to work, we can leave out the numpy version. One fewer thing to have around (maintenance, testing, etc.)

rsepassi:
Thanks for these updates and for summarizing the current status.

File structure: putting things in a `run_length_encoded/` subdirectory looks good to me.

Would it be possible to simply have rle and brle do the shape stuff they need internally instead of adding (Dynamic)ShapedTensor? If we find the logic is duplicated in various places, we can create an abstraction.

rsepassi:
(and sorry again for the review delay)

tensorflow_datasets/core/features/rle_feature/rle/tf_impl.py
jackd: ... well that's pretty obvious now you mention it. I've put something similar in. Would love to make it a `foldl` rather than a `while_loop`, but that's because I just don't feel comfortable with seeing `repeats[i]`. Not sure if [this](https://github.com/tensorflow/tensorflow/pull/26030) will get through but figured I'd give it a shot anyway.

A similar thing works with BRLE decoding, so I've put a check in there to switch between a `loop`y implementation or `RaggedTensor` version depending on `RaggedTensor` availability. Any thoughts on speed? That also removes any `tf_comp` issues.

jackd:
I've moved the reshaping into the base class... and abstracted ever-so-slightly to make that base class a usable substitute for `Tensor` feature which supports dynamic shapes (I've copied all the tests across and they're all passing for `ShapedTensor`). Obviously that means there's a lot of duplication there for the moment - but I'm a bit scared of replacing such a central piece...

rsepassi:
Hey @jackd is this ready for another look? Are all the comments addressed? GitHub's reviews don't make it easy to follow which threads are still open and which are not.

tensorflow_datasets/core/features/rle_feature/rle/np_impl.py
jackd: I've removed -almost- all duplication. Numpy implementations are used for encoding, tf for decoding. The only duplication is a couple of 1-liners (`rle_length` and `brle_length`).

Is there a simple way to using tf ops in the encoding stage? Once eager mode is default I can see it being straight forward, but until then I'm not sure how to handle it. I have very little experience with eager mode much in 1.x .

If there's an easy way around that and the `np` implementations are still an issue I can refactor them to `tf` variants, but I don't think there'll be much performance improvement if any (and that will be limited to the serialization stage), possibly at the cost of readability. If API stability is the main issue I could make the np versions private and make tf wrappers that can be optimized later if deemed necessary.

jackd:
Sorry about that. I've marked those that I've addressed without comment resolved (is that rude? Not sure on etiquette here. It seems odd that a single party in a conversation can decide when something's resolved. I intent marking resolved as "I've taken this change on board without disagreement and there's nothing really of note worth pointing out in my edit").

Current issues that need resolving:
* duplicated behaviour of `ShapedTensor` vs `Tensor` (see below)
* merge conflicts
* minor tidying (e.g. throwing in some `@api_utils.disallow_positional_args`, make sure code docs are up to date etc).

The main question I have right is one of API design regarding shape in tensors. If I can get a clear vision of what's desired on your end I'd be happy to resolve all of the above, but until then I wouldn't want to waste your time.

The current shape implementation (`run_length_encoded_feature.ShapedTensor`) duplicates much of `feature.Tensor`, with explicit `_get_flat_serialized_info`, `_encode_flat_values` and `_decode_flat_values`. The `ShapedTensor` implements those methods with versions straight from `Tensor`, where as the run-length-encoded versions override them with RLE implementations.

Keeping in mind implementing a `FeatureConnector` mostly means implementing 4 methods (`encode_example`, `decode_example`, `get_serialized_info` and `get_tensor_info`), I don't particularly like this design for 2 reasons:
1. 3 of the 4 overriden methods in `ShapedTensor` (all except `get_tensor_info`) have different code paths/return value structures based on the "staticness" of the shape. I feel separating this out into 2 classes would make code comprehension easier. It wouldn't be a breaking change (since `Tensor` currently doesn't support dynamic shapes), though it would require -slightly- more work from the user (e.g. choosing `Tensor` vs `DynamicTensor` manually when extending either. From a naming perspective ideally I would like to implement `StaticTensor` vs `DynamicTensor` as classes and make `Tensor` a factory that directs appropriately based on the provided shape... but that would break any other classes which have extended the current `Tensor` implementation.

To summarise, proposed solutions:
(a) one convoluted class with behaviour depending on constructor args (currently implemented)
(b) two different simple classes + factory constructor

2. The 3 methods required to implement a different flat encoding ( `_get_flat_serialized_info`, `_encode_flat_values` and `_decode_flat_values`) have direct analogues in the base `FeatureConnector` class. I feel composition is a better fit, i.e. make `Tensor.__init__` accept shape, dtype and flat_connector args, where flat_connector only worries about storing flat data and defaults to dense storage.

Solutions:
(a) implementations extend Tensor and override `_flat*` methods (currently implemented)
(b) `Tensor` accepts a `flat_connector` argument and implementations implement flat `FeatureConnector`.

The current implementation uses 1.(a) and 2.(a). If this is roughly what you want I can go ahead and clean things up right now. Personally I feel 1.(b) and 2.(b) would be the simplest to implement/test/maintain, but I'm not overly experienced when it comes to designing such things. Any combination is possible, though I feel 1.(b) and 2.(a) would be harder work to maintain/extend (I'd need to make `DynamicBRLE`, `StaticBRLE`, `DynamicRLE`, `StaticRLE` classes... not that they'd be overly complex, but I should probably write tests for all those variants etc). With the aid of a few factory constructors all options could have identical interfaces for creating/using objects, so the point of debate is purely class structure/ease of extension.

tensorflow_datasets/core/features/rle_feature/rle/np_impl.py
rsepassi: You could use TF in encode if you wanted to with:

```
with tf.Graph().as_default():
  tensor = ...TF ops...
  np_val = tfds.as_numpy(tensor)
```

If the NP and TF implementations here are actually disjoint, then it's fine to leave them in NP. If there's duplication, rm the NP stuff.

rsepassi:
I'd like @Conchylicultor to take a look at `ShapedTensor`.

I'm thrilled that you're contributing so much and hope to have you do much more, but in the future, let's definitely go with smaller PRs and more up-front discussion in an issue for additions/changes to core things.

tensorflow_datasets/core/features/rle_feature/rle/np_impl.py
jackd: ... mabe I'm misunderstanding, but this results in ~700x slower code on my fairly ordinary machine for a simple example (equivalent to `brle_length`).

```python
import numpy as np
import tensorflow as tf
import timeit

max_val = 1000
size = 1000


def f_tf():
    x = np.random.randint(max_val, size=size)
    with tf.Graph().as_default():
        return tfds.as_numpy(tf.reduce_sum(x))


def f_np():
    x = np.random.randint(max_val, size=size)
    return np.sum(x)


print(timeit.timeit(f_tf, number=1000))     # ~7 sec
print(timeit.timeit(f_np, number=1000))     # ~ 0.01 sec
````

That's using `tf 1.13`. That can be sped up a bit by reusing the same graph/session
```python
graph = tf.Graph()
with graph.as_default():
    x_pl = tf.placeholder(shape=(size,), dtype=tf.int64)
    out = tf.reduce_sum(x_pl)

with tf.Session(graph=graph) as sess:
    def f_sess():
        x = np.random.randint(max_val, size=size)
        return sess.run(out, feed_dict={x_pl: x})

    print(timeit.timeit(f_sess, number=1000))  # 1.5 sec
```

but that's still a heap slower, and requires additional infrastructure to check if `tf.executing_eagerly()`. Even using eager execution things don't get much better.
```python
import numpy as np
import tensorflow as tf
import timeit
tf.compat.v1.enable_eager_execution()

max_val = 1000
size = 1000

def f_tf_eager():
    x = np.random.randint(max_val, size=size)
    return tf.reduce_sum(x)

print(timeit.timeit(f_tf_eager, number=1000))  # ~1.2 sec
```

The story is largely the same for 2.0 preview.

Throwing some `with tf.device('/cpu:0)`s in there plus eager mode or graph/session reuse can get it down to about `0.25`, and some burn-in runs gets you to `0.13`... but we're still taking a 13x speed decrease (~65x when not in graph mode).

 That seems like a high price to pay just to avoid duplication of a couple of 1-line functions. If the only use by `tfds` is during the encoding stage (where the pipeline isn't generally dominated by tensorflow ops) should the `tf` versions be dropped to avoid the duplication? tf implementations _could_ be useful to users, though it's fairly unlikely. In these cases (where `brle_length` is literally just a `tf.reduce_sum` rather than a `np.sum` and `rle_length` isn't much trickier) users could probably work it out themselves, but I'm encountering other situations (not related to this PR) during encoding where numpy implementations are significantly different to tf implementations (mostly in cases with scatter updates or dynamic looping).

jackd:
@rsepassi I did some basic benchmarking on various versions of `tf_repeat` similar to what you suggested  above and in the end found them all to be incredibly slow - `50x` - `100x` slower than `tf.py_function` redirecting to `np.repeat`.

Unlike other performance issues discussed above relating to the encoding, `repeat` is used in the decoding (i.e. relevant to training/evaluation, not just preprocessing). I understand you want this to be able to run everywhere so native `tf` ops are preferable, but does the magnitude of the performance difference change this?

I've implemented my own custom kernel based on an abandoned tensorflow PR [here](https://github.com/jackd/tf-repeat)... but I'm guessing tensorflow datasets might not be so keen to maintain custom ops?

jackd:
dataset infos [here](https://gist.github.com/jackd/5c61a00694795205d4e66a41032e214a)

jackd:
Sorry to have left this pending so long. @rsepassi Can you shed any light on why the kokoro build is failing? I'm not familiar with the platform, but previous failures have been easy enough to view. With this last one I've after clicking through all the links I can and the most helpful information I've got it `Unable to retrieve invocation log`.

Regarding dynamic shaped tensors: should I open an issue to discuss this? There doesn't seem to have been much progress since I last checked in. I can remove the functionality if desired.

PR #46, opened 2019-02-08
Define number of images in cats_vs_dogs

rodrigob:
25k training, and 12.5k testing. Unclear if this release is train only or train and test. I am guessing it is train only.


rsepassi:
Thanks for contributing! In this case, you can see all this information on the [dataset documentation page](https://github.com/tensorflow/datasets/blob/master/docs/datasets.md#cats_vs_dogs).

Issue #49, opened 2019-02-12
[data request] Colorectal Histology (Collection of textures in colorectal cancer histology)


tarrade:
* Name of dataset:  Colorectal Histology (Collection of textures in colorectal cancer histology)
* URL of dataset: https://zenodo.org/record/53169#.XGL2CNFCfOR
* License of dataset: Open Access
* Short description of dataset and use case(s): 
The dataset serves as a much more interesting MNIST or CIFAR10 problem for biologists by focusing on histology tiles from patients with colorectal cancer. In particular, the data has 8 different classes of tissue (but Cancer/Not Cancer can also be an interesting problem).
https://www.kaggle.com/kmader/colorectal-histology-mnist/home
https://zenodo.org/record/53169#.XGL2CNFCfOR

Good example for people to see use case of AI for Good.

Can tried to help to do TFRecords files if needed (I am learning it)

Folks who would also like to see this dataset in `tensorflow/datasets`, please +1/thumbs-up so the developers can know which requests to prioritize.

cyfra:
Big thanks to @jackd for submitting this dataset.

PR #51, opened 2019-02-15
Added colorectal histology image builders

jackd:
As [requested](https://github.com/tensorflow/datasets/issues/49).

Possible issue: the data is made available as `tif` files and I'm using `PIL` to open them (and to save the dummy data for generation) - is that a problem? From what I understand tensorflow doesn't support `tif` out of the box and I'm not particularly enthusiastic about rolling my own just for this purpose...


jackd:
... any easy way around the `PIL` dependency? Are the datasets team prepared to allow optional requirements like this? I imagine most users would have PIL installed/be happy to install PIL -somewhere- in order to do the encoding (not necessarily the same machine as training)...

I'm happy to wrap the errors in a suitable test so they don't run anywhere without `PIL`, but that would seem to defeat the purpose of tests...

tensorflow_datasets/image/colorectal_histology.py
```diff
@@ -0,0 +1,169 @@
...
+            gen_kwargs=dict(
+                start_frac=0.9,
+                end_frac=1,
+                root_dir=folder
+            )
+        ),
+    ]
+
+  def _generate_examples(self, root_dir, start_frac, end_frac):
+    random.seed(_SEED)
```
Conchylicultor: This would set the random seed globally and may overwrite the any user defined seed.

Here you can remove it as tfds will automatically shuffle the examples for you.

In the general case, you should be using a local generator instead.
```
r = random.Random(seed=123)
r.shuffle(my_list)
```

jackd:
Changes made. Thanks for the detailed feedback.

Conchylicultor:
Thank you for your contribution. Merged it into master.

PR #78, opened 2019-02-25
Add CheXpert Dataset 

andrewk1:
CheXpert is a dataset of over 200,000 chest radiographs from the Stanford ML Group. Would love to see it here in TFDS!

Here is the requested [Github gist](https://gist.github.com/andrewk1/035b3e9b2e938645273175c32ae9749e)


tensorflow_datasets/image/chexpert.py
```diff
@@ -0,0 +1,114 @@
...
+
+_DESCRIPTION = '''\
+CheXpert is a large dataset of chest X-rays and competition for automated chest 
+x-ray interpretation, which features uncertainty labels and radiologist-labeled 
+reference standard evaluation sets. It consists of 224,316 chest radiographs 
+of 65,240 patients, where the chest radiographic examinations and the associated 
+radiology reports were retrospectively collected from Stanford Hospital. Each 
+report was labeled for the presence of 14 observations as positive, negative, 
+or uncertain. We decided on the 14 observations based on the prevalence in the 
+reports and clinical relevance.
```
rsepassi: This is where we'll want to make very clear that the user needs a special license.

tensorflow_datasets/image/chexpert.py
```diff
@@ -0,0 +1,114 @@
...
+  archivePrefix = {arXiv},
+  eprint    = {1901.07031},
+  timestamp = {Fri, 01 Feb 2019 13:39:59 +0100},
+  biburl    = {https://dblp.org/rec/bib/journals/corr/abs-1901-07031},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+'''
+
+# Path to images and category labels in data dir
+_DATA_DIR =                   'CheXpert-v1.0-small'
```
rsepassi: we tend not to rjustify like this.
also, to ensure this works on non-posix filesystems, use `os.path.join` instead of hardcoding `"/"`

tensorflow_datasets/image/chexpert.py
```diff
@@ -0,0 +1,114 @@
...
+                "label": tfds.features.Sequence(
+                    tfds.features.ClassLabel(names=_LABELS)),
+            }),
+            supervised_keys=("image", "label"),
+            urls=["https://stanfordmlgroup.github.io/competitions/chexpert/"],
+            citation=_CITATION
+        )
+
+    def _split_generators(self, dl_manager):
+        path = dl_manager.manual_dir
```
rsepassi: is it not possible to download this dataset from the open web?

tensorflow_datasets/image/chexpert.py
```diff
@@ -0,0 +1,114 @@
...
+                  f'website manually and place it into {path}.'
+            raise AssertionError(msg)
+
+        return [
+            tfds.core.SplitGenerator(
+                name=tfds.Split.TRAIN,
+                num_shards=100,
+                gen_kwargs={
+                    "imgs_path": path,  # Relative img path is provided in csv
+                    "csv_path": f'{path}/{_TRAIN_LABELS_FNAME}',
```
rsepassi: `os.path.join`

tensorflow_datasets/image/chexpert.py
```diff
@@ -0,0 +1,114 @@
...
+  eprint    = {1901.07031},
+  timestamp = {Fri, 01 Feb 2019 13:39:59 +0100},
+  biburl    = {https://dblp.org/rec/bib/journals/corr/abs-1901-07031},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+'''
+
+# Path to images and category labels in data dir
+_DATA_DIR =                   'CheXpert-v1.0-small'
+_TRAIN_DIR =                   f'{_DATA_DIR}/train'
```
rsepassi: we support Py 2 and 3, so we can't use f strings.

tensorflow_datasets/image/chexpert.py
```diff
@@ -0,0 +1,114 @@
...
+            for row in reader:
+                # Get image based on indicated path in csv
+                name = row['Path']
+                labels = [row[key] for key in label_keys]
+                data.append((name, labels))
+
+        for name, labels in data:
+            yield {
+                'name': name,
+                'image': f'{imgs_path}/{name}',
```
rsepassi: os.path

dynamicwebpaige:
Looks great to me! Thanks so much for submitting, Andrew. We'd love to have the CheXpert data set added to TFDS! 😄

tensorflow_datasets/image/chexpert.py
andrewk1: Sounds good 👍

tensorflow_datasets/image/chexpert.py
andrewk1: No, unfortunately a user must individually accept a usage agreement before downloading the dataset. This seems similar to the ImageNet case which also has such a restriction.

andrewk1:
Summary of changes:
- Replace f strings with `os.path.join`
- Clarify data research use agreement in `_DESCRIPTION`

Thanks!

tensorflow_datasets/image/chexpert.py
```diff
@@ -0,0 +1,118 @@
...
+
+# Path to images and category labels in data dir
+_DATA_DIR = 'CheXpert-v1.0-small'
+_TRAIN_DIR = os.path.join(_DATA_DIR, 'train')
+_VALIDATION_DIR = os.path.join(_DATA_DIR, 'valid')
+_TRAIN_LABELS_FNAME = os.path.join(_DATA_DIR, 'train.csv')
+_VALIDATION_LABELS_FNAME = os.path.join(_DATA_DIR, 'valid.csv')
+
+# Labels per category
+_LABELS = ['', '-1.0', '0.0', '1.0']
```
rsepassi: Are there more meaningful names you could put here?

andrewk1: Currently, each image has labels for 14 separate categories, where each label can take any value from `_LABELS`. Thus, `_LABELS` is consistent with those found in the dataset. I was under the assumption that the `names` for a `ClassLabel` must be the ones present in the data. Is there a way to make them more clear?

rsepassi: Hmm..yeah, we want to represent the data faithfully, but clarity too is a virtue.

What are the semantic meanings of these values? We could do, for example:

```
# Maps to 0=negative, 1=positive, 2=uncertain, 3=unknown
_LABELS = OrderedDict{
  "-1.0": "negative",
  "1.0": "positive",
  "0.0": "uncertain",
  "": "unknown",
}

# In info
ClassLabel(names=_LABELS.values())

# In generate
yield {
  ...
  "label": _LABELS[label]
}
```

andrewk1: The meanings are blank for unmentioned, 0 for negative, -1 for uncertain, and 1 for positive. I think your suggestion is a good way to maintain clarity! I will add that in 👍

andrewk1: Do you mean `_LABELS.keys()`? for `ClassLabel` since we want the true data labels?

andrewk1: Ah sorry on the previous comment - I understand what you meant now.

PR #95, opened 2019-02-27
Fixed broken svg

jidroid404:
tfds-kokoro-public/kokoro-build.svg was broken.


jidroid404:
I signed it!

rsepassi:
Thank you! We definitely need to fix this, but that svg is continuously updated on the GCS bucket by the testing infrastructure (Kokoro). I'm reaching out to them to see if they can set the right content type.

PR #125, opened 2019-03-03
Fixed errors and descriptions

kyscg:

PR #130, opened 2019-03-04
Fixed errors and cleaned descriptions

kyscg:


kyscg:
Fixed issue in #125

tensorflow_datasets/image/celebahq.py
```diff
@@ -43,7 +43,8 @@
 High-quality version of the CELEBA
 dataset, consisting of 30000 images in 1024 x 1024 resolution.
 
-WARNING: this dataset currently requires you to prepare images on your own.
+WARNING: This dataset currently requires you to prepare images on your own. Tensorflow also provides an image generator
```
cyfra: The link you added is to Tensor2Tensor library.

Not sure what you meant by 'provides an image generation'?
That library unfortunately also requires you to prepare the data yourself.

kyscg: I meant to provide a method for those interested in image generation, does it seem unnecessary?

cyfra: I think that it is not necessary.

PR #136, opened 2019-03-04
Fix typo in dataset.md

KumarArindam:
Found a small typo in the datasets.md


cyfra:
Thanks for the PR. This file is autogenerated and we've fixed the culprit in #130.
It should be updated soon.

PR #141, opened 2019-03-04
Displayed the Download Size for Each dataset

### LINKED ISSUE #120: [Documentation] Display the download size for each dataset.
Currently, just by looking at the [list of datasets available in TFDS](https://github.com/tensorflow/datasets/blob/master/docs/datasets.md#all-datasets), there is no way to know the size of each dataset prior to downloading. Users may be operating under constrained disk space, and should be informed of the size of the dataset before requesting.

This feature enhancement would **detail the download size of each dataset** on the markdown file referenced above.

ParthS007:
Fixes #120 

- Display the download size for each dataset.
- I have converted the size to MB for human-readable Formatting.
- Added an extra column and created a Markdown table for it.

@Conchylicultor @dynamicwebpaige Please review and let me know if I have to do any changes.

Thanks

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -29,47 +29,69 @@ np_datasets = tfds.as_numpy(datasets)
 ## All Datasets
 
 * [`audio`](#audio)
-  * [`"nsynth"`](#nsynth)
+
+    |            Name           	| Size(MiB)	|
+    |------------------------------	|----------	|
+    | [ `"nsynth"` ]( #nsynth ) 	|     NA   	|
+
 * [`image`](#image)
-  * [`"cats_vs_dogs"`](#cats_vs_dogs)
-  * [`"celeb_a"`](#celeb_a)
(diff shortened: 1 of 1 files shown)


ParthS007:
@rsepassi Please review and let me know if I have to do any changes.

Thanks

docs/datasets.md
```diff
@@ -29,47 +29,69 @@ np_datasets = tfds.as_numpy(datasets)
 ## All Datasets
 
 * [`audio`](#audio)
```
cyfra: This file is auto-generated.

You should change the generator instead: scripts/document_datasets.py

ParthS007: @cyfra Should I make a new PR for that and close this one or make changes here only?

Conchylicultor: Thanks for your contribution. I think https://github.com/tensorflow/datasets/pull/145 is already accomplishing this so we could probably close this one.

ParthS007: okay :+1:

PR #145, opened 2019-03-04
Added dataset size to the Documentation

### LINKED ISSUE #120: [Documentation] Display the download size for each dataset.

ChanchalKumarMaji:
Fixes #120 
I have added the size of each dataset which is available. Please check if there are any errors.
@Conchylicultor , @dynamicwebpaige , @rsepassi please review.

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -9,7 +9,7 @@ tfds.list_builders()
 
 # Load a given dataset by name, along with the DatasetInfo
 data, info = tfds.load("mnist", with_info=True)
-train_data, test_data = data['train'], data['test']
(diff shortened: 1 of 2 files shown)


cyfra:
Thanks - I'll merge it soon.

PR #148, opened 2019-03-04
Implemented Mozilla Common Voice Dataset (English)

captain-pool:
@rsepassi  @cyfra Please take a look.


tensorflow_datasets/audio/commonvoice.py
```diff
@@ -0,0 +1,57 @@
+import tensorflow as tf
+import numpy as np
+import tensorflow_datasets.public_api as tfds
+import os
+import tensorflow as tf
+_DOWNLOAD_LINK = {"en":"https://voice-prod-bundler-ee1969a6ce8178826482b88e843c335139bd3fb4.s3.amazonaws.com/cv-corpus-1/en.tar.gz"}
+_SPLITS = {tfds.Split.TRAIN:"train",tfds.Split.TEST:"test",tfds.Split.VALIDATION:"validated"}
+class commonvoice(tfds.core.GeneratorBasedBuilder):
+    VERSION = tfds.core.Version("1.0.0")
+    def _info(self):
```
cyfra: please leave:
 - 1 empty line before each function
 - 2 empty lines before class name

(see other files as example)

tensorflow_datasets/audio/commonvoice.py
```diff
@@ -0,0 +1,57 @@
+import tensorflow as tf
+import numpy as np
+import tensorflow_datasets.public_api as tfds
+import os
+import tensorflow as tf
+_DOWNLOAD_LINK = {"en":"https://voice-prod-bundler-ee1969a6ce8178826482b88e843c335139bd3fb4.s3.amazonaws.com/cv-corpus-1/en.tar.gz"}
```
cyfra: add empy line before

tensorflow_datasets/audio/commonvoice.py
```diff
@@ -0,0 +1,57 @@
...
+_SPLITS = {tfds.Split.TRAIN:"train",tfds.Split.TEST:"test",tfds.Split.VALIDATION:"validated"}
+class commonvoice(tfds.core.GeneratorBasedBuilder):
+    VERSION = tfds.core.Version("1.0.0")
+    def _info(self):
+       return tfds.core.DatasetInfo(
+       description = ("Mozilla Common Voice Dataset (English)"),
+       builder = self,
+       features = tfds.features.FeaturesDict({
+       "client_id":tfds.features.Text(),
```
cyfra: This should be indented more

tensorflow_datasets/audio/commonvoice.py
```diff
@@ -0,0 +1,57 @@
...
+        dl_path = dl_manager.extract(dl_manager.download(_DOWNLOAD_LINK))
+        clip_folder = os.path.join(dl_path["en"],"clips") # Need to Check for replacement
+        return [tfds.core.SplitGenerator(
+        name = k,
+        num_shards = 40,
+        gen_kwargs = {
+        "audio_path":clip_folder,
+        "label_path":os.path.join(dl_path["en"],"%s.tsv"%v)
+        }) for k,v in _SPLITS.items()]
+    def _generate_examples(self,audio_path,label_path):
```
cyfra: there should be a space after each comma:

def _generate_examples(self, audio_path, label_path)

tensorflow_datasets/audio/commonvoice.py
```diff
@@ -0,0 +1,57 @@
...
+        def decode_tsv(line):
```
cyfra: add comment on what this decode does.

tensorflow_datasets/audio/commonvoice.py
```diff
@@ -0,0 +1,57 @@
...
+            record_defaults = [""]*8
+            data = tf.decode_csv(line,record_defaults,field_delim = "\t")
+            return data
+        def convert(byte,type_):
+            assert callable(type_)
+            v = byte.decode("utf-8")
+            return type_(v) if len(v)>0 else 0
+        ds = tfds.as_numpy(tf.data.TextLineDataset(label_path).skip(1).map(decode_tsv))
```
cyfra: if this is a CSV file: you can use python's csv DictReader

cyfra:
Please run pylint over your code (should flag most of the issues I've mentioned)

captain-pool:
Sure @cyfra. Will Do. Thanks for Pointing Out.

captain-pool:
@cyfra Can you point me to the Test(and Fake Data) you were talking about? I couldn't find any such thing in 
[tensorflow_datasets/audio/nsynth.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/nsynth.py), [tensorflow_datasets/audio/librispeech.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/librispeech.py)

tensorflow_datasets/audio/commonvoice.py
captain-pool: @cyfra  It's a TSV file (Tab Seperated Values). 
Thanks for the Suggestion. Will keep in mind from next time.
Also, for this case, Is it necessary to use the `csv.DictReader`?

cyfra:
Please look at the "Testing my dataset" section of the documentation:
https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md#testing-mydataset

You have to create 'dataset-like' fake objects, put them in a proper directory and write a unittest.

Yes, you're right - we didn't create unittests for those- my bad. 
I've filed bug reports to get them fixed too.

cyfra:
And when you do the code fixes -please make sure to click to "Resolve comment" - so that I know that code is ready for another round of reviews.

tensorflow_datasets/audio/commonvoice.py
cyfra: The main argument for using DictReader  - it should read the header - and yield rows as a map rather than as a list.

This means that in line 47 below, you don't need to worry about having the correct order of arguments (which can lead to very subtle bugs) - as you can do:

for row in ds:
  ... row['path']... row['wave']

etc

captain-pool: Oh! I get it! Thanks for explaining. Will update the code.

captain-pool:
@cyfra How to run the unittest code? I tried installing the library using `python3 setup.py install` which didn't install the test_data/fake_examples directory. So, I manually copied the test_data folder to the location.
Now when I'm running the test, it is saying the class doesn't have `__builtins__` implemented in it.

PR #165, opened 2019-03-06
Fixed Kaggle Downloader

us:
When cleaning output of competition_files, it's added the empty ("") line to return value and then given an error. I added the last check with if condition.

Kaggle *competition_files* *output* is 
```
name,size,creationDate
train.csv,73MB,None
test.csv,49MB,None
sample_submission.csv,235KB,2016-05-16 01:41:34
```
and then I write that code:
```python
for x, fname in enumerate(downloader.competition_files):
    print("---- ", x, fname, type(fname))
```
It's seperate like that:
```
----  0  <class 'str'>
----  1 sample_submission.csv <class 'str'>
----  2 test.csv <class 'str'>
----  3 train.csv <class 'str'>
```
And then when I am downloading:
```downloader.download_file(fname, dir_)```
It's given error. Because of zeroth variable that is empty.

I fixed this with if condition.

### DIFF
--- a/tensorflow_datasets/testing/test_utils.py
+++ b/tensorflow_datasets/testing/test_utils.py
@@ -401,7 +401,7 @@ def mock_kaggle_api(filenames=None, err_msg=None):
                                             tf.compat.as_bytes(err_msg))
       return tf.compat.as_bytes(
           "\n".join(["name,size,creationDate"] +
-                    ["%s,34MB,None" % fname for fname in filenames]))
+                    ["%s,34MB,None\n" % fname for fname in filenames]))
 
     return check_output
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tensorflow_datasets/testing/test_utils.py [hunk 1]
```diff
@@ -401,7 +401,7 @@ def check_output(command_args):
                                             tf.compat.as_bytes(err_msg))
       return tf.compat.as_bytes(
           "\n".join(["name,size,creationDate"] +
-                    ["%s,34MB,None" % fname for fname in filenames]))
+                    ["%s,34MB,None\n" % fname for fname in filenames]))
```
kyscg: What is the purpose of this change?

us: Check the none line.

PR #169, opened 2019-03-06
Free disk size todo

us:


us:
#172 that's cleaned version of pr.

PR #172, opened 2019-03-06
Free disk size (todo)

us:
 I made many wrong mistakes in the previous pull request. Now I cleared it.

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -389,8 +389,8 @@ class DatasetBuilder(object):
         if dataset_size > utils.py_utils.free_disk_size(self._data_dir_root):
             raise IOError("Not enough disk space!!\nDataset size : {dataset_size} \nFree size : {free_disk_size} \nYou need to extra {needed_disk_size} to download."
                           .format(dataset_size=units.size_str(dataset_size),
-                                  free_disk_size=units.size_str(utils.py_utils.free_disk_size),
-                                  needed_disk_size=units.size_str(dataset_size - utils.py_utils.free_disk_size)))
+                                  free_disk_size=units.size_str(utils.py_utils.free_disk_size(self._data_dir_root)),
+                                  needed_disk_size=units.size_str(dataset_size - utils.py_utils.free_disk_size(self._data_dir_root))))
 
     _check_disk_size()
 
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -38,7 +38,6 @@ from tensorflow_datasets.core import registered
 from tensorflow_datasets.core import splits as splits_lib
 from tensorflow_datasets.core import units
 from tensorflow_datasets.core import utils
-from tensorflow_datasets.core.utils import py_utils
 
 import termcolor
 
@@ -386,11 +385,11 @@ class DatasetBuilder(object):
 
     dataset_size = self.info.size_in_bytes
     def _check_disk_size():
-        if dataset_size > utils.py_utils.free_disk_size(self._data_dir_root):
+        if dataset_size > utils.free_disk_size(self._data_dir_root):
             raise IOError("Not enough disk space!!\nDataset size : {dataset_size} \nFree size : {free_disk_size} \nYou need to extra {needed_disk_size} to download."
                           .format(dataset_size=units.size_str(dataset_size),
-                                  free_disk_size=units.size_str(utils.py_utils.free_disk_size(self._data_dir_root)),
-                                  needed_disk_size=units.size_str(dataset_size - utils.py_utils.free_disk_size(self._data_dir_root))))
+                                  free_disk_size=units.size_str(utils.free_disk_size(self._data_dir_root)),
+                                  needed_disk_size=units.size_str(dataset_size - utils.free_disk_size(self._data_dir_root))))
 
     _check_disk_size()
 
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -382,10 +382,10 @@ class DatasetBuilder(object):
     # Print is intentional: we want this to always go to stdout so user has
     # information needed to cancel download/preparation if needed.
     # This comes right before the progress bar.
+
     stat = os.statvfs('/')
     free_disk_size = (stat.f_bavail * stat.f_frsize) / 1024
     dataset_size = self.info.size_in_bytes
-
     def _check_disk_size():
         if (dataset_size > free_disk_size ) :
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/dataset_builder.py
```diff
@@ -382,13 +382,24 @@ def _log_download_bytes(self):
     # Print is intentional: we want this to always go to stdout so user has
     # information needed to cancel download/preparation if needed.
     # This comes right before the progress bar.
+
+    stat = os.statvfs('/')
```
rsepassi: let's move this into py_utils so that dataset builder just calls:

`if dataset_size > py_utils.free_disk_size(self._data_dir):`

rsepassi:
Thank you!

tensorflow_datasets/core/dataset_builder.py
us: Thank you. I moved this into py_utils.

But if call function like this `(self._data_dir)` when I run the test `python -m tensorflow_datasets.scripts.download_and_prepare --datasets=mnist`, I got an error `FileNotFoundError: [Errno 2] No such file or directory: '/Users/user1/tensorflow_datasets/mnist/1.0.0'
` because that file not downloaded. It's check before creating file. So I called the root directory.

rsepassi: (looks like you need to push the commit; I'm still seeing the old file)

rsepassi: And overall we need to think about doing this in a way that's robust to:
* OS: Linux, Windows
* Machines with multiple disks (i.e. we need to use the one that data_dir is on)
* Cloud filesystems (i.e. if the data_dir were set to `gs://...`; in that case we want to skip this check)

us: I import the file name like this `from tensorflow_datasets.core.utils import py_utils` because I've been warned before about that. But I don't remember clearly, is it problem ? Isn't it clean `from tensorflow_datasets.core.utils.py_utils import free_disk_size` ?

rsepassi: We always import modules, never individual items.

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -38,6 +38,7 @@
 from tensorflow_datasets.core import splits as splits_lib
 from tensorflow_datasets.core import units
 from tensorflow_datasets.core import utils
+from tensorflow_datasets.core.utils import py_utils
```
rsepassi: Actually utils already includes all of py_utils, so you can rm this and do `utils.free_disk_size`

tensorflow_datasets/core/utils/py_utils.py
```diff
@@ -275,3 +275,9 @@ def rgetattr(obj, attr, *args):
   def _getattr(obj, attr):
     return getattr(obj, attr, *args)
   return functools.reduce(_getattr, [obj] + attr.split("."))
+
+
+def free_disk_size(directory):
```
rsepassi: (copied from earlier)

Overall we need to think about doing this in a way that's robust to:
* OS: Linux, Windows
* Machines with multiple disks (i.e. we need to use the one that data_dir is on)
* Cloud filesystems (i.e. if the data_dir were set to `gs://...`; in that case we want to skip this check)

tensorflow_datasets/core/dataset_builder.py
us: Yes I see but I didn't know what to do because I just thought readability becomes poor, I was confused sorry!

tensorflow_datasets/core/utils/py_utils.py
us: Sorry I just missed it. I am fixed immediately.

tensorflow_datasets/core/dataset_builder.py
us: Okay I got it. Thank you

tensorflow_datasets/core/dataset_builder.py
us: Fixed it.

tensorflow_datasets/core/utils/py_utils.py
us: Fixed with using `_data_dir_root`.

tensorflow_datasets/core/dataset_builder.py
rsepassi: rm

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -382,13 +383,22 @@ def _log_download_bytes(self):
     # Print is intentional: we want this to always go to stdout so user has
     # information needed to cancel download/preparation if needed.
     # This comes right before the progress bar.
+
+    dataset_size = self.info.size_in_bytes
+    def _check_disk_size():
+        if dataset_size > utils.py_utils.free_disk_size(self._data_dir_root):
```
rsepassi: `utils.free_disk_size`

tensorflow_datasets/core/utils/py_utils.py
```diff
@@ -275,3 +275,9 @@ def rgetattr(obj, attr, *args):
...
+  stat = os.statvfs(directory)
```
rsepassi: does this work on Windows and Linux?
maybe we should wrap it in a try/except and if it fails, return `float("inf")`
i think that would also handle the case of cloud filesystems

tensorflow_datasets/core/dataset_builder.py
us: okey

tensorflow_datasets/core/dataset_builder.py
us: okey

tensorflow_datasets/core/utils/py_utils.py
us: can I use psutil package? It's work on windows and linux. 
I asked because I don't want to add more package to library.

us: psutil also have free disk usage method. `psutil.disk_usage(".").free` 
[link](https://stackoverflow.com/a/29944093/8401425)

us: I fixed like that [pr](https://github.com/tensorflow/datasets/pull/172/commits/7e38b62c838bfdbad351c3cab43551c981cad6b5)

rsepassi: Yeah, let's go with psutil. Fine to add as a dep.

us: Should I closed this pr and create a more clean ?

rsepassi:
Up to you. Fine by me to update in place.

us:
#255

Issue #154, opened 2019-03-05
Add test and fake data for nsynth


cyfra:
Nsyth dataset is missing unittest and fake data.

Please follow 
https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md#testing-mydataset

captain-pool:
@cyfra Can you assign this to me? I'll make the tests for both commonvoice and nsynth.

cyfra:
@captain-pool - Thanks!

captain-pool:
@cyfra Take a look. There's some bug with nsynth dataset builder, I created a new issue for that. That's a critical bug and it needs to be fixed for the unittest to start working.

captain-pool:
@cyfra Take a look. There's some bug with nsynth dataset builder, I created a new issue for that.(#170) That's a critical bug and it needs to be fixed for the unittest to start working.

PR #183, opened 2019-03-07
Added Fake Samples and Test for Nsynth Dataset

captain-pool:
#154 @cyfra


tensorflow_datasets/audio/commonvoice.py
```diff
@@ -0,0 +1,126 @@
+"""
```
cyfra: you seem to be adding commonvoice - is that intended ?

captain-pool: @cyfra I mistakenly committed the code to the same branch on which I was working for commonvoice. This resulted in appending of the Commit to the PR I made for commonvoice. So, I performed a git reset --soft, created a new branch for this issue and force-pushed. This branch creation step, appended all the commonvoice commits along with the Nsynth Commit.
Can you suggest me a way to fix this?

cyfra: There are couple different ways to fix it (and plenty of git documentation if you want to know the details).

The simplest one I'd suggest is:
- create a new branch from MASTER (so it doesn't contain any of your changes)
- use "git cherry-pick"  to import the changes that you need
- push this branch to github.

captain-pool: Sure. Then It would be better that I close this PR.

Issue #149, opened 2019-03-04
`download_and_prepare` when developing an in-development dataset redownloads data every time


rsepassi:
**Short description**
When developing a new dataset, it seems that the `download_and_preapre` redownloads the data every time. 

**Environment information**
* Operating System: Ubuntu
* Python version: 3.6
* `tensorflow-datasets`/`tfds-nightly` version: `tfds-nightly==1.0.1.dev201903040105`
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: `tensorflow==1.13.1`

**Reproduction instructions**

```
# Change the MNIST class name in the code so that it seems like it's a new dataset
# class MNISTZ

# Run once to see the data downloading, and kill it before it finishes
# Then run it again and see that it redownloads
python -m tensorflow_datasets.scripts.download_and_prepare --datasets=mnistz
```

**Expected behavior**
The downloads to be cached.

adarob:
Looking at the code, it seems that the sha256 hashes are only stored after the dataset building completes, which causes the files to be re-downloaded every time. It would be very nice if this was relaxed, at least for development.

ubershmekel:
This is really slowing me down :(

Conchylicultor:
The current hack is to locally change the https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/download/download_manager.py#L210:
From:
```
if (not self._force_download and resource.sha256 and
    resource.exists_locally()):
```

To:
```
if not self._force_download and resource.exists_locally():
```
But it has side effect elsewhere (like not computing checksums anymore).

Waiting for @pierrot0 for a more stable fix.

jackd:
Is the proposition to cache on the builder level, or the `data_dir` level? Is there any reason `mnistz` from above would not want to use downloaded data from `mnist`, given the url is the same?

pierrot0:
The proposition is to cache globally.
Path will land soon.

PR #194, opened 2019-03-08
Modelnet

jackd:
... this was supposed to just be modelnet10/40 with tri-mesh and point cloud sampling... then I found [this](https://github.com/charlesq34/pointnet2/tree/master/data) and extending it to their sampled data/shapenet parts didn't seem too much more. Apologies for not splitting it up.

One issue I haven't been able to resolve is repeated downloads for different configs of the same dataset. `builder.info.download_checksums` just seem to disappear occasionally between runs. Could be related to [this](https://github.com/tensorflow/datasets/issues/149), but I suspect it's a different issue.

Note sure what the feeling is regarding `volume` submodule - just seemed like a good catch-all for point cloud/mesh/voxels. I suspect there'll be less controversial aspects here compared to [this PR](https://github.com/tensorflow/datasets/pull/32), but once either is resolved I'm happy to modify the other so that all shapenet stuff is sitting together (assuming image -> volume datasets are more `volume`-ous than `image`-ish).

Also happy to add the shapenet ICCV2017 reconstruction track (to partner the point cloud segmentation one) once shape issues with RLE encodings have been resolved.


jackd:
dataset info [here](https://gist.github.com/jackd/90815cb20fbdabb00d3aafab6773b23a).

jackd:
I have no idea why this presubmit is failing... can't install tensorflow-datasets? All I did was add a lazy import...

ataiya:
Hi @jackd would you like to add this dataset under tensorflow_graphics instead?
I see this has been here for about a year with no updates.

jackd:
@ataiya Yeah, I gave up waiting here/submitting any more PRs and just made a [separate repository](https://github.com/jackd/shape-tfds) with this and some other shape stuff. If it's likely to get through at graphics then I'm happy to clean some of them up, but I wouldn't mind getting a clearer idea of what'll be prioritized at the other end before I make too many PRs like this one.

Issue #24, opened 2019-01-20
[data request] Horses and Zebras


joel-shor:
* Name of dataset: Horses and Zebras
* URL of dataset: https://github.com/junyanz/CycleGAN/blob/master/datasets/download_dataset.sh
* License of dataset: <license type>
* Short description of dataset and use case(s): Image to image examples using in CycleGAN.

Folks who would also like to see this dataset in `tensorflow/datasets`, please +1/thumbs-up so the developers can know which requests to prioritize.

ChanchalKumarMaji:
Can you please tell me, what exactly is needed to be done here ? Is knowledge of GAN necessary to contribute to this issue ?

rsepassi:
Hi @ChanchalKumarMaji, thanks for your interest! No, knowledge of GANs is not necessary. I'd recommend having a single CycleGAN `DatasetBuilder` that has a `BuilderConfig` for each of the [listed datasets](https://github.com/junyanz/CycleGAN/blob/bba55b0e4d1e002c86892959da251b3e0de706dc/datasets/download_dataset.sh#L4). You can start with a single `BuilderConfig` just for `horse2zebra`. And you'd include all the splits `trainA`, `trainB`, `testA`, `testB`. Does that help you get started?

This is [our guide on adding a dataset](https://tensorflow.org/datasets/add_dataset).

ChanchalKumarMaji:
Thanks @rsepassi for your response.

This is what I need to do to solve this issue (please correct me if I am wrong):
1. Create a new folder named <code>CycleGAN</code> under <code>datasets/tensorflow_datasets/</code>
2. In this folder, I need to have the <code>BuilderConfig</code> for the following DataSets : 
        -> <code>apple2orange</code> 
        -> <code>summer2winter_yosemite</code>
        -> <code>horse2zebra</code>
        -> <code>monet2photo</code>
        -> <code>cezanne2photo</code>
        -> <code>ukiyoe2photo</code>
        -> <code>vangogh2photo</code>
        -> <code>maps</code>
        -> <code>cityscapes</code>
        -> <code>facades</code>
        -> <code>iphone2dslr_flower</code> 
        -> <code>ae_photos</code>
3. I will first implement <code>BuilderConfig</code> for <code>horse2zebra</code>.
4. Will my implementation be very similar to <code>cats_vs_dogs.py</code> and <code>cats_vs_dogs_test.py</code> in the image folder. I am planning to reuse codes.
[...]

PR #212, opened 2019-03-11
New Dataset Cycle_GAN added

ChanchalKumarMaji:
Added Cycle_GAN Dataset.
Dataset Request #24 

@rsepassi @cyfra @Conchylicultor @dynamicwebpaige @tfds-copybara , Please review.


tensorflow_datasets/image/__init__.py
```diff
@@ -38,3 +38,5 @@
 from tensorflow_datasets.image.quickdraw import QuickdrawBitmap
 from tensorflow_datasets.image.rock_paper_scissors import RockPaperScissors
 from tensorflow_datasets.image.svhn import SvhnCropped
+from tensorflow_datasets.image.cycle_gan import CycleGANConfig
```
cyfra: this line is probably not needed.

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+      raise ValueError("data must be one of %s" % _DATA_OPTIONS)
+  
+    super(CycleGANConfig, self).__init__(**kwargs)
+    self.data = data
+
+    
+class CycleGAN(tfds.core.GeneratorBasedBuilder):
+  """CycleGAN dataset."""
+  
+  VERSION = tfds.core.Version("2.0.0")
```
cyfra: what's the reason for 2.0?

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+  
+  BUILDER_CONFIGS = [
+      CycleGANConfig(
```
cyfra: this is a lot of copy-paste (which is prone to typos).

Consider doing:

BUILDER_CONFIGS = [
  CycleGANConfig(name=config_name, ...) for config_name in _DATA_OPTIONS]

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+from __future__ import print_function
+
+import os
+
+import numpy as np
+import tensorflow as tf
+
+from tensorflow_datasets.core import api_utils
+import tensorflow_datasets.public_api as tfds
+
```
cyfra: citation missing

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+
+_DL_URL = "https://people.eecs.berkeley.edu/~taesung_park/CycleGAN/datasets/"
+_DL_URLS = {
```
cyfra: if the names are matching the zip files, you can do:

_DL_URLS = { name: _DL_URL + name + ".zip" for name in _DATA_OPTIONS}

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+            "label": tfds.features.ClassLabel(names=["A", "B"]), 
+        }),
+        supervised_keys=("image", "label"),
+        urls=["https://people.eecs.berkeley.edu/~taesung_park/CycleGAN/datasets/"],
+    )
+
+  def _split_generators(self, dl_manager):
+    url = _DL_URLS[self.builder_config.name]
+    data_dirs = dl_manager.download_and_extract(url) 
+    print(data_dirs)
```
cyfra: remove

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+    path_to_dataset = data_dirs + "/" + tf.io.gfile.listdir(data_dirs)[0]
+    print(path_to_dataset)
```
cyfra: remove

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+        }),
+        supervised_keys=("image", "label"),
+        urls=["https://people.eecs.berkeley.edu/~taesung_park/CycleGAN/datasets/"],
+    )
+
+  def _split_generators(self, dl_manager):
+    url = _DL_URLS[self.builder_config.name]
+    data_dirs = dl_manager.download_and_extract(url) 
+    print(data_dirs)
+    path_to_dataset = data_dirs + "/" + tf.io.gfile.listdir(data_dirs)[0]
```
cyfra: use os.path.join

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+            name="testB",
+            num_shards=1,
+            gen_kwargs={"files": testB_files,
+                        "label": "B",
+            }
+        ),
+    ]
+
+  def _generate_examples(self, files, label): 
+    path=files[0]+"/"
```
cyfra: os.path.join

tensorflow_datasets/image/cycle_gan.py
```diff
@@ -0,0 +1,213 @@
...
+    images=tf.io.gfile.listdir(path)  
+
+    for image in images:
+      yield{
+          "image": path+"/"+image,
```
cyfra: os.path.join

tensorflow_datasets/image/cycle_gan_test.py
```diff
@@ -0,0 +1,42 @@
...
+      "trainB": 2,
+      "testB": 2,
+  }
+  
+if __name__ == "__main__":
+  testing.test_main()
+  
+  
+  
+
```
cyfra: remove lines from the end of the file.

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: One thing I want to mention here is that, the dataset <code>ae_photos</code> has only testA and testB, (trainA and trainB missing). Should I remove such sets ?

tensorflow_datasets/image/__init__.py
ChanchalKumarMaji: Ok, removing this line produces no error, but can you please tell me why it is used [here](https://github.com/tensorflow/datasets/blob/96fe8c3156b43eb776456522ac739a0a6cdb86d2/tensorflow_datasets/text/__init__.py#L19), [here](https://github.com/tensorflow/datasets/blob/96fe8c3156b43eb776456522ac739a0a6cdb86d2/tensorflow_datasets/text/__init__.py#L21), [here](https://github.com/tensorflow/datasets/blob/96fe8c3156b43eb776456522ac739a0a6cdb86d2/tensorflow_datasets/audio/__init__.py#L19) and [here](https://github.com/tensorflow/datasets/blob/96fe8c3156b43eb776456522ac739a0a6cdb86d2/tensorflow_datasets/video/__init__.py#L21). Should I modify these also ?

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: What should I use here ? Should I use <code>tfds.core.Version('0.1.0')</code>. Please suggest. I see something as <code>MAJOR.MINOR.PATCH</code>.

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: I cannot find any CITATION [here](https://people.eecs.berkeley.edu/~taesung_park/CycleGAN/datasets/). What to do ?

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: Done

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: Done

ChanchalKumarMaji:
@cyfra I have made the necessary changes. Please review. Thanks.

tensorflow_datasets/image/__init__.py
cyfra: you're right. My bad.

tensorflow_datasets/image/cycle_gan.py
cyfra: https://arxiv.org/abs/1703.10593

tensorflow_datasets/image/cycle_gan.py
cyfra: We usually start with 0.1.0 - if dataset is under development and then eventually 1.0.0

Using 2.0 is not an error per se - but it suggests that we had an earlier version

tensorflow_datasets/image/cycle_gan.py
cyfra: Do you know if it is missing on purpose ?

In such case, I'd leave it out (And leave a comment saying why ae_photos is missing)
If it ever becomes needed, we can always add it back.

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: Yes, I changed it to "0.1.0". Thanks.

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: I don't know why they are missing, but I am removing them and adding a comment that these datasets are not present.

tensorflow_datasets/image/cycle_gan.py
ChanchalKumarMaji: Done. Please Check.

ChanchalKumarMaji:
What is it that the "kokoro" is failing ? Any suggestions.

ChanchalKumarMaji:
I think this dataset is ready now. Please Check. Thanks.

cyfra:
Thanks - I'll try to run the final set of tests and merge it.

cyfra:
@ChanchalKumarMaji - please generate fake images instead of including real ones.

ChanchalKumarMaji:
@cyfra , I removed the real images and replaced them with fake_examples. Please check. Thanks.

cyfra:
Great, thanks.

ChanchalKumarMaji:
@cyfra @rsepassi , why is the <code>kokoro</code> build failing. I cannot understand. Thanks.

cyfra:
Kokoro was failing with:
E   AssertionError: Splits 'testA' and 'trainA' are overlapping. Are you sure you want to have the same objects in those splits? If yes, add one one of them to OVERLAPPING_SPLITS class attribute.

you can see this error, when you run the test locally too.

Anyway - I've already fixed it in the local version that is being merged now, so you don't have to do any additional changes.

ChanchalKumarMaji:
My bad, I should not have used the same samples in "trainA" and "testA". Actually, I didn't ran the test locally after changing the images. If any changes are required then please do tell me.

ChanchalKumarMaji:
Ok, I have used different samples in the splits, and I ran the test locally with no errors. Thanks.

cyfra:
Merged. Thanks @ChanchalKumarMaji

PR #227, opened 2019-03-12
Added download_all_files method to kaggle downloader.

us:


tensorflow_datasets/core/download/kaggle.py
```diff
@@ -132,6 +132,13 @@ def download_file(self, fname, output_dir):
     _run_kaggle_command(command, self._competition_name)
     return os.path.join(output_dir, fname)
 
+  def download_all_files(self):
```
cyfra: maybe we should make it a static function?
I'd also add the output_dir argument.

so that people can just call: KaggleDownloader.download_all("foo", "/tmp")?

tensorflow_datasets/core/download/kaggle_test.py
```diff
@@ -32,14 +32,18 @@ class KaggleTest(testing.TestCase):
...
       self.assertEqual(downloader.competition_files, ["a", "b"])
       with testing.tmp_dir() as tmp_dir:
         for fname in downloader.competition_files:
           out_path = downloader.download_file(fname, tmp_dir)
           self.assertEqual(out_path, os.path.join(tmp_dir, fname))
           with tf.io.gfile.GFile(out_path) as f:
             self.assertEqual(fname, f.read())
+      with testing.tmp_dir():
+        downloader.download_all_files()
+        self.assertCountEqual(os.listdir(competition_name), downloader.competition_files)
```
cyfra: Kokoro fails on py2 and tf12.

Could you use six.assertCountEqual?

tensorflow_datasets/core/download/kaggle.py
us: Okayyy, I done it :)

tensorflow_datasets/core/download/kaggle_test.py
us: Thank you resolved!

us:
@cyfra fixed.

us:
Is it forgotten?

Conchylicultor:
I think the `dl_manager.download_kaggle_data(` already does this.

PR #239, opened 2019-03-13
added mpii human pose dataset/example

jackd:
The provided annotations file is the most hideous thing you'll ever lay eyes on. I've tried to make interroperating with it as easy as possible, but it doesn't make things easy.

It is a regression from images problem, so an argument could be made it should be under `images` rather than a separate `human_pose` module. That said, there are many datasets closely related that have other inputs, e.g. human3.6million uses videos, and a lot of approaches for 3D human pose focus solely on lifting poses from 2D to 3D (i.e. they skip the image processing all together). I think there's a strong case for keeping all these forms together in the same place, but I'm happy to move things around if required.

I'm not sure if the example is appreciated, but I wrote it for validating everything was working and figured it might be useful. Again not sure if it warants it's own directory and happy to move it.

I'm unsure if I've handled held-out data in the test set correctly. Some sequences I've set to have zero elements (and had to make changes to statistics generation to accomodate), others I've set to all zeros.


jackd:
dataset_info [here](https://gist.github.com/jackd/3267ffda30887b44e6f9e3442e91f986)

PR #240, opened 2019-03-13
Fix #237: byte strings got truncated by tf.compat.as_bytes

### LINKED ISSUE #237: FeatureConnector serialization broken for byte strings
**Short description**
I am currently writing a custom FeatureConnector for RLE encoded segmentations. In order to save some storage space I am going to use the byte encoding of the numpy arrays and `tf.io.decode_raw` in order to decode the information (since tf.train.Example only supports int64 for integer data types). However, this line breaks the serialization of the byte string: 
https://github.com/tensorflow/datasets/blob/c8c00f0843166775b44d3eba2e2171b44d2cb689/tensorflow_datasets/core/file_format_adapter.py#L397
One could argue, that this is a Tensorflow compat module bug, however, bytes returned from a FeatureConnector should directly be passed into the Protobuf Message instead of doing unnecessary conversions. 

**Environment information**
* Operating System: Ubuntu 18.04
* Python version: 3.6
* `tensorflow-datasets`/`tfds-nightly` version: tensorflow-datasets 1.0.1
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: tensorflow-gpu 1.13.1

**Reproduction instructions**

```python
import numpy as np
import tensorflow as tf
tf.enable_eager_execution()
arr = np.zeros((64, 64), dtype=np.uint8)
bytes = arr.tobytes()
print(len(bytes))  # 4096
v = np.asarray(bytes).flatten()  # tensorflow_datasets/core/file_format_adapter.py#390
bytes = tf.compat.as_bytes(v[0])  # tensorflow_datasets/core/file_format_adapter.py#397
print(len(bytes)) # 0
```

razorx89:
Fix for issue #237, where byte strings from a custom feature connector get truncated by `tf.compat.as_bytes`.

### DIFF
--- a/tensorflow_datasets/core/file_format_adapter.py
+++ b/tensorflow_datasets/core/file_format_adapter.py
@@ -386,8 +386,17 @@ def _item_to_tf_feature(item, key_name):
         "FeatureConnector should return a numpy array with the correct dtype "
         "instead of a Python list.".format(key_name)
     )
-  if isinstance(v, six.binary_type):
-    return tf.train.Feature(bytes_list=tf.train.BytesList(value=[v]))
+  if isinstance(v, (six.binary_type, six.string_types)):
+    v = [tf.compat.as_bytes(v)]
+    return tf.train.Feature(bytes_list=tf.train.BytesList(value=v))
+  elif isinstance(v, (tuple, list)) and \
+          all(isinstance(x, (six.binary_type, six.string_types)) for x in v):
+    v = [tf.compat.as_bytes(x) for x in v]
+    return tf.train.Feature(bytes_list=tf.train.BytesList(value=v))
+  elif isinstance(v, np.ndarray) and \
+          (v.dtype.kind in ("U", "S") or v.dtype == object):  # binary or unicode
+    v = [tf.compat.as_bytes(x) for x in v.flatten()]
+    return tf.train.Feature(bytes_list=tf.train.BytesList(value=v))
 
   v = np.array(v).flatten()  # Convert v into a 1-d array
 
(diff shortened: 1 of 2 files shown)


razorx89:
I signed it!

tensorflow_datasets/core/file_format_adapter.py
```diff
@@ -386,6 +386,8 @@ def _item_to_tf_feature(item, key_name):
         "FeatureConnector should return a numpy array with the correct dtype "
         "instead of a Python list.".format(key_name)
     )
+  if isinstance(v, bytes) and not isinstance(v, str):
```
rsepassi: Use `six.binary_type`. And this doesn't solve the issue if you pass a list of 0 strings (i.e. multiple all 0 strings)
(see my comment on #237; I think this is only an issue with all 0's).

razorx89:
```python
if isinstance(v, six.bytes_type):
  ...
elif isinstance(v, (list, tuple)) and all((isinstance(x, six.bytes_type) for x in v)):
  ...
```
Would that be acceptable for dealing with multiple byte strings?

tensorflow_datasets/core/file_format_adapter.py
```diff
@@ -386,6 +386,8 @@ def _item_to_tf_feature(item, key_name):
         "FeatureConnector should return a numpy array with the correct dtype "
         "instead of a Python list.".format(key_name)
     )
+  if isinstance(v, six.binary_type):
+    return tf.train.Feature(bytes_list=tf.train.BytesList(value=[v]))
```
rsepassi: rm the branch below that deals with binary/unicode. handle those separately up front here.
you can use six.binary_type and six.text_type
tf.compat.as_bytes can handle str/unicode/bytes.

rsepassi: and please cover the case of a list/tuple of such as well.

razorx89:
Are the changes sufficient or is here anything left to do?

PR #250, opened 2019-03-14
Iris Dataset

apark263:
Adding the small Iris dataset that is useful for toy problems


tensorflow_datasets/image/iris_test.py
```diff
@@ -0,0 +1,20 @@
+"""Tests for tensorflow_datasets.image.lsun."""
```
ursk: copy and paste error?

tensorflow_datasets/image/__init__.py
```diff
@@ -39,3 +39,4 @@
 from tensorflow_datasets.image.quickdraw import QuickdrawBitmap
 from tensorflow_datasets.image.rock_paper_scissors import RockPaperScissors
 from tensorflow_datasets.image.svhn import SvhnCropped
+from tensorflow_datasets.image.iris import Iris
```
ursk: Not an image dataset, move to `structured`?

apark263: yes, I never considered that we are just looking at measurements

tensorflow_datasets/structured/iris.py
```diff
@@ -0,0 +1,92 @@
...
+            urls=["https://archive.ics.uci.edu/ml/datasets/iris"],
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        iris_file = dl_manager.download(IRIS_URL)
+        all_lines = tf.io.gfile.GFile(iris_file).read().split("\n")
+        records = [l for l in all_lines if l]  # get rid of empty lines
+
+        split = 0.8
```
us: You should use [Subsplit](https://www.tensorflow.org/datasets/splits#subsplit) to split the train, test.

tensorflow_datasets/structured/iris.py
```diff
@@ -0,0 +1,92 @@
...
+3 classes of 50 instances each, where each class refers to a type of iris
+plant. One class is linearly separable from the other 2; the latter are NOT
+linearly separable from each other.
+"""
+
+
+class Iris(tfds.core.GeneratorBasedBuilder):
+    """Iris flower dataset"""
+
+    NUM_CLASSES = 3
```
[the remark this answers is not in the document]
apark263: Fixed, thanks

tensorflow_datasets/structured/iris.py
apark263: Looked into doing this, but I don't think this will preserve the class distribution

tensorflow_datasets/structured/iris.py
```diff
@@ -0,0 +1,93 @@
...
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=1,
+            gen_kwargs={"records": train_lines}),
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TEST,
+            num_shards=1,
+            gen_kwargs={"records": test_lines}),
+    ]
```
ChanchalKumarMaji: I think it would be better not to split the dataset.

Conchylicultor: +1. If the original dataset do not defines explicit split, then the dataset should only contains a single TRAIN split. The user can then use https://www.tensorflow.org/datasets/splits to create its own splits

PR #255, opened 2019-03-14
Free disk space

us:
#172 cleaned commit.


us:
@rsepassi I got the pyutil error on linux machines, linux need some dependencies [same error and solution](https://github.com/giampaolo/psutil/issues/1143#issuecomment-334695523)
Should I add my own function to utils? If psutil fix this, we'll use that again

Conchylicultor:
I've added py3.6-dev recently to the open-source installation process.
Hopefully this should unlock you.
https://github.com/tensorflow/datasets/blob/master/oss_scripts/oss_pip_install.sh#L25

You can try to sync and re-run the tests.

us:
Okay now it's passed.

rsepassi:
Thanks! Will merge shortly.

PR #259, opened 2019-03-16
Increased tfds.core version

kyscg:


kyscg:
Review @Conchylicultor @rsepassi ????

Conchylicultor:
Thank for your contribution but could you provide more info for this change ?
Each dataset has a version independant from the global tfds.core version (ex: tfds can be version 1.0.1 and dataset be at 0.3.2). The dataset version should be updated only when there is a change in the dataset (ex: changing the feature key names).
I guess we should provide more documentation about how the dataset versioning is working.

kyscg:
I agree, should we make a pass through the docs and check for inconsistency?

If so, are there any you have in mind I could get working on?

PR #280, opened 2019-03-18
Doc typo fixes

ctiijima:
Random doc fixes caught in my filter.


Conchylicultor:
Thanks for fixing this. Please note that those files are automatically generated, so you have to modify the original doc-string in the `.py` files.

Example: for SplitBase: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/splits.py#L57

ctiijima:
Thanks! I was not aware of that.  I'll fix it and make a new PR.

Issue #234, opened 2019-03-13
Extract multiple and nested files error.


us:
**Short description**
 The current tfds.core.download_manager is unable to extract multiple and nested files. For example we need to download nested compressed files and `dl_manager.download_and_extract` download but then given an error.
File structure like this:
.
+-- gzip.zip
�   +-- emnist-balanced-test-images-idx3-ubyte.gz
�   +-- emnist-balanced-test-labels-idx1-ubyte.gz
�   +-- emnist-balanced-train-images-idx3-ubyte.gz
�   +-- emnist-balanced-train-labels-idx1-ubyte.gz

It's just extract the gzip.zip doesn't extract later.

**Environment information**
* Operating System: `macOS 10.14.3 (Mojave) 64-bit`
* Python version: `3.6.8`
* `tensorflow-datasets`/`tfds-nightly` version: `1.0 / 1.14`
* `tensorflow` version: `1.2`

**Expected behavior**
dl_manager download all files recursively.

**Additional context**
I found this bug when I added emnist dataset.
Related pr #198 
Because of that I used to `dl_manager.manual_dir` 
[related line](https://github.com/tensorflow/datasets/pull/198/files/7699041a9157d9df5c1407fb3f98f28515b8a443#r264791715)

rsepassi:
@us does it error? Or is this a feature you would want?
Could you not call `dl_manager.extract` on each of the files in the interior directory?

rsepassi:
And are you saying that EMNIST could actually be sourced from the web and the only reason you used manual_dir was because of this nested extraction? If the `dl_manager.extract` method works (or `dl_manager.iter_archive`), could we switch EMNIST to not have to use `manual_dir`?

us:
Yes you are right I couldn't understand all of library. I will fixed it.
and yes this is a feature than a bug. I will corrected.

us:
#251

rsepassi:
Thank you. Do you think you could update EMNIST? It's much more usable without using manual_dir.

us:
Yes I will update as a soon as possible.

rsepassi:
Thanks!

PR #284, opened 2019-03-19
Added Emnist dataset downloader

us:
@rsepassi updated EMNIST. #234 
@cyfra #284


tensorflow_datasets/image/mnist_test.py
```diff
@@ -51,9 +50,23 @@ class KMNISTTest(MNISTTest):
...
+    version="1.0.1",
+  ),
+]
+
+
 class EMNISTTest(MNISTTest):
   DATASET_CLASS = mnist.EMNIST
-  BUILDER_CONFIG_NAMES_TO_TEST = ["test"]
+  DATASET_CLASS.BUILDER_CONFIGS = _emnist_test_builder_config
+  DL_EXTRACT_RESULT = "gzip.zip"
```
rsepassi: We prefer to have the fake examples unzipped (at least 1 level). You can use DL_EXTRACT_RESULT to point to the right files. See how MNIST does it.

rsepassi:
Awesome, thank you!

tensorflow_datasets/image/mnist_test.py
us: But when I downloaded the `gzip.zip` file, the download manager creates `gzip` folder that is include main files. So I need to extract one more time. If I understand it correctly `DL_EXTRACT_RESULT` changing the `dl_manager.download_and_extract(URL)` result. And `DL_EXTRACT_RESULT` want to str type not dict. It's given an error `TypeError: join() argument must be str or bytes, not 'set'`. So I didn't like MNIST. Is there a point I can't see. I'll be happy to fix this problem. Thank you.

PR #294, opened 2019-03-20
Added dataset for Cars196 --Issue 202

SoumyadeepJana:


tensorflow_datasets/image/cars196.py
```diff
@@ -79,100 +78,102 @@
...
-        builder=self,
-  
-        description=(_DESCRIPTION),
-        
-        features=tfds.features.FeaturesDict({
-            "image":tfds.features.Image(),
-            "label": tfds.features.ClassLabel(names = _NAMES),
-            "bbox":tfds.features.BBoxFeature(),
-        }),
+    def _info(self):
```
us: Indentation should be 2 as in other files.

SoumyadeepJana: @us  I messed up my PRs as my first PR was through master and now all other PRs have those changes,what do you suggest me to do?

us: You can create new clean branch from master and then cherry picking this commits.
Like that: `git cherry-pick 12429d1..2478033`
I assume: 
12429d1 : first commit 
2478033 : last commit

us: On the other hand your commits are too messy and too much to adding dataset. Also your commit messages should be more meaningful. When looking at commits, that's not good look and not understandable.

cyfra:
+1 to what @us is saying.

Please cleanup the branch. You can use the git rebase (or git cherrypick) - so that only the commits that are important stay.

us:
@cyfra @Conchylicultor I think we can cleanup branch when merging.

wookayin:
Why no squashing?

Issue #320, opened 2019-03-23
Dataset Template Generator Script


us:
**Is your feature request related to a problem? Please describe.**
There are some files and folders to create for each new dataset. It always takes time. 
On the other hand anyone can sometimes miss out something from the documentation and forgot some things.

**Describe the solution you'd like**
A utility script that creates common files, functions and their TODO's for each new dataset.

Sample usage:
`python -m tensorflow_datasets.scripts.generate_dataset_template --name dataset_name --type dataset_type`

You can check my pr #321 about details.

Conchylicultor:
Closing this as some support for this has been added for this in https://github.com/tensorflow/datasets/pull/321 (thanks again for setting this up)

The script can be improved over time depending on the feedbacks

PR #321, opened 2019-03-23
Dataset Template Generator Script #320

us:
This is just an initial  proof of concept to demonstrate how it can be done. I am willing to modify and extend it based on issue comments.

### DIFF
--- a/tensorflow_datasets/scripts/generate_dataset_template.py
+++ b/tensorflow_datasets/scripts/generate_dataset_template.py
@@ -1,0 +1,226 @@
+# coding=utf-8
+# Copyright 2019 The TensorFlow Datasets Authors.
+#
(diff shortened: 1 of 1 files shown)


ChanchalKumarMaji:
@us you can also add `BuilderConfig` block.

us:
Thank you for comments! I would like to discuss contents. I want to create more general structure. But I think `BuildConfig` is a point of detail. What do you think?

ChanchalKumarMaji:
> @ChanchalKumarMaji Thank you for comments! I would like to discuss contents. I want to create more general structure. But I think `BuildConfig` is a point of detail. What do you think?

@us, Probably a separate script for `BuilderConfig` would look good, as most datasets do not need it.

PR #323, opened 2019-03-23
Fixes #317: Added Support for Stanford Question Answering Dataset - 2.0

### LINKED ISSUE #317: [data request] Stanford Question Answering Dataset 2.0
* Name of dataset: Stanford Question Answering Dataset 2.0
* URL of dataset: https://rajpurkar.github.io/SQuAD-explorer/explore/v2.0/dev/
* Short description of dataset and use case(s): Stanford Question Answering Dataset 2.0 is a reading comprehension dataset, consisting of questions posed by crowd workers on a set of Wikipedia articles, where the answer to every question is a segment of text, or span, from the corresponding reading passage, or the question might be unanswerable.

Folks who would also like to see this dataset in `tensorflow/datasets`, please thumbs-up so the developers can know which requests to prioritize.

I am working on this. @cyfra @Conchylicultor Please assign this to me.

ParthS007:
Fixes #317 
@cyfra @Conchylicultor @dynamicwebpaige @us Please review and let me know if I have to do any changes. 
Thanks


tensorflow_datasets/text/stanford_quad.py
```diff
@@ -0,0 +1,210 @@
...
+        for the `tfds.features.text.TextEncoder` used for the features feature.
+      **kwargs: keyword arguments forwarded to super.
+    """
+    super(StanfordquadConfig, self).__init__(**kwargs)
+    self.text_encoder_config = (
+        text_encoder_config or tfds.features.text.TextEncoderConfig())
+
+
+class Stanfordquad(tfds.core.GeneratorBasedBuilder):
+  """The Stanford Question Answering Dataset. Version 2.0."""
```
us: You should add the tfds core version.

tensorflow_datasets/text/stanford_quad.py
```diff
@@ -0,0 +1,212 @@
...
+  _URL = "https://rajpurkar.github.io/SQuAD-explorer/dataset/"
+  _DEV_FILE = "dev-v2.0.json"
+  _TRAINING_FILE = "train-v2.0.json"
+
+  VERSION = tfds.core.Version("2.0.0")
```
us: Why do you start from version `2.0.0` ? 
Learn more about versioning by checking other datasets commits.

tensorflow_datasets/text/stanford_quad.py
```diff
@@ -0,0 +1,212 @@
+# coding=utf-8
```
cyfra: as we already have squad.py (that has 1.1) version - consider changing name of this file (and the class) to Squad2.0

tensorflow_datasets/text/stanford_quad.py
```diff
@@ -0,0 +1,212 @@
...
+                "answers": answers,
+            }
+            if len(example["answers"]) > 1:
+              yield {
+                  "question": example["question"],
+                  "first_answer": example["answers"][0],
+                  "context": example["context"]
+              }
+            else:
+              example["answers"] = ['No answer is supported by the paragraph']
```
cyfra: We should be as "close" to the raw dataset as possible.

Consider adding the "is_impossible" label (which exists now in 2.0) and in case where there is no solution: can you just return empty string as first_answer ?

ParthS007: @cyfra I am not able to return an empty string as it is giving `ValueError: zero-size array to reduction operation minimum which has no identity`

tensorflow_datasets/text/stanford_quad.py
ParthS007: Done @cyfra

tensorflow_datasets/text/stanford_quad.py
ParthS007: Done @cyfra

ParthS007:
@cyfra Can you please tell why Kokoro is failing. The test is running fine locally. Thanks :)

us:
It's given a connection error. It can be tested again if you update. Also you can check kokoro fails on [details](https://source.cloud.google.com/results/invocations/0c7d0755-15fc-4afc-bdea-fa7914e22536).

tensorflow_datasets/text/squad2.py
```diff
@@ -0,0 +1,218 @@
...
+            }
+            if len(example["answers"]) > 1:
+              yield {
+                  "question": example["question"],
+                  "first_answer": example["answers"][0],
+                  "context": example["context"],
+                  "is_impossible": example["is_impossible"]
+              }
+            else:
+              example["answers"] = ['No answer is supported by the paragraph']
```
cyfra: does such string exist in the dataset ?

If not - then consider returning an empty string instead.
(we don't to stay as close to dataset content as possible).

tensorflow_datasets/text/squad2.py
```diff
@@ -0,0 +1,218 @@
...
+              yield {
+                  "question": example["question"],
```
cyfra: 206 - 209 are identical as 214-217.

Maybe refactor your code, so that the "yield" part exists only once.

ParthS007: okay @cyfra, I will make the change. :+1:

tensorflow_datasets/text/squad2.py
ParthS007: Done @cyfra

ParthS007:
@cyfra @Conchylicultor @tfds-copybara Please review. Thanks :)

ParthS007:
@cyfra Please review. Thanks :)

ParthS007:
@tfds-copybara Please review.

ParthS007:
@us Can you please help me what is the issue here?

tensorflow_datasets/text/squad2.py
```diff
@@ -0,0 +1,213 @@
...
+        text_encoder_config or tfds.features.text.TextEncoderConfig())
+
+
+class Squad2(tfds.core.GeneratorBasedBuilder):
+  """The Stanford Question Answering Dataset. Version 2.0."""
+  _URL = "https://rajpurkar.github.io/SQuAD-explorer/dataset/"
+  _DEV_FILE = "dev-v2.0.json"
+  _TRAINING_FILE = "train-v2.0.json"
+
+  VERSION = tfds.core.Version("2.0.0")
```
us: You need to S3 update to false, because default value is true
like:
```VERSION = tfds.core.Version("1.0.0",
                              experiments={tfds.core.Experiment.S3: False})```
About the S3 update please check my [pr](https://github.com/tensorflow/datasets/pull/870)

PR #329, opened 2019-03-24
Reference added for "as_dataset_kwargs"

gshashank84:
Added reference for the argument **as_dataset_kwargs** in ```tfds.load```

for an example -
```
train_set = tfds.load("tf_flowers", as_supervised=True, as_dataset_kwargs={'shuffle_files':True})
```


Conchylicultor:
Thanks for clarifying the doc

PR #372, opened 2019-03-28
add Oxford-IIIT Pet dataset with image-level semantic labels only.

tfds-copybara:
add Oxford-IIIT Pet dataset with image-level semantic labels only.


cyfra:
obsolete - code submitted in other pull request.

PR #374, opened 2019-03-28
Updated docs for DatasetBuilder

gshashank84:
Added Doc for info of the ```tfds.builder("dataset_name").info```


rsepassi:
Thank you, but I think for this one we'll leave things as is. The info object is accessed on the line right above. If you'd like to contribute, please do take a look at our issues and work on one that looks interesting to you!

Issue #310, opened 2019-03-21
Option to turn tqdm progress bar off


sfujiwara:
**Is your feature request related to a problem? Please describe.**

Progress bar by tqdm is useful when we run the code on local.
But, it a little bit bothering when we run the code on server or cloud.
So, I think an option to turn tqdm progress bar off is useful.

**Describe the solution you'd like**

This problem is discussed in tqdm GitHub issue, but it seems not to be resolved yet:
https://github.com/tqdm/tqdm/issues/619

I want a method as below:

```python
tfds.disable_progress_bar()
```

**Describe alternatives you've considered**

We can wait the update of tqdm, but we can't tell when the issue is resolved.

sfujiwara:
I think this problem is important because it seems that tqdm can not be used (it occurs an error) on Google Cloud ML Engine as long as I investigated.

Conchylicultor:
All `tqdm` instances used internally in `tfds` should use `tfds.utils.tqdm` so in theory, it would be possible to achieve this by replacing `tqdm` by a no-op class, when the user call `tfds.disable_progress_bar()`.

https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/utils/tqdm.py#L27

Some point to concider:
* Tqdm is sometimes used as context manager, sometimes as iterator.
* When used as context manager, tqdm return a pbar object with method called latter in the program (`pbar.total += n`, `pbar.refresh()`, ...)

us:
I want to solve that problem. Please assign me.

Conchylicultor:
It should be fixed with #376, which will be available in the next nightly.

PR #376, opened 2019-03-28
Added Feature: tqdm progress bar disable func.

us:
Related: Option to turn tqdm progress bar off #310


Conchylicultor:
This won't disable the download bar globally, no ? Only for download, but not when shuffling,...
Is what you wanted ?

us:
Yes I wanted that because I understood the feature request like that. So I disable just tqdm progress bar.
Maybe we should ask @sfujiwara.

sfujiwara:
In some environment such as Google Cloud ML Engine, it seems that tqdm (including shuffling) occurs an error and does not work.
Thus, I think wee need a feature to globally disable tqdm.

tensorflow_datasets/core/utils/tqdm.py
```diff
@@ -59,28 +76,32 @@ class _TqdmPbarAsync(object):
...
+    self._lock = tqdm_lib.tqdm.get_lock()
     self._pbar = pbar
     self._tqdm_bars.append(pbar)
 
   def update_total(self, n=1):
     """Increment total pbar value."""
-    with self._lock:
-      self._pbar.total += n
-      self.refresh()
+    if active:
```
us: @Conchylicultor these `if statements` was little bad, how can I solve that ?

Conchylicultor:
Thanks for your contribution. Merged it even if I slightly modified the implementation.

PR #378, opened 2019-03-28
Fixed error in example code for add_dataset script

nvnmo:
Updated the example code which contained a wrong flag for specifying dataset name while scaffolding a new dataset using create_new_dataset script.


nvnmo:
Signed CLA

Conchylicultor:
Thanks for the fix. Merging it now

Issue #232, opened 2019-03-12
Download dataset from Kaggle using dataset name


SoumyadeepJana:
**Is your feature request related to a problem? Please describe.**
Currently, the download manager only supports downloading the datasets which are associated with a competition on kaggle. There are many datasets on kaggle that don't have a competition name associated with them.

**Describe the solution you'd like**
Extend the download manager to add support for downloading datasets from kaggle using the name of the dataset.

rsepassi:
Could you link to an example of a competition dataset and a dataset not associated with a competition? Can they be unified? i.e. Do competition datasets also just have dataset names? Maybe we can only use dataset names and get rid of competition names altogether.

SoumyadeepJana:
@respassi Dataset not associated with a competition [HAM10000](https://www.kaggle.com/kmader/skin-cancer-mnist-ham10000)  Dataset associated with competition[Histopathologic Cancer Detection](https://www.kaggle.com/c/histopathologic-cancer-detection/data).I don't think they have different names as such for competition or the dataset itself,every purpose is served using one name.We can use only the name to download the dataset,maybe that would make it more robust,your views?

vijayphoenix:
This issue can be closed as PR #1504 got merged internally

PR #381, opened 2019-03-29
Add datasets support to kaggle downloader

us:
Related issue #232

### DIFF
--- a/tensorflow_datasets/core/download/kaggle.py
+++ b/tensorflow_datasets/core/download/kaggle.py
@@ -79,6 +79,8 @@ class KaggleCompetitionDownloader(object):
   """Downloader for a Kaggle competition.
 
   Usage:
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/core/download/kaggle.py
```diff
@@ -93,9 +95,14 @@ def __init__(self, competition_name):
   @utils.memoized_property
   def competition_files(self):
     """List of competition files."""
+    if '/' in self._competition_name:
+      type = 'datasets'
+    else:
+      type = 'competitions'
```
Conchylicultor: I'm not sure I understand. Where is `type` used ?

(Additionally `type` is a python built-in that your are redefining here.)

Conchylicultor:
Thank you for adding this. Merging it now.

PR #382, opened 2019-03-29
Fix format render issue

howl-anderson:
Fix format render issue (in https://www.tensorflow.org/datasets/api_docs/python/tfds/features/ClassLabel)


Conchylicultor:
Thanks for your contribution, merging it now.
I've done the fix for you but in the future please note that this file is automatically generated from the python docstring, so you have to modify the original `class_label_features.py` file.

howl-anderson:
@Conchylicultor Thank you! I will modify correct files next time.

PR #413, opened 2019-04-03
FeaturesDict.md: use output instead of target

galeone:
The first example initially describes the features using the keys `input`, `target` when describing the DatasetInfo, but then (for generation time and tf.data.Dataset() time) uses `output` instead of `target`.


Conchylicultor:
Thanks for your contribution. Please note that those files are automatically generated. You should be modifying the original python docstring instead.

galeone:
Thank you for pointing that out! I'll update the original Python file then
(as soon as I can)

galeone:
I created a new pr ( #420 )

Issue #415, opened 2019-04-03
Documentation: Bair robot pushing dataset


oscartackstrom:
The documentation for the "bair_robot_pushing_dataset" seems wrong. From the dataset statistics it looks like this is the "bair robot pushing dataset" from https://sites.google.com/corp/view/sna-visual-mpc/. But the documentation mixes the description of the bair dataset with the "google robot pushing dataset" from https://sites.google.com/site/brainrobotdata/home/push-dataset.

us:
@oscartackstrom Thank you for reporting. I fixed it #416

PR #416, opened 2019-04-03
Correct Bair robot pushing dataset citation and url

us:
Related pr #415


us:
I think we should close because already tfds-copybara merge that.

PR #445, opened 2019-04-09
 example checksum: handle nested objects

galeone:
The `checksum` function used during the Dataset Builder tests right now works only on examples that are single level dictionaries; this PR makes it able to work with nested objects of every type.


Conchylicultor:
Thank you for your contribution. Could you add a test for this ?

galeone:
Yes, all the tests use this feature :+1:

PR #447, opened 2019-04-09
Add kaggle downloader extension to jupyter notebook

us:
Kaggle has a very extensive set of useful and interesting datasets. I designed and implemented graphical user interface functionality embedded in a jupyter notebook that will enable a user to list, filter, select and automatically download Kaggle datasets. 

(It's a demo version)
I will fix it as soon as possible.

### DIFF
--- a/tensorflow_datasets/core/download/kaggle.py
+++ b/tensorflow_datasets/core/download/kaggle.py
@@ -22,7 +22,6 @@ from __future__ import print_function
 import os
 import subprocess as sp
 
-
 from absl import logging
 import tensorflow as tf
 
--- a/tensorflow_datasets/core/download/kaggle_search.py
+++ b/tensorflow_datasets/core/download/kaggle_search.py
@@ -22,6 +22,7 @@ from __future__ import print_function
 import ipywidgets.widgets as widgets
 import IPython.display as display
 
+import os
 import tensorflow as tf
 
 from tensorflow_datasets.core.download import kaggle
@@ -184,7 +185,9 @@ def interactive_kaggle():
       if not downloader.competition_files:
         raise DataNotFoundError('Data is no longer available!')
       for fname in downloader.competition_files:
-        downloader.download_file(fname, dataset_name.value)
+        downloader.download_file(fname, os.path.expanduser(
+                                        os.path.join("~", "tensorflow_datasets",
+                                                     dataset_name.value)))
         print('Downloaded ', fname)
       print('Download Completed!')
 


dynamicwebpaige:
This is very cool! Thank you so much for your contribution - would love to see a Kaggle downloader extension. Could you please make sure to add some tests for this feature? 🙂

us:
Thank you !
 I will solve other errors and add as soon as possible.

us:
I checked kokoro, that's given error on `../oss_scripts/oss_tests.sh: line 45:  9848 Aborted                 (core dumped)`. I didn't understand what I did wrong. Can you check it ?

Thank you !

tensorflow_datasets/core/download/kaggle.py
```diff
@@ -151,3 +151,83 @@ def _run_kaggle_command(command_args, competition_name):
...
+    print(a.searching())
+
+    dataset_name = widgets.Text(placeholder='Dataset Name', description='Download:')
+    display(dataset_name)
+    download_button = widgets.Button(description="Download",
+                              style=widgets.ButtonStyle(button_color='lightblue'))
+    display(download_button)
+
+    def kag_downloader(sender):
+      downloader = KaggleCompetitionDownloader(dataset_name.value)
```
pierrot0: Where is KaggleCompetitionDownloader symbol coming from?

tensorflow_datasets/core/download/kaggle.py
```diff
@@ -151,3 +151,83 @@ def _run_kaggle_command(command_args, competition_name):
 def _log_command_output(output, error=False):
   log = logging.error if error else logging.info
   log("kaggle command output:\n%s", tf.compat.as_text(output))
+
+
+class KaggleSearch(object):
```
pierrot0: Could you move all code which is to be used from jupiter notebook to a different module please?

pierrot0:
Thanks Recep! This is very cool.

Would you mind writing a short document on how this feature would be used?
Similarly to the small video you posted, but from which people can copy/paste and try themselves.

Would there be a way to go one step further and have the UI generate the DatasetBuilder definition for the dataset? (or some skeleton the user can complete).
Contributing would feel a lot less scary for people. Have a submitted Builder definition has some benefit (reproducible experiments, easier to open-source code using the dataset, ...).

tensorflow_datasets/core/download/kaggle.py
```diff
@@ -144,10 +157,175 @@ def _run_kaggle_command(command_args, competition_name):
...
+    if self.where != 'competitions':
+      # Fit to the dataset table.
+      display.display(display.HTML("<style>.container { width:80% !important; }"
+                                   "</style>"))
+
+    output = _run_kaggle_command(command)
+    return output
+
+
+def search_and_download():
```
Conchylicultor: I would put "interactive" or "colab" or "notebook" in the name to make it more clear that it is intended to use in a notebook.

Conchylicultor:
This is nice thanks.

Concerning what @pierrot0 said about a better integration with tfds, we also had a plan to create a generic `KaggleBuilder` that people could use to easily create `tfds` datasets from the `Kaggle`.
Something like:

```
builder = tfds.core.KaggleBuilder('kaggle_dataset_name')
builder.download_and_prepate()
ds = builder.as_dataset(split='train')
```

It probably won't works with every dataset but because it seems that most kaggle datasets are using a standard `.csv` format, it may be possible to automatically infer the fields, split names,... of the data.

But unfortunately no one had time yet to look into it or even to investigate the feasibility.

tensorflow_datasets/core/download/kaggle.py
us: From the class, on the line 91.

tensorflow_datasets/core/download/kaggle.py
us: I thought that but I couldn't decide. Thank you! I will move to the new file

us:
Yeap I thought exactly like you! And I thought of such a structure for the further goals of the project. Easy dataset implementation! Maybe other next goals is automatic implementation to same data structures dataset. Because after a certain time, data structures be fall into repetition then become the same

Issue #440, opened 2019-04-07
[data request] GAP data set


sklan:
* Name of dataset: GAP
* URL of dataset: https://github.com/google-research-datasets/gap-coreference
* License of dataset: Apache License 2.0
* Short description of dataset and use case(s): GAP is a gender-balanced dataset containing 8,908 coreference-labeled pairs of (ambiguous pronoun, antecedent name), sampled from Wikipedia for the evaluation of coreference resolution in practical applications.
ArXiV Link: [https://arxiv.org/abs/1810.05201](https://arxiv.org/abs/1810.05201)
Folks who would also like to see this dataset in `tensorflow/datasets`, please thumbs-up so the developers can know which requests to prioritize.

And if you'd like to contribute the dataset (thank you!), see our [guide to adding a dataset](https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md).

sklan:
This is another data set I want to add. Can somebody please assign me?

shashvatshahi1998:
Can I add it @sklan please I am new to it.

sklan:
@shashvatshahi1998 I am actually mostly done already. However, you can find interesting datasets here:
[https://ai.google/tools/datasets/](https://ai.google/tools/datasets/)

cyfra:
Merged in #460

PR #460, opened 2019-04-11
Add the gap data set

sklan:
#440 
I added the gap data set
Gist: https://gist.github.com/sklan/5c6371b8a276c6f0aa4eda299b1826c6


us:
@Conchylicultor @cyfra I think it's okay.

PR #468, opened 2019-04-13
Added Dep of Tensorflow_io.lmdb  to LSUN

### LINKED ISSUE #34: Rm dep on tf.contrib.data.LMDBDataset for TF 2.0
tf.contrib.data.LMDBDataset will not make it to TF2. It may be moving to a new repo/package. If/when that happens, we should switch to it.

Part of #31 (TF 2.0 support)

ParthS007:
Fixes #34 
@rsepassi @Conchylicultor @cyfra Please review. Thanks :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,6 +95,7 @@ DATASET_EXTRAS = {
     'librispeech': ['pydub'],  # and ffmpeg installed
     'svhn': ['scipy'],
     'wikipedia': ['mwparserfromhell', 'apache_beam'],
+    'lsun': ['tensorflow-io']
 }
 
 all_dataset_extras = []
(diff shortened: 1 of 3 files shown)


ParthS007:
@tfds-copybara Can you please tell why Kokoro is failing?

us:
Kokoro run test codes using different versions of TF and python. Versions earlier than TF 2.0 give errors. So Kokoro failed. 
You need to check version of TF and if version greater than 2.0 you can call tf_io otherwise you need to call old structure.

ParthS007:
> You need to check version of TF and if version greater than 2.0 you can call tf_io otherwise you need to call old structure.

@us How can I do it?

us:
You can check `core/tf_combat.py` to compatibility tf versions.

us:
@ParthS007 hey did you check it ?

PR #535, opened 2019-05-03
Add Freesound Audio Tagging 2019

carlthome:
I'd love to see tfds include [the DCASE task that's on Kaggle currently](https://www.kaggle.com/c/freesound-audio-tagging-2019/), but I didn't find any issues or pull requests for adding it so here's my attempt. I'm opening a pull request early so we can pool efforts, in case anyone else would like to make this happen! :koala: 

I still have some reading of the contribution guidelines to do (on it!). Adding `pandas`, and `kaggle` for example, code formatting, unit test framework used by tfds, and so on are on my TODO.

## Loading audio?
`tfds.features.Audio` is really slow (it seems to do single threaded calls to pydub regardless of minibatch size), and I noticed that @adarob's Groove addition just now didn't go via the existing feature connector at all. Is stuff intended to move around? What encoding is preferred for audio datasets? 

Currently this pull request just lets the user consume the file path to the WAV file manually, by returning a `tf.string` with the path to audio file, then a subsequent `tf.io.read_file` and `tf.audio.decode_wav` call can be parallelised in `tf.data.Dataset.map`.

## How to encode multi-label datasets?
Throughout tfds there doesn't seem to be a strong convention for how to encode multi-label datasets, and there doesn't seem to be a feature connector for binary indicator variables yet. I copied the encoding used by the NSynth dataset, where a nested `FeaturesDict` simply contains individual `tf.bool`:s per label, but then it's up to the user to concatenate them which can be bug prone.

## Splitting the dataset
Since the competition is interested in training on noisy labels, and evaluating on curated labels, I think we should default to splitting the dataset in exactly that manner despite the fact that you'd probably rank higher on the Kaggle leaderboard by training on some of the curated data (as that's the same source that's used for the hidden test set).

Another discussion point is how to treat the fact that the test dataset doesn't have any labels. Currently, I'm just returning whatever [0.0,1.0] scores are set in the example_submission.csv provided by the competition.

### DIFF
--- a/tensorflow_datasets/audio/freesound_audio_tagging_2019.py
+++ b/tensorflow_datasets/audio/freesound_audio_tagging_2019.py
@@ -1,0 +1,152 @@
+from pathlib import Path
+
+import kaggle
+import pandas as pd
+import tensorflow as tf
+import tensorflow_datasets as tfds
+
+_TAGS = [
+    'Child_speech_and_kid_speaking', 'Cutlery_and_silverware',
+    'Marimba_and_xylophone', 'Hi-hat', 'Slam', 'Stream', 'Shatter',
+    'Mechanical_fan', 'Cupboard_open_or_close',
+    'Traffic_noise_and_roadway_noise', 'Sink_(filling_or_washing)',
+    'Female_singing', 'Raindrop', 'Zipper_(clothing)', 'Printer', 'Writing',
+    'Chewing_and_mastication', 'Bass_drum', 'Sigh', 'Trickle_and_dribble',
+    'Walk_and_footsteps', 'Church_bell', 'Buzz', 'Cricket',
+    'Computer_keyboard', 'Tick-tock', 'Male_singing',
+    'Female_speech_and_woman_speaking', 'Bass_guitar', 'Applause',
+    'Water_tap_and_faucet', 'Clapping', 'Microwave_oven',
+    'Drawer_open_or_close', 'Yell', 'Chirp_and_tweet', 'Tap', 'Cheering',
+    'Squeak', 'Screaming', 'Whispering', 'Harmonica', 'Finger_snapping',
+    'Crowd', 'Frying_(food)', 'Purr', 'Acoustic_guitar', 'Chink_and_clink',
+    'Glockenspiel', 'Run', 'Bus', 'Motorcycle', 'Hiss', 'Toilet_flush',
+    'Strum', 'Scissors', 'Male_speech_and_man_speaking', 'Gurgling',
+    'Dishes_and_pots_and_pans', 'Fart', 'Race_car_and_auto_racing',
+    'Waves_and_surf', 'Drip', 'Gong', 'Gasp',
+    'Accelerating_and_revving_and_vroom', 'Burping_and_eructation',
+    'Keys_jangling', 'Skateboard', 'Crackle', 'Electric_guitar',
+    'Bicycle_bell', 'Sneeze', 'Knock', 'Meow', 'Bark',
+    'Bathtub_(filling_or_washing)', 'Accordion', 'Fill_(with_liquid)',
+    'Car_passing_by'
+]
+
+# TODO
+_CITATION = '''\
+
+'''
+
+# TODO
+_DESCRIPTION = '''\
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/audio/freesound_audio_tagging_2019.py
```diff
@@ -0,0 +1,152 @@
+from pathlib import Path
+
+import kaggle
```
us: TFDS also have a kaggle downloader on `core/download/kaggle.py`. Please check it.

tensorflow_datasets/audio/freesound_audio_tagging_2019.py
```diff
@@ -0,0 +1,152 @@
...
+            tfds.Split.TEST: download_dir / 'sample_submission.csv',
+        }
+
+        zips = {
+            tfds.Split.TRAIN: download_dir / 'train_noisy.zip',
+            tfds.Split.VALIDATION: download_dir / 'train_curated.zip',
+            tfds.Split.TEST: download_dir / 'test.zip',
+        }
+
+        kaggle.api.authenticate()
```
us: .

tensorflow_datasets/audio/freesound_audio_tagging_2019.py
```diff
@@ -0,0 +1,152 @@
+from pathlib import Path
```
us: Tensorflow have own io library. Please check this [link](https://www.tensorflow.org/api_docs/python/tf/io).

tensorflow_datasets/audio/freesound_audio_tagging_2019.py
```diff
@@ -0,0 +1,152 @@
...
+            urls=_URLS,
+            citation=_CITATION,
+        )
+        return dataset_info
+
+    def _split_generators(self, dl_manager):
+        download_dir = Path(dl_manager._download_dir)
+
+        csvs = {
+            tfds.Split.TRAIN: download_dir / 'train_noisy.csv',
```
us: I think you miss quotes.

Use `os` library to be compatible with all environments.
Like that `os.path.join('download_dir', 'train_noisy.csv')`

tensorflow_datasets/audio/freesound_audio_tagging_2019_test.py
```diff
@@ -0,0 +1,74 @@
...
+        x = tf.transpose(x, (1, 2, 0))
+        spectrogram = tf.abs(x)
+
+        tags = row['tags']
+        tags = tf.stack(list(tags.values()))
+
+        return spectrogram, tags
+
+    dataset = tfds.load(
+        name='freesound_audio_tagging2019',
```
us: You should use fake examples to testing. Check this [link](https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md#testing-mydataset).

us:
Nice dataset.
Thanks for contributing.

tensorflow_datasets/audio/freesound_audio_tagging_2019.py
Conchylicultor: The Path library is nice but I am worried with Python2 compatibility.

If you want to use it, you would have to add it to the Python2 dependency list: https://github.com/tensorflow/datasets/blob/master/setup.py#L68

tensorflow_datasets/audio/freesound_audio_tagging_2019.py
```diff
@@ -0,0 +1,152 @@
+from pathlib import Path
+
+import kaggle
+import pandas as pd
```
Conchylicultor: Please have a look at lazy_imports for this: https://www.tensorflow.org/datasets/add_dataset#extra_dependencies

Conchylicultor:
Thank you for your contribution, for the audio encoding, I feel the correct solution would be to do a similar implementation as for `Image`, where the audio file is encoded as string inside the tf record file, and decoded with `tf.audio.decode_wav` inside `tfds.features.Audio.decode_example`.

Like: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/features/image_feature.py#L149

us:
Hey @carlthome, can you check @Conchylicultor pr review

Thanks again!

carlthome:
Closing due to stale

PR #548, opened 2019-05-08
Removes small typo from EMNIST dataset description

miykael:


miykael:
I signed it!

Conchylicultor:
Thank you very much for your contribution.
Please note that this file is automatically generated from the docstring and modifications should be applied to the original `.py` file instead.

miykael:
Thank you for the feedback! Modifications were now also done in the original `.py`file.

Issue #592, opened 2019-05-23
pip install does not include '.txt' label files required to 


luke-zhu:
I am unable to load datasets sun397 and caltech101 because `pip install tensorflow_datasets` does not copy the needed TXT files into `site-packages`. This is related to issue #157.

Taking the measures to include non-code files could solve this issue. See https://python-packaging.readthedocs.io/en/latest/non-code-files.html

**Environment information**
* Operating System: both MacOS + the Python 3 Docker image run on Linux
* Python version: 3.7
* `tensorflow-datasets`/`tfds-nightly` version: 1.0.2
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: 1.13.1

**Reproduction instructions**

1. Run `pip install tensorflow_datasets`
2. Run either `tensorflow_datasets.load('caltech101')` or tensorflow_datasets.builder('caltech101').

Both result in an error like the following:
`NotFoundError: ../lib/python3.7/site-packages/tensorflow_datasets/image/caltech101_labels.txt; No such file or directory`

**Link to logs**
If applicable, <link to gist with logs, stack trace>

**Expected behavior**
The commands should not fail. All needed TXT files should be downloaded to the right locations after I run `pip install`.

**Additional context**

jackd:
I came across this issue with my own forks. All non-`.py` files need to be specified in [DATASET_FILES](https://github.com/tensorflow/datasets/blob/master/setup.py#L78).

jackd:
Looks to be resolved in tfds-nightly

luke-zhu:
Awesome thanks!

PR #595, opened 2019-05-24
added missing DATASET_FILES

jackd:
As raised [here](https://github.com/tensorflow/datasets/issues/592)


Conchylicultor:
Thank you for your contribution, I'm merging this now.
(note that it may take a few days before being reflected into the public repository due to some internal synchronization issues)

PR #598, opened 2019-05-27
Added open_images_v4 class files to DATASETS_FILES

areche:


ChanchalKumarMaji:
There is already an existing PR #595 to this issue.

Issue #45, opened 2019-02-08
[data request]  Downsampled ImageNet


rodrigob:
* Name of dataset:  Downsampled ImageNet
* URL of dataset: http://image-net.org/small/download.php
* License of dataset: Same as ImageNet
* Short description of dataset and use case(s): great for prototyping and quick experimentation beyond MNIST.

Folks who would also like to see this dataset in `tensorflow/datasets`, please +1/thumbs-up so the developers can know which requests to prioritize.

tabshaikh:
As this is a good first issue would like to take this up 😀

tabshaikh:
There are 2 version of this dataset 32*32 images and 64*64 images should both be done ?

rsepassi:
That's great @tabshaikh, thank you! 

Yes, I think we should have both using ["heavy" configuration](https://www.tensorflow.org/datasets/add_dataset#dataset_configuration), but you can start with just one.

rsepassi:
I'll assign the issue to you as soon as you accept the collaborator invite! :smiley:

tabshaikh:
Sure thanks :)

rsepassi:
Sounds good! Thank you!

Anupam-tripathi:
@rsepassi I would love to collaborate with @tabshaikh and make a pull request for Downsampled ImageNet. Please assign it to me also.

rsepassi:
Hi @Anupam-tripathi, thanks for your interest! Are you already working directly with @tabshaikh? If not, let's give him a chance to get a PR in. If he'd like the help, then please do work together to get something in!

Anupam-tripathi:
No, I have not joined him till now but will surely contact him personally.

tabshaikh:
@Anupam-tripathi I would love to collaborate with you but the pr is almost done with a few changes left to do and hopefully, I have added the dataset correctly @rsepassi  I would do a pr till mid next week as I will be going for an ML hackathon during the weekend, I had some question too would ask in the draft pr
Cheers :)

rsepassi:
Sounds good. Thanks!

Anupam-tripathi:
Ya, surely I will prove to be a good teammate.

tabshaikh:
@rodrigob @rsepassi the link for http://image-net.org/small/download.php does not contain the whole dataset of the downsampled imagenet nor does it contain the labels. 
[...]

PR #613, opened 2019-05-29
Added Downsampled ImageNet dataset.

ChanchalKumarMaji:
Related issue #45.


us:
@ChanchalKumarMaji you can squash commits that's have unnecessary commit messages like merging commit, faulty commit etc.

tensorflow_datasets/image/downsampled_imagenet.py
```diff
@@ -0,0 +1,136 @@
...
+from __future__ import print_function
+
+import os
+
+import tensorflow as tf
+
+from tensorflow_datasets.core import api_utils
+import tensorflow_datasets.public_api as tfds
+
+# From https://arxiv.org/abs/1601.06759
```
us: I think it's unnecessary.

ChanchalKumarMaji: Actually I followed [this](https://github.com/tensorflow/datasets/blob/90c0b38438afcb56044f13b11a426d013d0c8848/tensorflow_datasets/image/cycle_gan.py#L29).

us: Already written [there](https://github.com/tensorflow/datasets/blob/90c0b38438afcb56044f13b11a426d013d0c8848/tensorflow_datasets/image/cycle_gan.py#L41) so it be written twice. It's also unnecessary.

tensorflow_datasets/image/downsampled_imagenet.py
```diff
@@ -0,0 +1,136 @@
...
+_CITATION = """\
+@article{DBLP:journals/corr/OordKK16,
+  author    = {A{\"{a}}ron van den Oord and
+               Nal Kalchbrenner and
+               Koray Kavukcuoglu},
+  title     = {Pixel Recurrent Neural Networks},
+  journal   = {CoRR},
+  volume    = {abs/1601.06759},
+  year      = {2016},
+  url       = {http://arxiv.org/abs/1601.06759},
```
us: url already written in there

tensorflow_datasets/image/downsampled_imagenet.py
ChanchalKumarMaji: Okk, I see.

tensorflow_datasets/image/downsampled_imagenet.py
ChanchalKumarMaji: Done.

PR #618, opened 2019-05-30
Fixed quickdraw_bitmap URLs in docs

tfujiwar:
The URLs of `quickdraw_bitmap` dataset in document was not displayed correctly.

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -4855,14 +4855,14 @@ TEST       |      3,669
   <meta itemprop="name" content="quickdraw_bitmap" />
   <meta itemprop="description" content="The Quick Draw Dataset is a collection of 50 million drawings across 345 categories, contributed by players of the game Quick, Draw!. The bitmap dataset contains these drawings converted from vector format into 28x28 grayscale images" />
   <meta itemprop="url" content="https://www.tensorflow.org/datasets/datasets#quickdraw_bitmap" />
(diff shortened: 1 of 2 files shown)


ChanchalKumarMaji:
Make sure that you generated `datasets.md` using [document_datasets.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/document_datasets.py).

tfujiwar:
@ChanchalKumarMaji 
Only the updated section in `datasets.md` was generated using `document_datasets.py` by specifying the name of the dataset in the script.

Because I failed to execute `document_datasets.py` with the error below. Should I fix this first?
```
...
I0530 20:22:02.297413 4781708736 dataset_info.py:398] Loading info from GCS for super_glue/copa/0.0.2
E0530 20:22:03.379710 4781708736 registered.py:171] Failed to construct dataset super_glue
...
ValueError: Trying to overwrite already defined ClassLabel names. Previous: ['choice1', 'choice2'] , new: ['False', 'True']
```

ChanchalKumarMaji:
Okk, if the results are in accordance with your expectations then you need not generate the whole file. They will be updated by the community members.

Conchylicultor:
Thanks for fixing this. Note that the docs will automatically updated, so no worries about that.
Thanks for reporting the error, it is due to a mismatch between our GCS version and code.

PR #627, opened 2019-06-03
Add gsutil support to downloader.

us:
This is just the draft I wrote to better explain my mind. We can discuss on this code to better understand.


ChanchalKumarMaji:
Any `test_file` for this ?

us:
I'll add after the after confirmation of this feature.

pierrot0:
Thanks!

Overall, this looks like what we want.
We would want however to make sure the gsutil binary is available (or otherwise fall back on standard url handling).

Have you checked kaggle doesn't provide an API instead of using external binary?
Is the usage similar on windows?

Conchylicultor:
Clsoing this as the PR seems abandoned. Please re-open the PR if you want to continue this

PR #635, opened 2019-06-04
Multi-names support for ClassLabel.

ChanchalKumarMaji:
Initial commit for `Multi-names support for ClassLabel`.
The main idea here is to expose the multiple ways of labelling as seen in `imagenet`, `cifar100`, etc..

Please do suggest changes or better ideas for the improvement.

### DIFF
--- a/tensorflow_datasets/core/features/class_label_feature.py
+++ b/tensorflow_datasets/core/features/class_label_feature.py
@@ -8,7 +8,7 @@
 #     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distribsuted on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/features/class_label_feature.py
```diff
@@ -50,6 +53,7 @@ def __init__(self, num_classes=None, names=None, names_file=None):
     self._num_classes = None
     self._str2int = None
     self._int2str = None
+    self._is_multilabel = False
```
Conchylicultor: If the user load labels from filename, you should set this to True. By checking the extension for instance.

```
features={
    'label': ClassLabel(names_file='label.csv'),
}
```
otherwise you will try to load the `.txt` file in `_load_names_from_file`

PR #642, opened 2019-06-05
Improve usability of `batch_size` in `as_dataset`.

ChanchalKumarMaji:
Currently, in the `as_dataset` function, `batch_size=1` won't add a batch dimension while `batch_size=2` will. This is inconsistent for the users as passing the `batch_size` as flag will modify the tensors shape.
Sol: Have `batch_size=None` (default) do not return batch_dimension while `batch_size=1` does.

```
>>> import tensorflow_datasets as tfds
>>> 
>>> ds=tfds.load('mnist', split='train', batch_size=None)
>>> ds
<DatasetV1Adapter shapes: {image: (28, 28, 1), label: ()}, types: {image: tf.uint8, label: tf.int64}>
>>> 
>>> ds=tfds.load('mnist', split='train', batch_size=1)
>>> ds
<DatasetV1Adapter shapes: {image: (?, 28, 28, 1), label: (?,)}, types: {image: tf.uint8, label: tf.int64}>
>>> 
>>> ds=tfds.load('mnist', split='train', batch_size=2)
>>> ds
<DatasetV1Adapter shapes: {image: (?, 28, 28, 1), label: (?,)}, types: {image: tf.uint8, label: tf.int64}>
```

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -298,7 +298,7 @@ class DatasetBuilder(object):
   @api_utils.disallow_positional_args
   def as_dataset(self,
                  split=None,
-                 batch_size=1,
+                 batch_size=None,
                  shuffle_files=None,
                  as_supervised=False):
     """Constructs a `tf.data.Dataset`.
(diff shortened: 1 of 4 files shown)


Conchylicultor:
Looks good thanks.
Just add a simple test and its good to go :)

Also `batch_size` should also be modified in `tfds.load`, and maybe at other places too.

ChanchalKumarMaji:
> Just add a simple test and its good to go :)

A simple test would be to check the shape of the `image: (?, 28, 28, 1)`, but how to get that ?

Conchylicultor:
You can test the `tf.data.Dataset` output shape:
https://www.tensorflow.org/api_docs/python/tf/data/Dataset#output_shapes

PR #657, opened 2019-06-09
Add a bigearthnet subsest with snow and clouds labels and a clean subset.

tfds-copybara:
Add a bigearthnet subsest with snow and clouds labels and a clean subset.


cyfra:
submitted.

Issue #275, opened 2019-03-18
option to be be able to use datasets behind a proxy


tarrade:
**Is your feature request related to a problem? Please describe.**
Right now behing a proxy, it is not working:
```

ds_train = tfds.load(name="cats_vs_dogs", split=tfds.Split.TRAIN)

C:\Program Files\Anaconda3\envs\env_gcp_dl_2_0_ds\lib\site-packages\requests\adapters.py in send(self, request, stream, timeout, verify, cert, proxies)
    514                 raise SSLError(e, request=request)
    515
--> 516             raise ConnectionError(e, request=request)
    517
    518         except ClosedPoolError as e:
ConnectionError: HTTPConnectionPool(host='storage.googleapis.com', port=80): Max retries exceeded with url: /tfds-data/ (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x00000192F3D06668>: Failed to establish a new connection: [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond',))
```
I don't think this is supported for now (I didn't see it in the documentation):
https://www.tensorflow.org/datasets/api_docs/python/tfds/load

This will impact quite a lot of people working in company and university

**Describe the solution you'd like**
I am not an expert but using `request` seems to be the standard way. Below on example from a Google GCP tool:

```
from google.cloud import storage
import os

os.environ['GOOGLE_APPLICATION_CREDENTIALS']=xxx
os.environ['HTTPS_PROXY']=xxx
os.environ['REQUESTS_CA_BUNDLE']=/xxx/xxx
client = storage.Client()
```
ignore the GOOGLE_APPLICATION_CREDENTIALS' whihc is specific to GCP. The user need to setup one or 2 env variables and everything is done in the backgroud (I guess this is using requests)

http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

captain-pool:
@rsepassi @Conchylicultor @cyfra Is this issue fixed? if not, can you assign this to me?

Conchylicultor:
It should be fixed with @captain-pool contribution #488

tarrade:
Thanks @captain-pool 

"To Configure the Proxy Settings, The User needs to set the Proxies for HTTP, HTTPS and FTP in the Environment Variables
TFDS_HTTP_PROXY, TFDS_HTTPS_PROXY, TFDS_FTP_PROXY respectively."

[...]

PR #663, opened 2019-06-11
Added Support for CA Files in Downloader #275

captain-pool:
Patch in reference to #275 
Added Support to CA Files, in Downloader.
The User needs to set an environment variable `TFDS_CA_BUNDLE`. Not Setting anything will skip any CA Checking.

cc: @Conchylicultor @rsepassi @cyfra

### DIFF
--- a/tensorflow_datasets/core/download/downloader.py
+++ b/tensorflow_datasets/core/download/downloader.py
@@ -186,8 +186,7 @@ class _Downloader(object):
       if proxies['ftp']:
         proxy = urllib.request.ProxyHandler({'ftp': proxies['ftp']})
         opener = urllib.request.build_opener(proxy)
-        urllib.request.install_opener(
-            opener)   # pylint: disable=too-many-function-args
+        urllib.request.install_opener(opener)   # pylint: disable=too-many-function-args
       request = urllib.request.Request(url)
 
       if CA_VERIFY['urllib'] is None:     # disable ssl context check for python version <= 2.7.8
--- a/tensorflow_datasets/core/download/downloader.py
+++ b/tensorflow_datasets/core/download/downloader.py
@@ -24,10 +24,11 @@ import hashlib
 import io
 import os
 import re
+import ssl
 import concurrent.futures
 import promise
 import requests
-
+from requests.utils import extract_zipped_paths
 from six.moves import urllib
 
 import tensorflow as tf
@@ -148,6 +149,14 @@ class _Downloader(object):
         'https': os.environ.get('TFDS_HTTPS_PROXY', None),
         'ftp': os.environ.get('TFDS_FTP_PROXY', None)
     }
+    CA_BUNDLE = os.environ.get('TFDS_CA_BUNDLE', None)
+    if CA_BUNDLE:
+      CA_BUNDLE = extract_zipped_paths(CA_BUNDLE)
+    CA_VERIFY = {
+        'urllib':  ssl._create_unverified_context() if CA_BUNDLE is None 
+                   else ssl.create_default_context(capath=CA_BUNDLE),
+        'requests': False if CA_BUNDLE is None else CA_BUNDLE
+    }
     if kaggle.KaggleFile.is_kaggle_url(url):
       if proxies['http']:
         os.environ['KAGGLE_PROXY'] = proxies['http']
@@ -163,6 +172,7 @@ class _Downloader(object):
 
     session = requests.Session()
     session.proxies = proxies
+    session.verify = CA_VERIFY['requests']
     if _DRIVE_URL.match(url):
       url = self._get_drive_url(url, session)
     use_urllib = url.startswith('ftp')
@@ -172,7 +182,7 @@ class _Downloader(object):
         opener = urllib.request.build_opener(proxy)
         urllib.request.install_opener(opener)   # pylint: disable=too-many-function-args
       request = urllib.request.Request(url)
-      response = urllib.request.urlopen(request)
+      response = urllib.request.urlopen(request, context=CA_VERIFY['urllib'])
     else:
       response = session.get(url, stream=True)
       if response.status_code != 200:
--- a/tensorflow_datasets/core/download/downloader.py
+++ b/tensorflow_datasets/core/download/downloader.py
@@ -150,13 +150,25 @@ class _Downloader(object):
         'ftp': os.environ.get('TFDS_FTP_PROXY', None)
     }
     CA_BUNDLE = os.environ.get('TFDS_CA_BUNDLE', None)
+    
+    if not hasattr(ssl, "_create_unverified_context"):
+      # For python Version <= 2.7.16
+      def py2_fn():
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+        return ctx
+      ssl.__dict__["_create_unverified_context"] = py2_fn
+
     if CA_BUNDLE:
       CA_BUNDLE = extract_zipped_paths(CA_BUNDLE)
+    
     CA_VERIFY = {
         'urllib': ssl._create_unverified_context() if CA_BUNDLE is None
         else ssl.create_default_context(capath=CA_BUNDLE),
         'requests': False if CA_BUNDLE is None else CA_BUNDLE
     }
+    
     if kaggle.KaggleFile.is_kaggle_url(url):
       if proxies['http']:
         os.environ['KAGGLE_PROXY'] = proxies['http']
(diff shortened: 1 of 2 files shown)


captain-pool:
Disabled ssl skipping for python version <= 2.7.8
cc: @cyfra  @rsepassi @Conchylicultor take a look.

tensorflow_datasets/core/download/downloader.py
```diff
@@ -148,21 +149,39 @@ def _sync_download(self, url, destination_path):
         'https': os.environ.get('TFDS_HTTPS_PROXY', None),
         'ftp': os.environ.get('TFDS_FTP_PROXY', None)
     }
+    CA_BUNDLE = os.environ.get(
```
cyfra: Please add documentation on how this feature should be used (what things to set etc).

as I guess this is not only the matter of CA_BUNDLE, but user also needs to set the proxies, right ?

tensorflow_datasets/core/download/downloader.py
```diff
@@ -148,21 +149,39 @@ def _sync_download(self, url, destination_path):
...
+    if CA_BUNDLE:
+      CA_BUNDLE = extract_zipped_paths(CA_BUNDLE)
+    if not hasattr(ssl, '_create_unverified_context'):
+      # For python version <= 2.7.8
+      ssl.__dict__['_create_unverified_context'] = lambda: None
+      ssl.__dict__['create_default_context'] = lambda *args, **kwargs: None
+    CA_VERIFY = {
+        'urllib': ssl._create_unverified_context() if CA_BUNDLE is None or len(CA_BUNDLE) == 0
+        else ssl.create_default_context(capath=CA_BUNDLE),
+        'requests': False if CA_BUNDLE is None else CA_BUNDLE
```
cyfra: what is the type of 'requests' ?

I think you wanted it to be bool, but here it is either False, or path to the bundle, right ? Consider changing to:

'requests': CA_BUNDLE is None

tensorflow_datasets/core/download/downloader.py
```diff
@@ -148,21 +149,39 @@ def _sync_download(self, url, destination_path):
...
+        'TFDS_CA_BUNDLE',
+        None) or os.environ.get(
+        'REQUESTS_CA_BUNDLE',
+        None)
+    if CA_BUNDLE:
+      CA_BUNDLE = extract_zipped_paths(CA_BUNDLE)
+    if not hasattr(ssl, '_create_unverified_context'):
+      # For python version <= 2.7.8
+      ssl.__dict__['_create_unverified_context'] = lambda: None
+      ssl.__dict__['create_default_context'] = lambda *args, **kwargs: None
```
cyfra: In such case - we should be printing a warning inside.

So that users are not surprised that the certificates are not used.

tensorflow_datasets/core/download/downloader.py
```diff
@@ -148,21 +149,39 @@ def _sync_download(self, url, destination_path):
...
+    CA_VERIFY = {
+        'urllib': ssl._create_unverified_context() if CA_BUNDLE is None or len(CA_BUNDLE) == 0
```
cyfra: "not X" takes care of X is None or len(X) == 0.

'urllib': ssl.create_default_context(CA_BUNDLE) if CA_BUNDLE else ssl_,create_unverified...

tensorflow_datasets/core/download/downloader.py
cyfra: local variables should not be written with ALL_CAPS

tensorflow_datasets/core/download/downloader.py
```diff
@@ -172,7 +191,11 @@ def _sync_download(self, url, destination_path):
         opener = urllib.request.build_opener(proxy)
         urllib.request.install_opener(opener)   # pylint: disable=too-many-function-args
       request = urllib.request.Request(url)
-      response = urllib.request.urlopen(request)
+
+      if CA_VERIFY['urllib'] is None:     # disable ssl context check for python version <= 2.7.8
```
cyfra: do I read the code correctly, that what you're doing is just for the ftp:// flow ? 

If yes - please mention it in the CL description

tensorflow_datasets/core/download/downloader_test.py
```diff
@@ -144,6 +143,15 @@ def test_ftp_error(self):
     promise = self.downloader.download(url, self.tmp_dir)
     with self.assertRaises(downloader.urllib.error.URLError):
       promise.get()
+  
+  def test_ssl_mock(self):
+    ssl_mock_dict = downloader.ssl.__dict__.copy()
```
cyfra: This test doesn't seem to be verifying the feature that you've added.

(for example - you should somehow check that the call to fetch the request did contain the "context" parameter etc.

Did you also do manual verification ?

tensorflow_datasets/core/download/downloader.py
captain-pool: Not really. It gives the file path if CA_BUNDLE is not None else it returns False

tensorflow_datasets/core/download/downloader.py
captain-pool: Okay. I'm not really sure how to fetch fetch the default logger from tfds. Can you help me out? Like some function or something which returns an object to the logger. I think just `print()`-ing it out is not an option.

tensorflow_datasets/core/download/downloader.py
captain-pool: Yes. The disabled SSL skipping mechanism is only for ftps:// networks, for python 2.7.8.

tensorflow_datasets/core/download/downloader_test.py
captain-pool: I just checked if it is throwing any error for python 2.7.8 flow. I'm not really sure how to mimic SSLError. with either requests or urllib. Can you help me out with that?

tensorflow_datasets/core/download/downloader.py
captain-pool: Sure. :)

captain-pool: Not really. Proxy is a different option. It is already been merged to the master branch. User might or might not set proxy. And also, user might or might not set the path to CA Bundle. Both are different things. And they depend on different environment variables. :)

tarrade: right CA_BUNDLE is only needed for SSL interception proxy:

```
export TFDS_HTTPS_PROXY="http://user:password@ip:port/"
export TFDS_CA_BUNDLE=path/ca_certs
```

Otherwise:

```
export TFDS_HTTPS_PROXY="http://user:password@ip:port/"
```
 is enough

I will validate if a full path or a local path is needed for the CA Certificate

cyfra: Right - even more important to have all these options documented.

tensorflow_datasets/core/download/downloader_test.py
cyfra: For starters: the path that you're requesting in this test, needs to be ftp (otherwise your code is not executed).

Then, you have to mock/set side effects on the urllib part (see the test_ftp methods).

And afterwards you can verify that the urlopen was called with the context object, by using
assert_called_once_with

tensorflow_datasets/core/download/downloader.py
captain-pool: > I'd like variables to have concrete types assigned (rather than "either bool or string" :-)

Sure :)

tensorflow_datasets/core/download/downloader.py
captain-pool: > from absl import logging
> 
> ssl.**dict**['_create_unverified_context'] = lambda: logging.info("XXXXX")

That looks great! On it!

tensorflow_datasets/core/download/downloader.py
captain-pool: Cool :)

tensorflow_datasets/core/download/downloader.py
captain-pool: Actually for this parameter, if `verify == False`, it will not verify, if `verify is None`, it will verify using the default system certificates. And if verify is set to any CA Path, it will verify using that property. So, I think even if we go for only one datatype here, I need to perform another if-else check in the and pass the parameters, causing code duplication. :(
What do you suggest? What should I do?

tensorflow_datasets/core/download/downloader_test.py
captain-pool: That sounds great! Thanks!

captain-pool:
Hey @cyfra why is the kokoro build not starting?

captain-pool:
Review Requested.
CC: @cyfra @Conchylicultor @rsepassi

PR #670, opened 2019-06-13
Add `show_examples(ds)` option for image datasets.

ChanchalKumarMaji:
This is an initial commit. Please suggest changes to make it better. Thanks.


tensorflow_datasets/core/dataset_info.py
```diff
@@ -407,7 +408,53 @@ def initialize_from_bucket(self):
       out_fname = os.path.join(tmp_dir, os.path.basename(fname))
       gcs_utils.download_gcs_file(fname, out_fname)
     self.read_from_directory(tmp_dir)
-
+
```
Conchylicultor: Please add a test, even if you do not check that it is actually displaying something, it is important to check that the code is not raising error.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -407,7 +408,53 @@ def initialize_from_bucket(self):
...
+    for i in range(rows*columns):
+      fig.add_subplot(rows, columns, i+1)
+      img = X[i]
+      if img.shape[2] == 1:
+        img = img.reshape(img.shape[:2])
+      plt.imshow(img, cmap='gray')
+      if show_label:
+        plt.xlabel("label: int={}, str={}".format(y[i], self._features['label'].int2str(y[i])))
+    plt.show()
+
```
Conchylicultor: As this file start being bigger and bigger, and this function is probably gonna get more complicated overtime, maybe you could move it to another module, and call it from this:

```
def show_examples(self, *args, **kwargs):
  return show_example_lib.show_examples(*args, **kwargs)
```

tensorflow_datasets/core/dataset_info.py
```diff
@@ -407,7 +408,53 @@ def initialize_from_bucket(self):
       out_fname = os.path.join(tmp_dir, os.path.basename(fname))
       gcs_utils.download_gcs_file(fname, out_fname)
     self.read_from_directory(tmp_dir)
-
+  
+  def show_batch(self, ds, max_images=None, rows=None, factor=3):
+    """Shows some random image samples from the image-dataset.
+       Works only with batch data. If image shapes are different then creating
+       batch will raise error. Possibly resize and create the batch data or use
+       batch_size=1.
```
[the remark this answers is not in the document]
ChanchalKumarMaji: Yes, sure. But there is a problem for images of dataset like `coco`. The shapes of the images are not consistent and so I think batch cannot be formed, without resizing to the identical shapes. So, while iterating raises an error due to different shapes.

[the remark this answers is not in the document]
Conchylicultor: I'm not sure I understand the issue. If the batch are not present, then you should not try to create it.
I'm thinking about something like:

```
examples = []

ds = iter(dataset_utils.as_numpy(ds.take(num_examples)))
for ex in ds:
  if len(ex['image'].shape) == 4:  # Batch dimension is present
    all_exs = batch_to_list(ex)  # Return a list of individual examples
    examples.extend(all_exs)
  if len(ex['image'].shape) == 3:  # Batch dimension non present
    examples.append(ex)

# examples contains the list of individual examples to display, with potentially different shapes
```

Conchylicultor:
Any updates on this ?

ChanchalKumarMaji:
> Any updates on this ?

Yes, I will complete this within 2-3 days.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -407,7 +408,53 @@ def initialize_from_bucket(self):
       out_fname = os.path.join(tmp_dir, os.path.basename(fname))
       gcs_utils.download_gcs_file(fname, out_fname)
     self.read_from_directory(tmp_dir)
-
+  
+  def show_batch(self, ds, max_images=None, rows=None, factor=3):
```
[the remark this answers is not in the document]
ChanchalKumarMaji: Actually, I adjusted the values by doing some experiment. For eg, if we use figsize=(3, 4) for rows=3 and cols=4 would give a bad plot, whereas figsize=(4, 3) will give a good plot, but figsize=(factor*4, factor*3) would give the best plot with nice and measured spacing.

tensorflow_datasets/core/dataset_info.py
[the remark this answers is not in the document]
ChanchalKumarMaji: Can we exclude the case for `batch` data ? Is it necessary ?

Conchylicultor:
Friendly ping :). What is the current status ?

ChanchalKumarMaji:
I left some comments - 
1. Actually, I adjusted the values by doing some experiment. For eg, if we use `figsize=(3, 4)` for `rows=3 and cols=4` would give a bad plot, whereas `figsize=(4, 3)` will give a good plot, but `figsize=(factor*4, factor*3)` would give the best plot with nice and measured spacing. So, keeping `factor` would be a good idea here, as it would just scale the images as per the user needs. 

2. Can we exclude the case for `batch` data ? Is it necessary ?

tensorflow_datasets/core/dataset_info.py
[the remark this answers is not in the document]
Conchylicultor: Sounds good. Let's drop the `.batch` case and only assume that batch size hasn't been set.

Add this assumption in the docstring

tensorflow_datasets/core/dataset_info.py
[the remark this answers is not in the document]
Conchylicultor: Ok, but I would rename `factor` => `scale` or `plot_scale` which I believe is more explicit.

You could also replace
```
  ds = iter(tfds.as_numpy(ds.take(num_examples)))
  
  examples = []
  for ex in ds:
    examples.append(ex)
```
by:
```
examples = list(tfds.as_numpy(ds.take(num_examples)))
```

Conchylicultor:
Additionally, currently the keys are hardcoded (`ex['image']`), which make it crash with a lot of dataset with `KeyError: image not present in dict`.

I feel it would be more generic to auto-detect the images/label:

```
label_keys = [
    k for k, feature in info.features.items()
    if isinstance(feature, features_lib.ClassLabel)
]
image_keys = [
    k for k, feature in info.features.items()
    if isinstance(feature, features_lib.Image)
]
if len(label_keys) != 1 and len(image_keys) != 1:
  raise ValueError('Visualisation not yet supported for dataset {}'.format(info.name))

```

Conchylicultor:
Or alternative implementation:
```
is_valid = False
if info.supervised_keys:
  image_key, label_key = info.supervised_keys
  if (isinstance(info.features[image_key], features_lib.Image) and
       isinstance(info.features[label_key], features_lib.ClassLabel)):
    is_valid = True

if not is_valid:
  raise ValueError(
      'Visualisation not yet supported for dataset {}. Was not able to auto-infer '
      'image and label keys'.format(info.name))
```

ChanchalKumarMaji:
One last question: How should I write the tests. Should I take some existing `fake examples` of mnist from the testing folder ?

setup.py
```diff
@@ -42,13 +39,13 @@
     'absl-py',
     'attrs',
     'dill',  # TODO(tfds): move to TESTS_REQUIRE.
-    'cityhash',
     'future',
+    'matplotlib'
```
Conchylicultor: Please rebase to remove merge artifacts

tensorflow_datasets/core/dataset_info.py
```diff
@@ -408,19 +409,52 @@ def initialize_from_bucket(self):
...
+    label_keys = [
+        k for k, feature in self.features.items()
+        if isinstance(feature, features_lib.ClassLabel)
+    ]
+    
+    if len(image_keys) != 1:
+      raise ValueError(
+          'Visualisation not supported for dataset `{}`. Was not able to auto-infer '
+          'image and label keys.'.format(self.name))
+    else:
```
Conchylicultor: Remove this `else` as this code is only reached if the condition is failed due to the exception

tensorflow_datasets/core/dataset_info.py
```diff
@@ -408,19 +409,52 @@ def initialize_from_bucket(self):
...
-    features_dict = self.features
-    features_pprint = "%s({\n %s\n    }" % (
-        type(features_dict).__name__,
-        pprint.pformat({
-            k: features_dict[k] for k in sorted(list(features_dict.keys()))
-        }, indent=8, width=1)[1:-1])
-    citation_pprint = '"""\n%s\n    """' % "\n".join(
-        [u" " * 8 + line for line in self.citation.split(u"\n")])
+  def show_examples(self, ds, rows=3, cols=3, plot_scale=3):
+    """Display some random (rows*columns) images from the image-dataset
```
Conchylicultor: Could you add some documentation about which assumptions are made ? For which dataset will this works,...

Also update the kwargs descriptions

Conchylicultor:
There is a `testing.DummyMnist()` class that you can use. However I'm not sure how to test the visualisation part with matlib. I think it is fine if the test only check that it runs without crashing.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -408,19 +409,52 @@ def initialize_from_bucket(self):
...
+        fig.add_subplot(rows, cols, i+1)
+        image = ex[image_key]
+        if image.shape[2] == 1:
+          image = image.reshape(image.shape[:2])
+        plt.imshow(image, cmap='gray')
+        if len(label_keys) == 1:
+          label_key = label_keys[0]
+          label = ex[label_key]
+          plt.xlabel("label: int={}, str={}".format(label, self.features[label_key].int2str(label)))
+      plt.show()
```
Conchylicultor: Maybe also return plt, so that users can save the plot afterwards.

tensorflow_datasets/core/dataset_info.py
ChanchalKumarMaji: Actually it will work with all `image` datasets, but labels in some datasets will not be displayed.
What do u mean by `kwargs` ? Are u talking about `ds, row, cols, plot_scale` ?

Conchylicultor: This function is available for all datasets, so it is important to clarify that this will only works for images dataset which only have a single main image.

Yes, the current docstring do not match the `ds, row, cols, plot_scale` (where is plot_scale ?).

tensorflow_datasets/core/dataset_info.py
ChanchalKumarMaji: Is this necessary, actually when it is displayed we get an option to save the plot manually.

ChanchalKumarMaji:
When I am testing with - 
```python3
  def test_show_examples(self):
    with testing.tmp_dir(self.get_temp_dir()) as tmp_dir:
      builder = RandomShapedImageGenerator(data_dir=tmp_dir)
    ds = builder.as_dataset(split='train')
    builder.info.show_examples(ds)
```
the test halts temporarily due to the display of matplotlib window. How to deal with this ?

us:
@ChanchalKumarMaji You can close window with `plt.close('all')` or maybe you can write to a temp file instead.

us:
> When I am testing with -
> 
> ```python
>   def test_show_examples(self):
>     with testing.tmp_dir(self.get_temp_dir()) as tmp_dir:
>       builder = RandomShapedImageGenerator(data_dir=tmp_dir)
>     builder.download_and_prepare()
>     ds = builder.as_dataset(split='train')
>     builder.info.show_examples(ds)
> ```
> 
> the test halts temporarily due to the display of matplotlib window. How to deal with this ?
> 
> @us @Conchylicultor

Other thing about testing, you should show examples from test files because it's smaller and faster than train files.

tensorflow_datasets/core/dataset_info.py
ChanchalKumarMaji: It will works only for image datasets which only have a single main image. Otherwise it will raise that ValueError.

ChanchalKumarMaji:
> @ChanchalKumarMaji You can close window with `plt.close('all')` or maybe you can write to a temp file instead.

Actually I have a plot function in dataset_info.py and I am testing it in dataset_info_test.py. So, while testing I dont want to display the plot. plt.close('all') is not working. Any better ideas ?

tensorflow_datasets/core/dataset_info_test.py
```diff
@@ -301,6 +302,13 @@ def test_updates_on_bucket_info(self):
     # These are dynamically computed, so will be updated.
     self.assertEqual(40, info.splits.total_num_examples)
     self.assertEqual(2, len(info.as_proto.schema.feature))
+    
+  def test_show_examples(self):
+    with testing.tmp_dir(self.get_temp_dir()) as tmp_dir:
+      builder = RandomShapedImageGenerator(data_dir=tmp_dir)
+    builder.download_and_prepare()
+    ds = builder.as_dataset(split='train')
```
us: Maybe `split='test'`be faster

tensorflow_datasets/core/dataset_info.py
```diff
@@ -407,6 +409,51 @@ def initialize_from_bucket(self):
...
+    This function will only work for image datasets which only have 
+    a single main image.
+    
+    Args:
+      ds: `tf.data.Dataset`. Does not accept batch data.
+      rows: `int`, number of rows of the display grid.
+      cols: `int`, number of columns of the display grid.
+      plot_scale: `int`, controls the plot size of the images. Keep this
+        value around 3 to get a good plot. High and low values may cause
+        the labels to get overlapped.
```
us: Can you write an `Usage` example of this method?

tensorflow_datasets/core/dataset_info_test.py
```diff
@@ -301,6 +302,13 @@ def test_updates_on_bucket_info(self):
...
+    #builder.info.show_examples(ds)
```
us: You can use a mock library for your problem. Please check this [link](https://stackoverflow.com/questions/24343640/how-to-run-nosetests-without-showing-of-my-matplotlibs-graph).

tensorflow_datasets/core/dataset_info_test.py
ChanchalKumarMaji: Actually, its a dummy dataset with few random images. So, I think it will not vary much.

tensorflow_datasets/core/dataset_info.py
ChanchalKumarMaji: Yes, I will write in the documentation, once this is approved.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -42,12 +42,14 @@
 import tempfile
 
 from absl import logging
+import matplotlib.pyplot as plt
```
Conchylicultor: Please use matplotlib as lazy deps

tensorflow_datasets/core/dataset_info_test.py
ChanchalKumarMaji: Ok, thanks for pointing this. I am checking it.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -407,6 +409,51 @@ def initialize_from_bucket(self):
...
+    """Display some random (rows*columns) images from the image-dataset.
+    
+    This function will only work for image datasets which only have 
+    a single main image.
+    
+    Args:
+      ds: `tf.data.Dataset`. Does not accept batch data.
+      rows: `int`, number of rows of the display grid.
+      cols: `int`, number of columns of the display grid.
+      plot_scale: `int`, controls the plot size of the images. Keep this
```
Conchylicultor: Shouldn't scale be float ? `plot_scale=3.`

tensorflow_datasets/core/dataset_info.py
Conchylicultor: I would prefer. It does not cost anything and some people may prefer controlling the plot from their python code rather than UI

tensorflow_datasets/core/dataset_info_test.py
Conchylicultor: +1, mock seems the right tool to use

tensorflow_datasets/core/dataset_info.py
Conchylicultor: Adding in the documentation is good, but having an example snippet in the docstring is also important for people reading the code.

Conchylicultor:
Any updates for this ?

ChanchalKumarMaji:
I think if I return the plot, I cannot return its `xlabel`. Any better ideas ?

Conchylicultor:
> I think if I return the plot, I cannot return its `xlabel`. Any better ideas ?

I don't understand. What is the issue with:

```
def show_example(...):
  fig = plt.figure()
  plt.xlabel('xlabel')
  plt.ylabel('ylabel')
  return fig

fig = show_example()
fig.suptitle('test title')
fig.savefig('test.jpg')
```

Axis will be saved with the figure I believe. Also it's fine if the figure returned isn't exactly the same as the one displayed.

tensorflow_datasets/core/dataset_info.py
ChanchalKumarMaji: Where to add matplotlib as lazy dep ?

Conchylicultor: In: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/lazy_imports.py

More info in: https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md#extra-dependencies

ChanchalKumarMaji: Its already added - https://github.com/tensorflow/datasets/blob/2c43a797fec4fb10c246b229bc47c7b391346f8c/tensorflow_datasets/core/lazy_imports.py#L64

Conchylicultor: But you still have to us it in your function, instead of global import.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -447,19 +448,19 @@ def show_examples(self, ds, rows=3, cols=3, plot_scale=3):
     image_key = image_keys[0]
     num_examples = rows * cols
     examples = list(dataset_utils.as_numpy(ds.take(num_examples)))
-    fig = plt.figure(figsize=(plot_scale*cols, plot_scale*rows))
+    fig = tfds.core.lazy_imports.matplotlib.pyplot.figure(figsize=(plot_scale*cols, plot_scale*rows))
```
Conchylicultor: It is good to avoid reusing code whenever possible.
Here, it would be cleaner to have:
```
plt = tfds.core.lazy_imports.matplotlib.pyplot

plt.imshow(...)
plt.xlabel(...)
```

tensorflow_datasets/core/dataset_info.py
```diff
@@ -58,6 +59,8 @@
 from tensorflow_metadata.proto.v0 import schema_pb2
 from tensorflow_metadata.proto.v0 import statistics_pb2
 
+import tensorflow_datasets.public_api as tfds
```
Conchylicultor: This creates circular dependency (A import B, B import A). You should import individual files instead.

Conchylicultor:
Thanks. Almost good to merge. Have you tested this was still working after those changes?

Also please fix the kokoro test. This is failing due to circular dependency

tensorflow_datasets/core/dataset_info.py
ChanchalKumarMaji: I am doing
```python
from tensorflow_datasets.core import lazy_imports


plt = lazy_imports.matplotlib.pyplot 
```
but getting error - 
```
AttributeError: module 'tensorflow_datasets.core.lazy_imports' has no attribute 'matplotlib'
```

Conchylicultor: When you get an error like this, it is good to search for examples of usage of this feature to make sure that you're using it correctly. In this case, by searching [usage of lazy_imports](https://github.com/tensorflow/datasets/search?q=%22from+tensorflow_datasets.core+import+lazy_imports%22&unscoped_q=%22from+tensorflow_datasets.core+import+lazy_imports%22), you would have noticed that `tensorflow_datasets.core.lazy_imports` is the file, while you want to access the `lazy_imports` object present in this file.

```
from tensorflow_datasets.core import lazy_imports
plt = lazy_imports.lazy_imports.matplotlib.pyplot 
```

Conchylicultor:
The reason is that the server on which the tests run do not have monitors. You should try to mock `plt.figure` so that Tk don't get called.

PR #678, opened 2019-06-14
Add the missing voc2007 link in index

ilango100:
Added the missing voc2007 link in index of list of datasets. Checked for any other missing links, found out that wmt14, 15 , etc doesn't work with md file in github, but works in [html page](https://www.tensorflow.org/datasets/datasets).

I didn't find any other missing links so far.

### DIFF
--- a/docs/datasets.md
+++ b/docs/datasets.md
@@ -79,6 +79,7 @@ nightly package `tfds-nightly`.
     *   [`"svhn_cropped"`](#svhn_cropped)
     *   [`"tf_flowers"`](#tf_flowers)
     *   [`"uc_merced"`](#uc_merced)
+    *   [`"voc2007"`](#voc2007)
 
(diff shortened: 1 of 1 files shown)


Conchylicultor:
Note that this file is automatically generated, so the generation script should be fixed instead:
https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/document_datasets.py

Conchylicultor:
@ChanchalKumarMaji @us Do one of you want to have a look at this ?
It seems there is a bug in https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/document_datasets.py which skip some datasets from the table of content.

us:
@Conchylicultor Okay I can check it.

Conchylicultor:
There was an issue in our code. This should be fixed. Thanks for reporting.

PR #682, opened 2019-06-17
imdb.py fails on Windows

svobora:
TF2 Tutorial fails in _generate_examples() on Windows. Function re.compile has problems with "\\" character. There is a double interpretation of "\\", so it is necessary to replace it with "\\\\" such that after double interpretation, there is only single "\".


svobora:
I signed it!

Conchylicultor:
This was merged with https://github.com/tensorflow/datasets/commit/4d0e1447a8d1727c3cb6f9de491b3f648e885397

I'm not sure why our internal system didn't marked the PR as close.

svobora:
Because somebody copied the code and commited without actually merging my branch... Nevermind!

Conchylicultor:
This is a bug with our internal system. When we import the PR, it automatically creates a copy. However, it should also have automatically marked this PR as merged after we push the commit internally. I don't understand why it hasn't done it here.

PR #698, opened 2019-06-20
Add the missing links in index.

us:
I just run script and as far as i can see it all datasets added except `Librespeech`. 

`Librespeech` also not seems in the `list_builders()`.

@Conchylicultor


Conchylicultor:
Thank you for investigating this :)
I looked at our code, and it seems to be an internal issue which stripped the last dataset line. I've sent a fix so hopefully this should be resolved in the next automated doc generation.

Conchylicultor:
Closing this as the dataset list has been updated.
I looked at librispeech too and it seems that it was marked as `IN_DEVELOPMENT`, so it is skipped from the official list.

from `/audio/librispeech.py?g=0&l=152`:

```
IN_DEVELOPMENT = True
```
It seems that this dataset has never been completed. Not sure what was the issue

PR #713, opened 2019-06-27
mnist dataset falls back to its original website

facaiy:
I find sometimes the network connection becomes unstable when visiting storage.googleapis.com in China, and it results in download error. That's why I want to make it fall back to the original website.


.gitignore
```diff
@@ -17,4 +17,5 @@ dist/
 .pytest_cache/
 
 # Other
-*.DS_Store
\ No newline at end of file
+*.DS_Store
```
us: Please that .gitignore changes.

tensorflow_datasets/image/mnist.py
```diff
@@ -111,21 +112,27 @@ def _info(self):
             "label": tfds.features.ClassLabel(num_classes=10),
         }),
         supervised_keys=("image", "label"),
-        urls=[self.URL],
+        urls=[self.URL, _MNIST_URL_FALLBACK_],
```
us: It's a main page of mnist not a download link.

tensorflow_datasets/image/mnist.py
```diff
@@ -111,21 +112,27 @@ def _info(self):
...
     # Download the full MNIST Database
-    filenames = {
-        "train_data": _MNIST_TRAIN_DATA_FILENAME,
-        "train_labels": _MNIST_TRAIN_LABELS_FILENAME,
-        "test_data": _MNIST_TEST_DATA_FILENAME,
-        "test_labels": _MNIST_TEST_LABELS_FILENAME,
-    }
-    mnist_files = dl_manager.download_and_extract(
-        {k: urllib.parse.urljoin(self.URL, v) for k, v in filenames.items()})
+    def download_and_extract(url):
```
us: Why create a new function?

us:
Thank you for contributing.

.gitignore
facaiy: bypass pycharm(popular Python IDE) configurations.

tensorflow_datasets/image/mnist.py
facaiy: I think it makes try...except block easy to understand. So, let's remove the function or not?

facaiy:
@us thank you, can you take a look again?

tensorflow_datasets/image/mnist.py
us: I think inner function is confusing so you can delete. 

You can split `dl_manager.download` and `dl_manager.extract` functions. Now try..except block is be better.

.gitignore
facaiy: done, please to check #739

tensorflow_datasets/image/mnist.py
facaiy: done :-)

tensorflow_datasets/image/mnist.py
```diff
@@ -124,8 +125,15 @@ def _split_generators(self, dl_manager):
         "test_data": _MNIST_TEST_DATA_FILENAME,
         "test_labels": _MNIST_TEST_LABELS_FILENAME,
     }
-    mnist_files = dl_manager.download_and_extract(
-        {k: urllib.parse.urljoin(self.URL, v) for k, v in filenames.items()})
+
+    try:
```
us: the urls data have same folder structure so we can use try..exception block for just a `dl_manager.download()`, then use the `dl_manager.extract` method. 

And you need to be more clear when catch exception, Exception is too general for dl_manager. You should just catch the `tfds.core.download.downloader.DownloadError`.

facaiy: Thank you, fix it

PR #714, opened 2019-06-27
FACETS Visualization

us:
Example of Titanic dataset visualization.

facets

### DIFF
--- a/tensorflow_datasets/core/dataset_info.py
+++ b/tensorflow_datasets/core/dataset_info.py
@@ -47,7 +47,7 @@ import absl
 import numpy as np
 import six
 import tensorflow as tf
-import tensorflow_data_validation as tfdv
+import tensorflow_datasets.public_api as tfds
 
 from tensorflow_datasets.core import api_utils
 from tensorflow_datasets.core import naming
@@ -279,17 +279,15 @@ class DatasetInfo(object):
         "Cannot generate statistics for filetype {}".format(filetype_suffix))
     filepattern = naming.filepattern_for_dataset_split(
       builder.name, split, builder.data_dir, filetype_suffix)
-
     if filetype_suffix == "csv":
-      statistics = tfdv.generate_statistics_from_csv(
+      statistics = tfds.core.lazy_imports.tfdv.generate_statistics_from_csv(
         filepattern)
     else:
-      statistics = tfdv.generate_statistics_from_tfrecord(
+      statistics = tfds.core.lazy_imports.tfdv.generate_statistics_from_tfrecord(
         filepattern)
-
     logger = logging.getLogger()
     logger.disabled = True
-    schema = tfdv.visualize_statistics(statistics)
+    schema = tfds.core.lazy_imports.tfdv.visualize_statistics(statistics)
     logger.disabled = False
     return schema
 
@@ -297,7 +295,7 @@ class DatasetInfo(object):
     self._compute_dynamic_properties(self._builder)
     self._fully_initialized = True
 
-  def _compute_dynamic_properties(self, builder):
+  def _compute_dynamic_gproperties(self, builder):
     """Update from the DatasetBuilder."""
     # Fill other things by going over the dataset.
     splits = self.splits
--- a/tensorflow_datasets/core/dataset_info.py
+++ b/tensorflow_datasets/core/dataset_info.py
@@ -41,13 +41,16 @@ import os
 import posixpath
 import pprint
 import tempfile
+import logging
 
-from absl import logging
+import absl
 import numpy as np
 import six
 import tensorflow as tf
+import tensorflow_data_validation as tfdv
 
 from tensorflow_datasets.core import api_utils
+from tensorflow_datasets.core import naming
 from tensorflow_datasets.core import dataset_utils
 from tensorflow_datasets.core import splits as splits_lib
 from tensorflow_datasets.core import utils
@@ -265,6 +268,31 @@ class DatasetInfo(object):
   def _license_path(self, dataset_info_dir):
     return os.path.join(dataset_info_dir, LICENSE_FILENAME)
 
+  def visualize_dataset(self):
+    return self._visualize_dataset(self._builder, 'train')
+
+  def _visualize_dataset(self, builder, split):
+    """Visualize dataset statistics with FACETS DIVE."""
+    filetype_suffix = builder._file_format_adapter.filetype_suffix  # pylint: disable=protected-access
+    if filetype_suffix not in ["tfrecord", "csv"]:
+      raise ValueError(
+        "Cannot generate statistics for filetype {}".format(filetype_suffix))
+    filepattern = naming.filepattern_for_dataset_split(
+      builder.name, split, builder.data_dir, filetype_suffix)
+
+    if filetype_suffix == "csv":
+      statistics = tfdv.generate_statistics_from_csv(
+        filepattern)
+    else:
+      statistics = tfdv.generate_statistics_from_tfrecord(
+        filepattern)
+
+    logger = logging.getLogger()
+    logger.disabled = True
+    schema = tfdv.visualize_statistics(statistics)
+    logger.disabled = False
+    return schema
+
   def compute_dynamic_properties(self):
     self._compute_dynamic_properties(self._builder)
     self._fully_initialized = True
@@ -291,7 +319,7 @@ class DatasetInfo(object):
       except tf.errors.InvalidArgumentError:
         # This means there is no such split, even though it was specified in the
         # info, the least we can do is to log this.
-        logging.error(("%s's info() property specifies split %s, but it "
+        absl.logging.error(("%s's info() property specifies split %s, but it "
                        "doesn't seem to have been generated. Please ensure "
                        "that the data was downloaded for this split and re-run "
                        "download_and_prepare."), self.name, split_name)
@@ -372,7 +400,7 @@ class DatasetInfo(object):
       # If field is defined in code, we ignore the value
       if is_defined:
         if field_value != field_value_restored:
-          logging.info(
+          absl.logging.info(
               "Field info.%s from disk and from code do not match. Keeping "
               "the one from code.", field_name)
         continue
@@ -402,7 +430,7 @@ class DatasetInfo(object):
     data_files = gcs_utils.gcs_dataset_info_files(self.full_name)
     if not data_files:
       return
-    logging.info("Loading info from GCS for %s", self.full_name)
+    absl.logging.info("Loading info from GCS for %s", self.full_name)
     for fname in data_files:
       out_fname = os.path.join(tmp_dir, os.path.basename(fname))
       gcs_utils.download_gcs_file(fname, out_fname)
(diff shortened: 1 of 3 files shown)


tensorflow_datasets/core/lazy_imports.py
```diff
@@ -102,6 +102,11 @@ def skimage(cls):
     _try_import("skimage.external.tifffile")
     return _try_import("skimage")
 
+  @utils.classproperty
+  @classmethod
+  def tfdv(cls):
+    return _try_import("tensorflow_data_validation")
```
pierrot0: You would need to edit setup.py, no?

tensorflow_datasets/core/dataset_info.py
```diff
@@ -265,6 +268,29 @@ def _dataset_info_path(self, dataset_info_dir):
   def _license_path(self, dataset_info_dir):
     return os.path.join(dataset_info_dir, LICENSE_FILENAME)
 
+  def visualize_dataset(self):
+    return self._visualize_dataset(self._builder, 'train')
+
+  def _visualize_dataset(self, builder, split):
```
pierrot0: Please merge with above function.
Or is there a reason to have both visualize_dataset and _visualize_dataset functions?

tensorflow_datasets/core/dataset_info.py
```diff
@@ -265,6 +268,29 @@ def _dataset_info_path(self, dataset_info_dir):
...
+    """Visualize dataset statistics with FACETS DIVE."""
+    filetype_suffix = builder._file_format_adapter.filetype_suffix  # pylint: disable=protected-access
+    if filetype_suffix not in ["tfrecord", "csv"]:
```
pierrot0: We dropped support for csv. You can remove.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -265,6 +268,29 @@ def _dataset_info_path(self, dataset_info_dir):
...
+    filepattern = naming.filepattern_for_dataset_split(
+      builder.name, split, builder.data_dir, filetype_suffix)
+    if filetype_suffix == "csv":
+      statistics = tfds.core.lazy_imports.tfdv.generate_statistics_from_csv(
+        filepattern)
+    else:
+      statistics = tfds.core.lazy_imports.tfdv.generate_statistics_from_tfrecord(
+        filepattern)
+    logger = logging.getLogger()
+    logger.disabled = True
```
pierrot0: You should add a comment on why this is necessary.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -265,6 +268,29 @@ def _dataset_info_path(self, dataset_info_dir):
   def _license_path(self, dataset_info_dir):
     return os.path.join(dataset_info_dir, LICENSE_FILENAME)
 
+  def visualize_dataset(self):
+    return self._visualize_dataset(self._builder, 'train')
+
+  def _visualize_dataset(self, builder, split):
+    """Visualize dataset statistics with FACETS DIVE."""
```
pierrot0: You should describe what is being returned (type, ...) and how to use it.
Maybe with an example?

tensorflow_datasets/core/dataset_info.py
us: Because tfdv has a lot of warning so I needed to close.

tensorflow_datasets/core/dataset_info.py
us: Oky!

tensorflow_datasets/core/lazy_imports.py
us: Etienne suggest the add tfdv as a lazy import due to its size.
`• As tfdv is a big dependency, and create some issues on some system (ex: Colab). Make sure that tfdv is used as lazy dependency. As statistics are not essential to the program, you can skip the statistics computation if tfdv isn't installed.
`

tensorflow_datasets/core/dataset_info.py
us: No, just a little mistake. I think I was confused when I writing this.

us:
Yes it's enough to display.

us:
What's the problem on the `kokoro`? I checked it but didn't understand.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -41,13 +41,16 @@
...
 import pprint
 import tempfile
+import logging
 
-from absl import logging
+import absl
 import numpy as np
 import six
 import tensorflow as tf
+import tensorflow_datasets.public_api as tfds
```
Conchylicultor: You should import individual modules, instead of the full API, otherwise it creates circular dependency: `tfds => dataset_info => tfds`.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -265,6 +268,41 @@ def _dataset_info_path(self, dataset_info_dir):
   def _license_path(self, dataset_info_dir):
     return os.path.join(dataset_info_dir, LICENSE_FILENAME)
 
+  def visualize_dataset(self):
+    """Visualize dataset statistics with FACETS DIVE.
+
+       Example Usage:
+
+          titanic_builder = tfds.builder("titanic")
+          titatanic_builder.download_and_prepare()
```
Conchylicultor: typo

tensorflow_datasets/core/dataset_info.py
```diff
@@ -265,6 +268,41 @@ def _dataset_info_path(self, dataset_info_dir):
...
+    if filetype_suffix != "tfrecord":
+      raise ValueError(
+        "Cannot visualize dataset statistics because filetype {}".format(
+          filetype_suffix))
+
+    # Find the train .tfrecord files pattern of dataset.
+    filepattern = naming.filepattern_for_dataset_split(
+      self._builder.name, 'train', self._builder.data_dir, filetype_suffix)
+
+    statistics = tfds.core.lazy_imports.tfdv.generate_statistics_from_tfrecord(
```
Conchylicultor: I think there is confusion, in our design, we want to generate the TFDV statistics during dataset creation time, similarly to: https://github.com/tensorflow/datasets/pull/383

Then, the visualisation can be done simply `tfdv.visualize_statistics(info.statistics)`

The main advantage is that as some datasets are huge, we don't want to re-compute statistics every times (ex: iterating over imagenet may takes multiple hours).
Because we are pre-computing and updating the statistics info files on GCS, user can directly download the statistics without having to generate the dataset first. Like:
```
builder = tfds.builder('imagenet_2012')  # Download pre-computed statistics from GCS
builder.info.visualize_dataset()
```

Is your plan to restore  https://github.com/tensorflow/datasets/pull/383 in a new PR ?

tensorflow_datasets/core/dataset_info.py
Conchylicultor: Sounds good, but in this case, I would add this as a kwargs in case the user still want to access the logs, something like:

`info.visualize_dataset(disable_logging=True)`

tensorflow_datasets/core/dataset_info.py
```diff
@@ -41,13 +41,16 @@
 import posixpath
 import pprint
 import tempfile
+import logging
```
Conchylicultor: I would use `import logging as py_logging`, and keep the `from absl import logging`. Otherwise, there is a risk that people uses the wrong `logging.info(`

tensorflow_datasets/core/lazy_imports.py
Conchylicultor: Even if it is added as lazy deps, external users shouldn't encounter those issues which are Colab and Google internal specific.
So it should be safe to add it in setup.py

Conchylicultor:
> What's the problem on the `kokoro`? I checked it but didn't understand.

The issue seems to be because you import the full tfds API, thus it creates circular dependency: tfds => dataset_info => tfds.

tensorflow_datasets/core/dataset_info.py
us: Sorry about that. `tfdv.visualize_statistics` hasn't got warnings, these warnings are given by `generate_statistics_from_tfrecord`. And I think this warnings are unnecessary. Check this [log](https://gist.github.com/us/e6ec3ece40a250b5231fdda096ed49dc).

tensorflow_datasets/core/dataset_info.py
us: Yep I see the problem.
Do we pull infos automatically or do users have to download? If it's auto where is it?

us: Also if we keep the statistics on gcs, why compute the statistics again on users computer?

us: > Is your plan to restore #383 in a new PR ?
If I restore it, will we use it ? Because of tfdv problems.

us:
@Conchylicultor Still giving an error for `tf1.3`

Conchylicultor:
The issue seem to be a segfault when running TF 1.13. However I do not understand why how this is related to this cl as this feature isn't tested.
Can you reproduce the issue on your local machine, by running the test script https://github.com/tensorflow/datasets/blob/master/oss_scripts/oss_tests.sh ?

us:
I remove tfdv from setup.py and Kokoro passed now. 
Won't we add the tfdv to setup.py?
What we have to do?

us:
@Conchylicultor @cyfra please check last fixes

Conchylicultor:
This looks good thank you. Before merging it, it seems better to restore the TFDV statistics computation (https://github.com/tensorflow/datasets/pull/383), so that statistics are not recomputed before each visualisation. It would also works with S3.
But I probably won't have time to work on this before the end of the week.

tensorflow_datasets/core/dataset_info.py
```diff
@@ -265,6 +268,42 @@ def _dataset_info_path(self, dataset_info_dir):
...
+      raise ValueError(
+        "Cannot visualize dataset statistics because filetype {}".format(
+          filetype_suffix))
+
+    # Find the train .tfrecord files pattern of dataset.
+    filepattern = naming.filepattern_for_dataset_split(
+      self._builder.name, 'train', self._builder.data_dir, filetype_suffix)
+
+    statistics = LazyImporter.tfdv.\
+                              generate_statistics_from_tfrecord(filepattern)
```
us: @Conchylicultor How can I get the generated statistics ?

Conchylicultor: You can try to patch #782, then you should be able to use `info.splits['train'].statistics` directly.
 have issue with merging #782 as it doesn't works well with pytest-xdist and it significantly slow down the tests. Not sure yet what to do.

us: Should I wait or merging this patch?

Conchylicultor: You can test it on another branch if you want to test it, but I would probably wait for this PR.

us: Okay I will wait PR.

cyfra: This is cool - can you also mention it in the main documentation ?

Is this PR ready to merge ?

Conchylicultor: @cyfra There are unresolved  internal issues (see changelist 257065400).
I think it is best to wait for tfdv integration before merging this as it will simplify the statistics computations (the current code is not compatible with S3 I believe).
For tfdv, I'm waiting for the migration to S3 to complete and legacy split API has been removed.

Conchylicultor:
Closing this as this was merged with `tfds.show_statistics`

PR #191, opened 2019-03-08
Initial commit, added Coco2017 dataset with Panoptic annotations

kaustumbh7:
Fixes https://github.com/tensorflow/datasets/issues/25

Signed-off-by: Kaustumbh <kaustumbhjaiswal11@gmail.com>

Issue #25, opened 2019-01-22
[data request] COCO 2018


rodrigob:
* Name of dataset: COCO 2018
* URL of dataset: http://cocodataset.org/#panoptic-2018
* License of dataset: CC 4.0 see http://cocodataset.org/#termsofuse
* Short description of dataset and use case(s): Thing and stuff annotations over Coco images

Folks who would also like to see this dataset in `tensorflow/datasets`, please +1/thumbs-up so the developers can know which requests to prioritize.

kaustumbh7:
Can anyone please guide me what all has to be done here? I am interested in working on this.

Conchylicultor:
That would be great. Thank you!
It shouldn't be too much work.

You can look at the current Coco2014 to create a similar Coco2018 by changing the downloaded data used. Try to reuse code as possible (by using inheritance).
https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/coco.py

Additionally, while doing this, it would be great if you also where adding a `"image/id": tfds.features.Text(),` field to the `FeaturesDict` corresponding to the Coco image_id associated with each image. This was another feature requested by some users. This is required to use the official Coco evaluation tool.

kaustumbh7:
Okay. No problem! If I have any doubts can I ask you @Conchylicultor ?

Conchylicultor:
Yes of course, don't hesitate if you encounter any issues

kaustumbh7:
@rodrigob Can you please share the link from where the dataset can be downloaded ? I am unable to find it because it is not available in the COCO api.

Conchylicultor:
I simply found the links from the official website: http://cocodataset.org/#download

I think 2018 uses the same data from 2017 (see http://cocodataset.org/#panoptic-2018):
`Participants are recommended but not restricted to train their algorithms on COCO 2017 train and val sets. The download page has links to all COCO 2017 data.`

The dataset should be named `Coco2017` then

Conchylicultor:
I agree their website is confusing. from what I understand, "2017 Train/Val/Test"should contains 2018 annotations:
`2017 Train/Val/Test:  Detection 2017, Keypoints 2017, Stuff 2017, Detection 2018, Keypoints 2018, Stuff 2018, Panoptic 2018`

`Note: the split year refers to the year the image splits were released, not the year in which the annotations were released.`

kaustumbh7:
Yes, even I thought so. Thanks for the help! 😊

[...]

PR #716, opened 2019-06-28
coco2017 and coco2017_panoptic

ChanchalKumarMaji:
Issue #25 . Closes #191 .


tensorflow_datasets/image/coco.py
```diff
@@ -262,6 +264,413 @@ def build_bbox(x, y, width, height):
...
+    categories_id2name = {c["id"]: c["name"] for c in categories}
+    
+    # Extract the panoptic_{}.zip folder containing the png images
+    panoptic_image_filename = "panoptic_{}.zip"
+    panoptic_image_path = os.path.join(
+        annotation_dir,
+        "annotations",
+        panoptic_image_filename.format(split_type)
+    )
+    with ZipFile(panoptic_image_path) as f:
```
us: Why didn't use `dl_manager.extract` method?

ChanchalKumarMaji: I got something similar [here](https://github.com/tensorflow/datasets/blob/813b5d2aa9c8c1a4c8d0aae7130239babe276983/tensorflow_datasets/image/imagenet.py#L138).

us: @Conchylicultor @pierrot0 what's the difference of `zipfile` package and `dl_manager`?

pierrot0: The problem here is that ZipFile will not use tf.gfile, which it should.
Using dl_manager will do the right thing (open the file with tf.gfile), and use the right format if not zip.

So I would prefer you to use dl_manager Chanchal.
Otherwise Zipfile is fine if code is not a repeat of the extractor and you know it's alway a zip.

ChanchalKumarMaji: Actually, the file is always a `zip` file. So, should I change ?

Conchylicultor: Yes, as Pierre say, `ZipFile` won't works for non-standard file systems (ex: cannot read from GCS & cie), so it is better to use `dl_manager.extract()`.

Additionally, `dl_manager` will cache and reuse the result, while using `ZipFile` will re-extract the data or may crash due to data already has been extracted once.

Conchylicultor:
Looks good, but it seems that the three Coco2014, Coco2017 and Panoptic duplicates a lot of code.
Would it be possible instead to refactor the code to avoid duplication ? For instance by using a base class from which Coco2014 and Coco2017 would inherit ?

ChanchalKumarMaji:
> Looks good, but it seems that the three Coco2014, Coco2017 and Panoptic duplicates a lot of code.
> Would it be possible instead to refactor the code to avoid duplication ? For instance by using a base class from which Coco2014 and Coco2017 would inherit ?

Actually I tried to reduce the codes by using `inheritance` or something, but its not possible here. Because every part contains some changes so its difficult to do that.

For coco2014 and coco2017 `builder_configs` could not be used because they slightly differ in the structure. I thought not to change coco2014 as it has been created perfectly.

Conchylicultor:
Thank you @ChanchalKumarMaji, I merged it.

The main advantage of sharing code is that when updating it (ex: adding more metadata, like captioning), all versions benefit from it without having to update three class individually.

I agree here it is trickier due to the fact that each version is slightly different. I've done the code sharing before merging it. You can have a look at https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/coco.py to see how this was done.

PR #739, opened 2019-07-05
add .idea in the gitignore

facaiy:
@us suggest to open a new PR in #713


nok:
In addition we should ignore `*.iml` files which come with IntelliJ and PyCharm.

Issue #728, opened 2019-07-03
Seepdup kokoro tests


pierrot0:
pytest-xdist might be useful here, to run tests in parallel.
S3 (auto sharding) should be useful to reduce the number of images big datasets need for tests.
Fake data might provide pictures which are more easily compressible.

pierrot0:
https://github.com/tensorflow/datasets/blob/ac89b873ccfe3c13bca7e838e1275cfdf170d127/oss_scripts/oss_tests.sh#L41

ChanchalKumarMaji:
Terminal Log when `no error` occurs in tests - https://gist.github.com/ChanchalKumarMaji/844fd51fcbb050a7cd16196a658936df
Terminal Log when `some error` occurs in tests - https://gist.github.com/ChanchalKumarMaji/c5bf33fe5afe903ab85bc865ea6a91fe

Just check if this looks good or not, then I will create a PR. I think some trivial changes are required in this issue.

Conchylicultor:
Close issue as tests are using pytest-xdist now

PR #500, opened 2019-04-24
Parallelize tests with pytest-xdist

tfds-copybara:
Parallelize tests with pytest-xdist

PR #743, opened 2019-07-07
use `pytest-xdist` to speed up kokoro.

ChanchalKumarMaji:
Issue #728.


us:
related pr #500

Conchylicultor:
I've tested locally and it seems to works as expected. Merging it soon. Many thanks.

PR #784, opened 2019-07-17
Adding examples to DatasetBuilder.as_dataset()

habernal:
As the output type of DatasetBuilder().as_dataset() depends on the requested parameters (splits, supervised, etc.) which might be confusing for a beginner, here's a few detailed examples on a standard dataset that should help to understand the behavior.

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -319,6 +319,46 @@ class DatasetBuilder(object):
     """Constructs a `tf.data.Dataset`.
 
     Callers must pass arguments as keyword arguments.
+    
+    The output types vary depending on the parameters. Examples:
+    
+    ```python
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/core/dataset_builder.py
```diff
@@ -319,6 +319,46 @@ def as_dataset(self,
...
     Callers must pass arguments as keyword arguments.
+    
+    The output types vary depending on the parameters. Examples:
+    
+    ```python
+    ds_builder = tfds.text.imdb.IMDBReviews()
+    ds_builder.download_and_prepare()
+    
+    # Default parameters
+    ds1 = ds_builder.as_dataset()
```
us: Yes it's a small script example but the names of `ds1` and `ds2` may be more clearer and understandable.

us:
@habernal Thanks for contributing!

Issue #765, opened 2019-07-12
ucf101 is failing for Python 3 and TF 1.13


Conchylicultor:
Since yesterday, Kokoro continous tests show this dataset as failing for Python 3 and TF 1.13.
The error message isn't particularly explicit so not clear what is the issue.

```
=================================== FAILURES ===================================
___________________ tensorflow_datasets/video/ucf101_test.py ___________________
[gw3] linux -- Python 3.6.2 /home/kbuilder/virtualenv/tfds_env/bin/python3.6
worker 'gw3' crashed while running 'tensorflow_datasets/video/ucf101_test.py::Ucf101Test::test_download_and_prepare_as_dataset'
===== 1 failed, 883 passed, 203 skipped, 9159 warnings in 1016.35 seconds ======
set_status
+ set_status
+ local last_status=1
+ [[ 1 -ne 0 ]]
+ echo '<<<<<<FAILED>>>>>> Exit code: 1'
<<<<<<FAILED>>>>>> Exit code: 1
+ STATUS=1
# If not running with TF2, ensure Eager is not enabled by default
if [[ "$TF_VERSION" != "tf2" ]]
then
  pytest \
    --disable-warnings \
    tensorflow_datasets/eager_not_enabled_by_default_test.py
  set_status
fi
+ [[ 1.13.1 != \t\f\2 ]]
+ pytest --disable-warnings tensorflow_datasets/eager_not_enabled_by_default_test.py
```

nok:
JFIY, yesterday I cloned the repository (commit id 501ce7c) and ran your mentioned test successfully.

My environment:
- macOS 10.14.5
- Python 3.5.6
- TF CPU 1.13.1 (1.13.0 isn't provided as pre built for Python 3.5)

Nevertheless in other tests I ran in `tensorflow.python.framework.errors_impl.ResourceExhaustedError` exceptions, but that isn't subject of this issue.

Conchylicultor:
Yes, it may be a bug from Kokoro and pytest when running in parallel.

Conchylicultor:
Seems obsolete

PR #786, opened 2019-07-17
Fix checksum input

us:
Related issue #769.

### DIFF
--- a/tensorflow_datasets/core/download/resource.py
+++ b/tensorflow_datasets/core/download/resource.py
@@ -184,7 +184,7 @@ def get_dl_fname(url, checksum):
   Returns:
     string of 90 chars max.
   """
-  checksum = base64.urlsafe_b64encode(_decode_hex(checksum))
+  checksum = base64.urlsafe_b64encode(_decode_hex(checksum.strip()))
   checksum = tf.compat.as_text(checksum)[:-1]
   name, extension = _sanitize_url(url, max_length=46)
   return '%s%s%s' % (name, checksum, extension)


Conchylicultor:
Thank you for fixing this. Can you provide more context to help us avoid this in the future ? Why was a space added in the checksum ? Why was it working before ?

ChanchalKumarMaji:
Some workers are crashing which is causing `kokoro` to fail. What to do ? Here `wikipedia_test.py` is failing. 
This is also an issue on their side as well as I saw many discussions are going on related to this.

us:
The only thing that changes is the version of python. So I thought there was a change that came with the update.

@ChanchalKumarMaji I agree with you. Crash cause is kokoro, testing is okay on my computer.

Conchylicultor:
@us @ChanchalKumarMaji Yes, the crashes are due to pytest-xdist, similar with #765 which make some tests flaky. I don't have a good solution to fix this currently, except re-running Kokoro

20zinnm:
Can we merge this PR? This issue is very annoying since I develop on a Windows desktop. @us

us:
@20zinnm This pr not merged because tests are not finished. Until this code merges, you can edit this line and compile from the source code.

Conchylicultor:
@20zinnm from what I understand, the issue was the 260 characters limitation on windows. So this fix attempt would not have solved your issue.
So if your issue is solved, can we close this PR ?

20zinnm:
No, that is a separate issue. This PR resolves the issue of "odd-length
string" while the other resolves the file not found error when creating the
folder.

20zinnm:
So can this be merged @Conchylicultor ? This PR resolves an issue on windows that prevents any datasets from being downloaded.

@us The kokoro build finished, unless you mean there need to be new tests written for this case?

Conchylicultor:
@20zinnm I would first like to understand why does this bug only appear on windows and where does the extra character comes from.
If the checksum indeed contains an extra space, then the issue is probably located somewhere else in the code. But it is difficult to debug as we do not have windows computer to debug the issue.

20zinnm:
@Conchylicultor The checksum does not contain an extra space, but rather an extra `\r` at the end of the line.

This is because Python uses `\n` as the separator for reading in by

I just dropped a breakpoint at [`get_dl_fname`](https://github.com/tensorflow/datasets/blob/efcd4ca21e9f1fd7be66e01bbca4ffa1a3b4d873/tensorflow_datasets/core/download/resource.py#L187) and found the value for MNIST to be `440fcabf73cc546fa21475e81ea370265605f56be210a4024d2ca8f203523609\r`.

The `\r` is because when Git checks out code to Windows, by default it replaces `LF` with `CRLF`, and vice versa when pushing. This is why `pip install tensorflow_datasets` does not encounter this error, but `git clone https://github.com/tensorflow/datasets.git && cd datasets && pip install -e .` does, and it's Windows-specific.

A simple solution is to add a `.gitattributes` file to the root of the repository which tells Git to not replace `LF` with `CRLF` for any checksum file.

20zinnm:
I can submit a PR to resolve this without changing the code. Give me a few minutes.

Conchylicultor:
@20zinnm  Thank you for this detailed explanation. This will be fixed today.

Conchylicultor:
Should be fixed with 76325da.
Thanks @us for the fix. This was the right fix, but I just moved it when file is read to avoid this bug in other places.

PR #790, opened 2019-07-18
Documenting DatasetInfo parameter in detail

habernal:
As `supervised_keys` influences the way the dataset might be converted using `as_dataset()`, it is worth documenting this parameter in more detail.


Conchylicultor:
Thanks for improving our doc. Will merge this soon.

PR #794, opened 2019-07-19
Add TensorFlow Datasets and TF 1.14+ installation to add_dataset.md

8bitmp3:
Suggestion to the [Add a dataset](https://www.tensorflow.org/datasets/add_dataset) guide: 

- Add `"first install TensorFlow (1.14+) and TFDS by running pip install..."` before running the `create_new_dataset.py` script. Otherwise, it will return an error: `ModuleNotFoundError: No module named 'tensorflow_datasets'`. 

The installation procedure is mentioned in the [Overview](https://www.tensorflow.org/datasets/overview) but it's not explicitly stated in [Add a dataset](https://www.tensorflow.org/datasets/add_dataset) and a user may skip the Overview before encountering this issue. 

Let me know if you have any questions.

### DIFF
--- a/docs/add_dataset.md
+++ b/docs/add_dataset.md
@@ -67,9 +67,9 @@ generate on a single machine. See the
 
 If you want to
 [contribute to our repo](https://github.com/tensorflow/datasets/blob/master/CONTRIBUTING.md)
-and add a new dataset, the following script will help you get started by
-generating the required python files,...
-To use it, clone the `tfds` repository and run the following command:
+and add a new dataset, first install TensorFlow (1.14+) and TFDS by running `pip install tensorflow tensorflow-datasets`. 
+The following script will help you get started by generating the required Python files.
+To use it, fork the `tfds` repository, clone it and run the following command:
 
 ```
 python tensorflow_datasets/scripts/create_new_dataset.py \


docs/add_dataset.md
```diff
@@ -67,9 +67,9 @@ generate on a single machine. See the
 
 If you want to
 [contribute to our repo](https://github.com/tensorflow/datasets/blob/master/CONTRIBUTING.md)
-and add a new dataset, the following script will help you get started by
-generating the required python files,...
-To use it, clone the `tfds` repository and run the following command:
+and add a new dataset, first install TensorFlow (1.14+) and TFDS by running `pip install tensorflow tensorflow-datasets`.
```
Conchylicultor: If the tensorflow-datasets github repo should be cloned, the pip install is useless no ?

8bitmp3: Maybe we can add: run the `setup.py` with pip, then carry on, What do you think? @Conchylicultor

Conchylicultor: Why ? If the user has cloned the repository, it can work on the cloned version directly. No installation needed.

If he want to just use the tfds API but without contributing, then indeed he can use the pip version and implement the dataset independently.

8bitmp3: I understand @Conchylicultor . It is just that the issue arose when actually attempting to add a dataset while following the https://www.tensorflow.org/datasets/add_dataset guide without looking at the TFDS overview first. This is the reason why I suggested this little change in the PR. If you believe it is something that is not necessary we shall close the PR, no problem.

cyfra: Thanks for your PR @8bitmp3.
But @Conchylicultor  is right - you should not install the package from pip this way (as it will be installing older version etc).

The proper flow should be: clone the code, and then use "pip -e ." within the directory.
This both makes sure that dependencies (like tensorflow) are installed. And also installs the package in "edit / developer" mode (basically creating softlinks to your current directory)

cyfra: Hey @8bitmp3 - do you plan to keep working on this PR or should I close it?

PR #795, opened 2019-07-20
 Add dataset-name to download_dir.

us:
Downloaded files have complicated names to understand so with this offer the downloaded files will be have better names that is more understandable name.

*Old version*:
`tensorflow_datasets/downloads/openml.org_get_csv_16826755_phpMYEkMlxhfbLHRwcWJQ9vABvlEwTHa8yIFVJ6uLrnNL3KBzVzc`

*Suggested version*:
`tensorflow_datasets/downloads/titanic/openml.org_get_csv_16826755_phpMYEkMlxhfbLHRwcWJQ9vABvlEwTHa8yIFVJ6uLrnNL3KBzVzc`

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -645,7 +645,7 @@ class DatasetBuilder(object):
 
   def _make_download_manager(self, download_dir, download_config):
     download_dir = download_dir or os.path.join(self._data_dir_root,
-                                                "downloads")
+                                                "downloads", self.name)
     extract_dir = (download_config.extract_dir or
                    os.path.join(download_dir, "extracted"))
     manual_dir = (download_config.manual_dir or
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -647,7 +647,7 @@ class DatasetBuilder(object):
     download_dir = download_dir or os.path.join(self._data_dir_root,
                                                 "downloads", self.name)
     extract_dir = (download_config.extract_dir or
-                   os.path.join(download_dir, "extracted"))
+                   os.path.join(self._data_dir_root, "extracted", self.name))
     manual_dir = (download_config.manual_dir or
                   os.path.join(download_dir, "manual"))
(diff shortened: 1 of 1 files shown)


Conchylicultor:
Multiple datasets can reuse the same data (ex: `imagenet2012` and `imagenet2012_corrupted`) so we cannot hardcode the dataset name in the path to avoid downloading data twice

us:
Yes tfds understands it but has no readability for humans. Any developer cannot review all `.INFO` files to find the dataset if there are too many downloaded datasets. Maybe we can write a checker for into the download folders or we can save downloaded files in one place.

Conchylicultor:
Yes, we could add a small script which lists all downloads directories for a specific dataset.

us:
Like that ?

```python 
def download_folder_finder(dataset_name):
	home = os.path.expanduser('~')
	path = os.path.join(home, 'tensorflow_datasets', 'downloads')

	for r, d, f in os.walk(path):
		for file in f:
			if ".INFO" in file:
				aha = os.path.join(r, file)
				filename = os.path.splitext(aha)[0]
				with open(aha) as data_file:
					data_item = json.load(data_file)
					if data_item['dataset_names'][0] == dataset_name:
						return filename
	return None
```

Conchylicultor:
Yes, with some small modification:
* Replace `if data_item['dataset_names'][0] == dataset_name:` because multiple dataset may download the same file, so you need to iterate over all of it.
* Use gfile instead of `os.walk` and `open`
* `file.endswith('.INFO')`
* The download_dir may not be the default Home directory, so the user should be able to customizing it

I also had another question about the auto-testing program. Do you detect datasets using `dl_manager.iter_archive` VS `dl_manager.download_and_extract` ? One way would be to look at the extract folder to see if archives have been extracted.

us:
Thanks for modifications ! 

Oh this was a question that had been on my mind for a week. This is a nice trick the detecting iter vs extract problem. I will use it.

Conchylicultor:
Closing this to cleanup our PR list.
We should still add some script to list downloads from a specific dataset, by opening a new PR.

PR #796, opened 2019-07-20
Optimizing Imports with Pycharm

us:
Remove unused import statements for all project.
Add missing import statements.
Arrange order of names.


Conchylicultor:
Thank you, but unfortunately, those changes are not compatible with pylint. (Try to run https://github.com/tensorflow/datasets/blob/master/oss_scripts/lint.sh)

It seems that pycharm and google are using different conventions. This choice is arbitrary but we should be consistent with the rest of Google projects.

us:
Okay! Sorry I missed it pylint. 
I checked all files again and I send a new commit for unused imports.

PR #279, opened 2019-03-18
Added `tensorflow_datasets.__version__`

ChanchalKumarMaji:
Solves #278 . @rsepassi @cyfra @Conchylicultor  please check.

output:
```
>>> import tensorflow_datasets
>>> tensorflow_datasets.__version__
'1.0.1-dev'
```

PR #798, opened 2019-07-22
Include __version__ in __all__

craymichael:
Self-explanatory - I believe `__version__` should be included in `__all__` so TFDS version can be checked as

    tfds.__version__

instead of

    from tensorflow_datasets import version as tfds_version
    tfds_version.__version__


ChanchalKumarMaji:
My PR #279 was like this, but it was modified by the authors while merging.

Conchylicultor:
Thanks. Merging soon

PR #804, opened 2019-07-23
Change `tf.io.gfile` instead of os functions

us:


Conchylicultor:
Thanks merging it now. Note that the scripts in testing/ to generate the fake data will always be generated on local machine, so here gfile isn't required

Issue #758, opened 2019-07-11
[data request] binary-mnist


pierrot0:
* Name of dataset: binary-mnist
* URL of dataset: https://github.com/yburda/iwae/blob/master/datasets/BinaryMNIST/README
* License of dataset: MIT
* Short description of dataset and use case(s): Starting with the MNIST dataset, a train-test-valid split was introduced, values were scaled to [0, 1], and pixels were then sampled as Bernoulli random variables from those values to produce this official dataset. The dataset was first used in Salakhutdinov and Murray (2008) and since used in papers such as NADE, DARN, and Variational Autoencoders.

Label information is not included, so the relevant tasks are unsupervised learning and generation.

us:
@Conchylicultor @cyfra I think it can be closed.

PR #809, opened 2019-07-24
Add binarized_mnist to TFDS.

tfds-copybara:
Add binarized_mnist to TFDS.


us:
Related issue #758.

tensorflow_datasets/image/binarized_mnist.py
```diff
@@ -0,0 +1,112 @@
...
+  VERSION = tfds.core.Version("1.0.0")
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            "image": tfds.features.Image(
+                shape=mnist.MNIST_IMAGE_SHAPE)}),
+        urls=[self.URL],
```
us: This url should be dataset page link, not download link.

us:
@gjtucker why didn't inherit anything from `Mnist` class?

tensorflow_datasets/core/download/checksums.py
```diff
@@ -76,7 +76,7 @@ def _get_sizes_checksums(checksums_path):
   """Returns {URL: (size, checksum)}s stored within file."""
   checksums = {}
   for line in _read_file(checksums_path).split('\n'):
-    if not line:
+    if not line or line.strip().startswith('#'):
```
us: Commits must do a particular job. you have already made this change on #808.

gjtucker: Fixed

gjtucker:
> @gjtucker why didn't inherit anything from `Mnist` class?

I started that way, but ended up using none of the super class so changed it not to inherit.

tensorflow_datasets/image/binarized_mnist.py
gjtucker: Unfortunately, I don't know of one.

PR #814, opened 2019-07-26
Would like to contribute a new Dataset to translate directory

praveenjune17:
I have collected 170000 lines of english, tamil parallel text from http://ufal.mff.cuni.cz/~ramasamy/parallel/data/v2/en-ta-parallel-v2.tar.gz . I would like to contribute this to TFDS

License :- CC


praveenjune17:
Working on to add few more files .

PR #858, opened 2019-08-06
Enable S3 experiment for Quickdraw

20zinnm:
I just spent 14 hours downloading/extracting the dataset, only to discover it does not use S3 record naming by default. This PR changes one line from `False` to `True` since everything else seems to be implemented already.


Conchylicultor:
Note that you can already use S3 if you explicitly specify one of the future version:

```
tfds.load('quickdraw_bitmap:3.0.0', split='train')
```

20zinnm:
Thank you, that's very good to know.

Conchylicultor:
The S3 version will be made default for all dataset in the future. Currently this is just an opt-in feature.
See: https://www.tensorflow.org/datasets/datasets_versioning for info about versioning.

20zinnm:
@Conchylicultor By opt-in do you mean users must manually opt in to use it on older datasets but all new contributed datasets use it by default? Because I originally had support for both versions in the Fruits 360 PR but was told to remove the old version. Thus, I assumed that the old version was deprecated and subject to removal.

I didn't realize it could be manually enabled, so I'm going to close this PR. We should definitely consider a PR that drops all support for the old version to make the library more consistent.

Conchylicultor:
Yes, you're correct. New datasets uses it by default. Older ones will be updated gradually. The long term plan is to remove the legacy version entirely.

Issue #737, opened 2019-07-04
S3 roll-out


pierrot0:
This is about rolling out the S3 (new sharding shuffling slicing mechanism).

Currently, the hash function being used is siphash, with csiphash library, which causes tests to crash on kokoro with py3 and tf13. We don't understand why at this point, and consider switching to md5 (slower - but faster than pure python implementation of siphash - but hey, at least it's guaranteed to work everywhere).

pierrot0:
Recep, Chanchal, if you have time, please do not hesitate to make new versions with S3 support for DatasetBuilder classes which don't have that already. It should be doable for all builders, and beam ones once #677 is fixed.

Conchylicultor:
Closing this now S3 has been rolled out. 
https://github.com/tensorflow/datasets/issues/1519 to track cleanup legacy code.

PR #864, opened 2019-08-07
Launch S3 on all text datasets.

us:
Except wikipedia because it uses beam.
Related issue: #737


tensorflow_datasets/text/definite_pronoun_resolution.py
```diff
@@ -93,19 +100,23 @@ def _split_generators(self, dl_manager):
             gen_kwargs={'filepath': files['train']}),
     ]
 
+  @tfds.core.drop_key_if_not_s3
```
pierrot0: Can be removed now.

tensorflow_datasets/text/cnn_dailymail.py
```diff
@@ -219,14 +225,14 @@ class CnnDailymail(tfds.core.GeneratorBasedBuilder):
           name='bytes',
           description=('Uses byte-level text encoding with '
                        '`tfds.features.text.ByteTextEncoder`'),
-          text_encoder_config=tfds.features.text.TextEncoderConfig(
+        text_encoder_config=tfds.features.text.TextEncoderConfig(
```
Conchylicultor: Could you run `pylint` to format your workspace ? For instance why removing spaces here ?

tensorflow_datasets/text/cnn_dailymail.py
```diff
@@ -249,7 +255,7 @@ def _info(self):
     )
 
   def _vocab_text_gen(self, paths):
-    for ex in self._generate_examples(paths):
+    for fname, ex in self._generate_examples(paths):
```
Conchylicultor: If the variable is unused, then you can use `_`:

```
for _, ex in self._generate_examples(paths):
```

tensorflow_datasets/text/definite_pronoun_resolution.py
```diff
@@ -49,11 +49,18 @@
...
+  # 2.0.0: S3 with new hashing function (different shuffle).
+  # 1.0.0: S3 (new shuffling, sharding and slicing mechanism).
   BUILDER_CONFIGS = [
       tfds.core.BuilderConfig(
           name='plain_text',
           version=tfds.core.Version(
               '0.0.1', experiments={tfds.core.Experiment.S3: False}),
+          supported_versions=[
+            tfds.core.Version("2.0.0"),
+            tfds.core.Version("1.0.0"),
```
Conchylicultor: You should only add a single new version, as only the new hash function is supported now.

tensorflow_datasets/text/cnn_dailymail.py
us: I run and fixed

PR #890, opened 2019-08-09
add the300w_lp dataset

lambdaflow:
Add 300W-LP Dataset

Generated dataset info available at Gist:
https://gist.github.com/lambdaflow/9f661080abfc0c6677103a9d885e8a37

@us @Conchylicultor @cyfra


tensorflow_datasets/image/the300w_lp.py
```diff
@@ -0,0 +1,149 @@
...
+          "landmarks_3d": pt3d,
+          "roi": roi,
+          "illum_params": illum_params,
+          "color_params": color_params,
+          "tex_params": tex_params,
+          "shape_params": shape_params,
+          "exp_params": exp_params,
+          "pose_params": pose_params
+      }
+      yield image_file, record
```
pierrot0: The key (in this case image_file) should not vary from one machine to the next.
Here the path will depend on where the user is running.

You should just use the basename (+ directory name if that is not enough to get uniqueness).

tensorflow_datasets/image/the300w_lp.py
```diff
@@ -0,0 +1,149 @@
...
+
+import os
+import numpy as np
+import tensorflow as tf
+import tensorflow_datasets as tfds
+
+_DATASET_URL = "https://drive.google.com/uc?export=download&" \
+               "id=0B7OEHD3T4eCkVGs0TkhUWFN6N1k"
+
+_PROJECT_URL = "http://www.cbsr.ia.ac.cn/users/xiangyuzhu/" \
```
us: hmt'L' missing

lambdaflow: > hmt'L' missing

the project site is really ended without 'L'.
"http://www.cbsr.ia.ac.cn/users/xiangyuzhu/projects/3DDFA/main.htm"

tensorflow_datasets/image/the300w_lp.py
```diff
@@ -0,0 +1,149 @@
...
+        citation=_CITATION,
+    )
+
+  def _split_generators(self, dl_manager):
+    """Returns SplitGenerators."""
+    extracted_path = dl_manager.download_and_extract(_DATASET_URL)
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=1,
```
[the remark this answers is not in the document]
lambdaflow: Removed num_shards in d9e394d

tensorflow_datasets/image/the300w_lp.py
lambdaflow: > The key (in this case image_file) should not vary from one machine to the next.
> Here the path will depend on where the user is running.
> 
> You should just use the basename (+ directory name if that is not enough to get uniqueness).

changed to use basename as key in d9e394d, since no name conflicts in image filenames.

PR #712, opened 2019-06-27
uint16 for png encode Issue #213

Suhasnama:
Added uint16 for png encoded Image Feature

PR #892, opened 2019-08-09
Add support for uint16 PNG images

### LINKED ISSUE #213: Allow uint16 for png encoded image feature
Currently, `tfds.features.Image` with `png` encoding defaults to `tf.uint8` data type and cannot be customized. 
https://github.com/tensorflow/datasets/blob/96fe8c3156b43eb776456522ac739a0a6cdb86d2/tensorflow_datasets/core/features/image_feature.py#L149
However, `tf.image.decode_png` also allows `tf.uint16`, which is for instance important for medical data. One option would be to make the `dtype` for the decoding function configurable.
The encoding constraint at
https://github.com/tensorflow/datasets/blob/96fe8c3156b43eb776456522ac739a0a6cdb86d2/tensorflow_datasets/core/features/image_feature.py#L130 
should also be modified for `png` encoding to allow `np.uint8` as well as `np.uint16`.

lgeiger:
Currently, `tfds.features.Image` with `png` encoding defaults to `tf.uint8` data type and cannot be customised.

This PR adds support for `tf.uint16` PNG encoding, which is useful for medical data or depth estimation tasks.

Fixes #213 
Closes #712

//cc @Conchylicultor

### DIFF
--- a/tensorflow_datasets/core/features/image_feature.py
+++ b/tensorflow_datasets/core/features/image_feature.py
@@ -40,6 +40,11 @@ ACCEPTABLE_CHANNELS = {
     'jpeg': (0, 1, 3),
 }
 
+ACCEPTABLE_DTYPES = {
+    'png': [tf.uint8, tf.uint16],
+    'jpeg': [tf.uint8],
+}
+
 
 class Image(feature.FeatureConnector):
   """`FeatureConnector` for images.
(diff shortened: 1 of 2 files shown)


Conchylicultor:
Thanks for finishing this. Looks good to me. I will merge this today

Issue #884, opened 2019-08-09
Script to generate list of datasets with their versions


pierrot0:
Problem: Right now, there is no easy way to tell what are the available versions of each dataset, beside looking at the source code.

We want to ask people to pin the major component of the version they are using.

Proposed solution:
Have a script / function to generate the list of datasets with their available versions (and which is the canonical).

 We already have a function to generate the list of datasets. Maybe we can just update it.
The description of what each function is about is not available programmatically, so I would just have the versions, it's sufficient for now.

Conchylicultor:
Instead of creating a new list, I think this information should be included in the current dataset documentation instead.
This could also be added programatically. Something like: `builder.info.print_available_versions()`

Additionally, maybe release note should be part of `tfds.core.Version('1.1.0', release_note='Add new text field')`, instead of being a comment. This would allow the above functions to provide more information. And we could surface those release note when an error is raised about versioning.

pierrot0:
Agree on the release notes eventually being part of the version. But next
iteration. For now it would already be nice to have a list with versions.

PR #897, opened 2019-08-12
Add versions to list builders.

us:
Related issue:  #884

Example of output:
```
[('abstract_reasoning', ['0.0.2']),
 ('amazon_us_reviews', ['0.1.0']),
 ('bair_robot_pushing_small', ['1.0.0', '2.0.0']),
 ('bigearthnet', ['0.0.2']),
 ('binarized_mnist', ['1.0.0']),
 ('binary_alpha_digits', ['1.0.0']),
 ('caltech101', ['1.1.0', '2.0.0', '3.0.0']),
 ('caltech_birds2010', ['0.1.0']),
 ('caltech_birds2011', ['0.1.0']),
 ('cats_vs_dogs', ['2.0.1', '3.0.0', '4.0.0']),
 ('celeb_a', ['0.3.0', '1.0.0', '2.0.0']),
.
.
.
 ('wmt19_translate', ['0.0.3']),
 ('wmt_t2t_translate', ['0.0.1']),
 ('wmt_translate', []),
 ('xnli', ['0.0.1'])]

```

### DIFF
--- a/tensorflow_datasets/core/registered.py
+++ b/tensorflow_datasets/core/registered.py
@@ -141,7 +141,7 @@ class RegisteredDataset(abc.ABCMeta):
 
 
 def list_builders(versions=False):
-  """Returns the string names of all `tfds.core.DatasetBuilder`s."""
+  """Returns the string names and version of all `tfds.core.DatasetBuilder`s."""
   registry = {}
   if versions:
     for name, vers in _DATASET_REGISTRY.items():
--- a/tensorflow_datasets/core/registered.py
+++ b/tensorflow_datasets/core/registered.py
@@ -44,7 +44,6 @@ __all__ = [
 
 # Internal registry containing <str registered_name, DatasetBuilder subclass>
 _DATASET_REGISTRY = {}
-
 # Internal registry containing:
 # <str snake_cased_name, abstract DatasetBuilder subclass>
 _ABSTRACT_DATASET_REGISTRY = {}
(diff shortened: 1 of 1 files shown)


Conchylicultor:
Thanks, but I don't think we want to modify the `tfds.list_builder`. I haven't fully understood what @pierrot0 want.
For me just exposing the available versions to the markdown doc would be enough I think.

pierrot0:
@Conchylicultor: I think we need both.
Exposing version in doc is good for the stable version, but we should have a way of getting that list for head.

Maybe we could have an optional_argument to also return the versions. Would that work for you Etienne?

pierrot0:
@us Thank you Recep, I left a few comments, but overall looks good.

Conchylicultor:
@pierrot0 Where Even if we have a list of all available dataset, I don't think we should have this in `tfds.list_builder`.
 * The current API allows testing like `name in tfds.list_builders()` which I think is a more natural user API than having list_builders to return the tuple (ds_name, version).
* This is also already used multiple times, including outside our repo.
* This do not works well with BuildConfig which can have multiple different versions for the same dataset.

If you really need this, it would make more sense to add this to a new method, which would also properly support builder config.
Where would you plan to use this list for ? Do you want to access it programatically or just a txt file would be enough ?

pierrot0:
As talked offline, let's make a new function.

Conchylicultor:
The [supported.txt](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/testing/metadata/supported.txt) and [missing.txt](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/testing/metadata/missing.txt) files ca be used for this purpose. So closing this.

PR #918, opened 2019-08-15
Fix the link to the list of available datasets in the documentation

ageron:
The proper link is https://github.com/tensorflow/datasets/tree/master/docs/add_dataset.md


Conchylicultor:
Thanks, this should be fixed now so closing this. Our internal system could not import your PR due to special char in your user name :'( so I had to create an other PR. Désolé pour ça.

PR #924, opened 2019-08-16
Launch S3 to video datasets.

ChanchalKumarMaji:
Related Issue - #737.


ChanchalKumarMaji:
`moving_mnist` doesn't have a test. I will add that later.

tensorflow_datasets/video/starcraft.py
```diff
@@ -211,12 +217,14 @@ def _generate_examples(self, files):
           self._parse_single_video,
           num_parallel_calls=tf.data.experimental.AUTOTUNE)
```
Conchylicultor: The examples won't be generated in deterministic order:
* The files order in `tf.data.TFRecordDataset(files)` may not be deterministic
* The `num_parallel_calls=tf.data.experimental.AUTOTUNE` is porbably another source of un-determinism.

Would it be possible to have a more robust key instead of a counter ? If there is only a single example per file, the filename could be used as key. Something like:

```
def _make_tfrecord_dataset(filename):
  ds_record = tf.data.TFRecordDataset(os.path.join(root_dir, filename))
  ds_record = ds_record.map(self._parse_single_video)
  ds_filename = tf.data.Dataset.from_tensor(filename)
  return tf.data.Dataset.zip(ds_filename, ds_record)

ds = tf.data.Dataset.from_tensor_slice([filenames])
ds = ds.interleave(_make_tfrecord_dataset)

...
filename, video = sess.run(iterator)
```

Conchylicultor:
@ChanchalKumarMaji Any updates ? Do you encounter any issues ?

tensorflow_datasets/video/starcraft.py
ChanchalKumarMaji: In tf.data.Dataset.interleave(), I have used `cycle_length = 1`. Is it Okay ?

ChanchalKumarMaji:
ping @Conchylicultor .

Conchylicultor:
Thank you Chanchal. I think pierre finished what you started, so now all datasets have been migrated.

PR #225, opened 2019-03-11
Added cityscapes dataset

f1recracker:
Added Cityscapes dataset with segmentation labels. Fixes #214.

PR #950, opened 2019-08-27
Add 'Lost and Found' Road Hazard Dataset

hermannsblum:
This PR adds the [Lost and Found](http://www.6d-vision.com/lostandfounddataset) dataset.

I mostly followed the code from PR #225 as the datasets follow the same structure, but added a script to generate fake-data.

The gist of the `dataset.info` files is [here](https://gist.github.com/hermannsblum/cb237dfa98e897797e37c96040dd10bc).

This is my first contribution, so please excuse if I missed some step, I tried to follow the contributing guidelines closely.

I so far did not get any information about a CLA between ETH Zurich and Google. I am sure there is one, but my universities contact person will be on holidays for at least another week. If you can find out in the meantime whether there is an agreement and how I can add my email to it that would be great!


hermannsblum:
As a suggestion, it was not initially clear that I could sign individually even if my university/company did not sign a CLA, as long as I have general rights to open-source my work. The CLA behind the link made that clear, so I could sign now :+1:

tensorflow_datasets/image/lost_and_found.py
```diff
@@ -0,0 +1,225 @@
...
+        features = {
+            'image_id': image_id,
+            **{feat: path.join(paths_city_root[feat],
+                               '{}_{}.png'.format(image_id, file_suffix[feat]))
+               for feat in paths}}
+
+        yield image_id, features
+
+# Helper functions
+
```
us: Please use pylint. For TF code style check this [link](https://www.tensorflow.org/community/contribute/code_style)

tensorflow_datasets/image/lost_and_found_test.py
```diff
@@ -0,0 +1,61 @@
...
+      'image_right': 'rightImg8bit.zip',
+      'disparity_map': 'disparity.zip',
+      'gt': 'gtCoarse.zip'}
+
+
+if __name__ == "__main__":
+  tf.compat.v1.enable_eager_execution()
+
+  # create fake files
+  example_dir = ('tensorflow_datasets/testing/test_data/fake_examples/'
```
us: tfds check this as default, you can delete this.

tensorflow_datasets/image/lost_and_found_test.py
```diff
@@ -0,0 +1,61 @@
...
+                 'lost_and_found')
+  testing.test_utils.remake_dir(example_dir)
```
us: You shouldn't create fake examples on every testing. You should just create and add fake examples.

tensorflow_datasets/image/lost_and_found.py
hermannsblum: I assume you would like to reduce the 2 blank lines to 1?

I checked all files with pylint using the TF code style (following the provided script in `oss_scripts/lint.sh`), however there seems to be no linter (neither pylint nor pycodestyle) that actually checks for that. Fixed it anyways.

tensorflow_datasets/image/lost_and_found_test.py
hermannsblum: @us I did as suggested and added the data to the repository in 06f7bca. However, this adds 120 MiB to the repository. Maybe there is another solution that I am missing? E.g. creating the data as part of the package installation, or only if tests are invoked and the data is not already there?

cyfra: @hermannsblum - Thanks for the PR.

the 120 MB is too much. One way to bring it down: create the fake image that it not a noise but (for example) the simple pattern (even one-color).

This way such files would compress as png lot better.

tensorflow_datasets/image/lost_and_found.py
```diff
@@ -0,0 +1,223 @@
...
+        'image_left': base_url.format(self.builder_config.left_image_string)}
+    if 'image_right' in self.builder_config.features:
+      download_urls['image_right'] = base_url.format(
+          self.builder_config.right_image_string)
+    if 'segmentation_label' in self.builder_config.features \
+            or 'instance_id' in self.builder_config.features:
+      download_urls['gt'] = base_url.format('gtCoarse')
+    if 'disparity_map' in self.builder_config.features:
+      download_urls['disparity_map'] = base_url.format('disparity')
+    # split into two steps to save space for testing data
```
cyfra: I didn't understand this part. Can you explain more why you decided not to use download_and_extract ?

tensorflow_datasets/image/lost_and_found.py
```diff
@@ -0,0 +1,223 @@
...
+    """Returns SplitGenerators."""
+    base_url = 'http://www.dhbw-stuttgart.de/~sgehrig/lostAndFoundDataset/{}.zip'
+    download_urls = {
+        'image_left': base_url.format(self.builder_config.left_image_string)}
+    if 'image_right' in self.builder_config.features:
+      download_urls['image_right'] = base_url.format(
+          self.builder_config.right_image_string)
+    if 'segmentation_label' in self.builder_config.features \
+            or 'instance_id' in self.builder_config.features:
+      download_urls['gt'] = base_url.format('gtCoarse')
```
cyfra: AFAIK - you can pass the same URL multiple times to download_and_extract (and will do the "smart" thing).

this might make your code easier (as you can pass download_urls['segmentation_label'] and download_urls['instance_id'])

tensorflow_datasets/image/lost_and_found.py
hermannsblum: If I run `download_and_extract` the test will fail because it apparently skips the whole command during testing, instead of extracting those zip-files specified in `DL_EXTRACT_RESULTS` in `lost_and_found_test.py`.

Splitting the two steps correctly replaces the results from the download-function with `DL_EXTRACT_RESULTS` and subsequently extracts the archives in the unittest, which means that the test-data can be stored in compressed zip-archives, which saves space.

I will update the comment to explain this better.
Or is there another workaround to fix this issue?

tensorflow_datasets/image/lost_and_found.py
hermannsblum: I will update this, your suggestions is much more elegant and also better readable, thanks!

tensorflow_datasets/image/lost_and_found_test.py
hermannsblum: @cyfra Totally agree. I adjusted the random png generation in `fake_data_utils.py` to always generate pngs from a 4x4 random pattern. However, this function is potentially also used by other scripts, so should I add an argument to switch this on and off for backwards compatibility? I don't think huge png files are the wanted behaviour in any case, but I don't knwo how you usually handle this here.

hermannsblum:
@cyfra I resolved the python2 issue. The python3 issue appears to be unrelated to this code, so I merged the latest master version in. Please rerun CI.

tensorflow_datasets/image/lost_and_found_test.py
cyfra: Yes, it'd be great to have such argument (probably as default?).
But please do it in separate pull request.

Issue #27, opened 2019-01-22
[data request] Pascal VOC


rodrigob:
* Name of dataset: Pascal VOC12 with SDB extension
* URL of dataset: http://host.robots.ox.ac.uk/pascal/VOC/ and http://home.bharathh.info/pubs/codes/SBD/download.html
* License of dataset: ?? (to be investigated)
* Short description of dataset and use case(s): the classic detection and semantic segmentation dataset. Used in many works related to weakly supervised learning (since a bit easier than Coco).

Folks who would also like to see this dataset in `tensorflow/datasets`, please +1/thumbs-up so the developers can know which requests to prioritize.

ParthS007:
Hi @rsepassi, I would like to work on adding this dataset. I have gone though gitter discussions and want to get started with this Project. Can you please help me getting started and assign this issue to me.

Thanks

rsepassi:
Great! Let me know when you've accepted the collaborator invite, and I'll
assign you. The guide for adding a new dataset is here:
https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md

ParthS007:
@rsepassi I have accepted the invite. I will go through the doc and will ask here if I face any difficulty.

Thanks for the help.

ParthS007:
@rodrigob Can you please provide the exact links of training and validation datasets?

amanp592:
@rodrigob I would love to work on this dataset if anyone not working on it can you assign it to me??

ParthS007:
@amanp592 You can work on it.

Conchylicultor:
@ParthS007 @amanp592 Please note that someone already started working on this: https://github.com/tensorflow/datasets/pull/343

jpuigcerver:
Hi,

I'm the author of #343: JFYI my pull request only adds VOC2007, not VOC2012.

Conchylicultor:
@jpuigcerver oops, my bad, indeed. Sorry about that.

rodrigob:
@ParthS007 Any progress on this front ?

AdamHillier:
I needed VOC2012 for something so added it to my fork of tfds. It is simply @jpuigcerver's VOC2007 code adapted very slightly to point to the right tar files.

I've opened a PR #966 to have it merged.

As it includes only the object detection ground truths (axis aligned bounding boxes) and not the segmentation ground truths (or the SDB extension) it doesn't fully satisfy this dataset request, but it might be useful to some people.

[...]

PR #966, opened 2019-09-03
Add PASCAL VOC2012 dataset.

AdamHillier:


AdamHillier:
`dataset_info.json` gist: https://gist.github.com/AdamHillier/869b1927517b7acc42f0e472ee944e0d

tensorflow_datasets/image/__init__.py
```diff
@@ -81,4 +81,4 @@
 from tensorflow_datasets.image.the300w_lp import The300wLp
 from tensorflow_datasets.image.uc_merced import UcMerced
 from tensorflow_datasets.image.visual_domain_decathlon import VisualDomainDecathlon
-from tensorflow_datasets.image.voc import Voc2007
+from tensorflow_datasets.image.voc import Voc2007, Voc2012
```
ChanchalKumarMaji: please import in separate lines to be consistent with other imports.

tensorflow_datasets/image/voc.py
```diff
@@ -193,3 +221,78 @@ def _generate_example(self, data_path, image_id):
...
+class Voc2012(tfds.core.GeneratorBasedBuilder):
+  """Pascal VOC 2012."""
+
+  VERSION = tfds.core.Version("1.0.0",
+                              experiments={tfds.core.Experiment.S3: False})
+  SUPPORTED_VERSIONS = [
+      tfds.core.Version("3.0.0"),
+      tfds.core.Version("2.0.0"),
+  ]
+
```
ChanchalKumarMaji: For new datasets use only a single new version. Replace with only `VERSION = tfds.core.Version("1.0.0")`

tensorflow_datasets/image/voc.py
```diff
@@ -193,3 +221,78 @@ def _generate_example(self, data_path, image_id):
...
+                "is_truncated": tf.bool,
+                "is_difficult": tf.bool,
+            }),
+            "labels": tfds.features.Sequence(
+                tfds.features.ClassLabel(names=_VOC2012_LABELS)),
+            "labels_no_difficult": tfds.features.Sequence(
+                tfds.features.ClassLabel(names=_VOC2012_LABELS)),
+        }),
+        urls=[_VOC2012_URL],
+        citation=_VOC2012_CITATION)
```
ChanchalKumarMaji: `)` next line.

tensorflow_datasets/image/voc.py
```diff
@@ -193,3 +221,78 @@ def _generate_example(self, data_path, image_id):
...
+
+  def _split_generators(self, dl_manager):
+    trainval_path = dl_manager.download_and_extract(
+        os.path.join(_VOC2012_DATA_URL, "VOCtrainval_11-May-2012.tar"))
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=1,
```
ChanchalKumarMaji: no need of `num_shards`. Please remove it.

tensorflow_datasets/image/voc.py
```diff
@@ -193,3 +221,78 @@ def _generate_example(self, data_path, image_id):
...
+            gen_kwargs=dict(data_path=trainval_path, set_name="train")),
+        tfds.core.SplitGenerator(
+            name=tfds.Split.VALIDATION,
+            num_shards=1,
```
ChanchalKumarMaji: Please remove `num_shards`.

tensorflow_datasets/image/voc_test.py
```diff
@@ -36,5 +36,17 @@ class Voc2007S3Test(Voc2007Test):
...
+  DATASET_CLASS = voc.Voc2012
+  SPLITS = {
+      'train': 1,
+      'validation': 2,
+  }
+
+
+class Voc2012S3Test(Voc2012Test):
+  VERSION = 'experimental_latest'
+
```
ChanchalKumarMaji: Remove `Voc2012S3Test`. No need of this for new datasets.

AdamHillier:
@ChanchalKumarMaji could you possibly take another look at this?

AdamHillier:
Having written this, it might be nice to, instead of having a `voc2007` dataset and a `voc2012` dataset, instead have a single `voc` dataset with different builder configs (i.e. one for 2007, one for 2012), in the same way that [Coco does](https://github.com/tensorflow/datasets/blob/34f394c5445c3695f18612793785e0f343635ab1/tensorflow_datasets/image/coco.py#L101). This would reduce duplication of logic, and would also make it cleaner to add VOC segmentation data in the future, as asked for in #27.

However, it would be a breaking change as the current `voc2007` dataset would become a builder config of a `voc` dataset.

Conchylicultor:
@AdamHillier I have modified this PR to have builder config this (`voc/2007` and `voc/2012`). It is in the process of being merged soon.

jackd:
@Conchylicultor I might be failing at navigating github, but from what I can tell this PR is based on separate builders. Is your modified PR available anywhere? I'm keen to add the semantic segmentation data in a separate config. Happy to do that in a separate PR, would just like to start from the most up-to-date point.

netw0rkf10w:
@AdamHillier Thanks for your PR. As I understand, in the current version only bounding box annotations are supported and not semantic segmentation masks? Thanks.

Conchylicultor:
@netw0rkf10w This is correct. I believe there was some py2/3 incompatibility issue for mask computation when we added the dataset, so we didn't included them. It may have been fixed since, we haven't tried.

netw0rkf10w:
@Conchylicultor Thanks for the reply. Could you tell me what should be done to have the masks added? I would say this is the most popular dataset for semantic segmentation, so it's a pity it's not supported.

Conchylicultor:
I'm actually confusing this with `coco`. I'm not sure how masks are done with Voc. Don't hesitate to send a PR if you find how to add them.

This could be done as a BuilderConfig option, along with the year (`voc/2012`, `voc/2012-segmentation`)

AdamHillier:
@netw0rkf10w I have just submitted a PR #1198 to add VOC segmentation data including the SBD extension.

netw0rkf10w:
@AdamHillier Woah, thank you very much!
I saw that you added a download link for the augmented training set. Is it supported in this PR?

AdamHillier:
@netw0rkf10w Yes that PR supports the augmented training set. To use it you need to use the split `'train+sbd_train+sbd_validation'` (there isn't an explicit `'train_aug'` split because it would be a superset of the the `'train'` split and overlapping splits aren't allowed).

Issue #26, opened 2019-01-22
[data request] ADE20k


rodrigob:
* Name of dataset: ADE20k
* URL of dataset: http://groups.csail.mit.edu/vision/datasets/ADE20K/
* License of dataset: ?? (not specified on website)
* Short description of dataset and use case(s): dense stuff, things, and parts annotations over 20k images.

Folks who would also like to see this dataset in `tensorflow/datasets`, please +1/thumbs-up so the developers can know which requests to prioritize.

Anupam-tripathi:
I would like to contribute in this dataset as a good start for me.

Anupam-tripathi:
I will like to see this dataset in tensorflow/dataset and would like to contribute in this issue.

rsepassi:
Great, thanks! Let me know when you accept the collaborator invite, and I'll assign you.

Anupam-tripathi:
Ya, please assign it to me I will accept the collaborator invite.

Anupam-tripathi:
Sorry, Sir for being not active for two days. Please send me a collaborator invite for this issue I have thoroughly studied the issue and will surely make a pull request.

cyfra:
@Anupam-tripathi - you should already see your collaborator invite here:
https://github.com/tensorflow/datasets/invitations

aayu04:
i would like to see this dataset too.

aayu04:
Hi @Anupam-tripathi .please let me know if you are taking this issue or not cause i want to contribute in this issue.

Anupam-tripathi:
@aayu04 if you want you can make a pr.. I am still working on it

aayu04:
Ok

mandroid6:
@rsepassi I would like to add this dataset to tf.datasets, if no one else is actively working on it.

amanp592:
@rsepassi  Is anyone up here working on this dataset?? if not then please let me work on it.

mandroid6:
hey @amanp592, I am currently working on this dataset. Will be sending a PR in 2 days

veqtor:
This would be great!

ANONYMOUS-GURU:
@rsepassi  and @cyfra I would like to work on this. I want to contribute in this issue. Can you give me a collaborator invite.

rodrigob:
This ticket seems dead. What is blocking a merge ?

PR #980, opened 2019-09-10
added ade20k dataset

jackd:
As [requested](https://github.com/tensorflow/datasets/issues/26)


tensorflow_datasets/image/ade20k.py
```diff
@@ -0,0 +1,364 @@
...
+    """Returns SplitGenerators."""
+    base_dir = dl_manager.download_and_extract(
+        "http://data.csail.mit.edu/places/ADEchallenge/ADEChallengeData2016.zip")
+
+    dirs = '%s/ADEChallengeData2016/{data}/{split}/' % base_dir
+
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=16,
```
ChanchalKumarMaji: `num_shards` is no longer needed, it is computed automatically. Please remove it.

jackd: Excellent - I never really had a good way of choosing that anyway :)

Issue #677, opened 2019-06-14
Sorted TFrecords writer for beam


pierrot0:
tfrecords_writer will eventually replace writer in file_format_adapater module (See S3 experiment, used in imagenet, mnist and cifar).
tfrecords_writer writes records in order (sorted by key), which enables the slicing API to be stable.

beam writer still writes round-robin fashion. We should have something equivalent to tfrecords_writer, to order a pcollection by key and write to tfrecords in that order (if possible, with same shards distribution as in tfrecords_writer.

We most probably want to follow the same pattern as for tfrecords_writer:
 1- all records are written to temporary files. Total size and number of examples are recorded.
 2- Number of shards and distribution is computed.
 3- Records stored in temporary files are re-distributes to shards, while keeping order.

2 is the same as it is now. 1 and 3 should be using beam.
The number of shards for temporary storage will most probably have to be specified before the number of records or total size are known. Number of shard for temporary storage could be set to an arbitrary high number of care is taken not to open all files simultaneously when running on local machine.

pierrot0:
Let me add a few more details here.

We already have a pcollection given by the BeamBuilder. The BeamBuilder needs to be changed to return a pcollection of (key, record) tuples instead of record instances.

We need to count the number of elements and total size to compute the number of desired shards. We could be done with something like that:

```py
from apache_beam import typehints
from apache_beam.transforms import core
from apache_beam.transforms import ptransform


@typehints.with_input_types(str)
@typehints.with_output_types(typehints.Tuple[int, int])
class _CountAndSizeCombineFn(core.CombineFn):
  """CombineFn for computing PCollection length and size (bytes)."""

  def create_accumulator(self):
    return (0, 0)  # (Count, size)

  def add_input(self, accumulator, element):
    count, size = accumulator
    count += 1
    size += len(element)
    return count, size

  def add_inputs(self, accumulator, elements):
    count, size = accumulator
    count += len(elements)
    size += sum(len(element) for element in elements)
    return count, size

  def merge_accumulators(self, accumulators):
    count = sum(acc[0] for acc in accumulators)
    size = sum(acc[1] for acc in accumulators)
    return count, size

  def extract_output(self, accumulator):
    return accumulator


# To get the number of elements and size (bytes) of pcollection.
@ptransform.ptransform_fn
def get_len_size(pcollection):
  return pcollection | 'GetLenSize' >> core.CombineGlobally(
      _CountAndSizeCombineFn())
```

[...]

PR #1002, opened 2019-09-19
Initial commit for beam-tfrecords

ChanchalKumarMaji:
Related Issue - #677 .

I am following these steps - 
* The `beam_dataset` generates a `pcollection` of (key, record) tuples,
* Then we transform the `pcollection` to (_key, record) tuples where `_key` is 128 bits, then we do a ReshufflePerKey to sort the `pcollection`,
* Then we use the `example_serializer` to convert the each record in pcollection to a format suitable for writing,
* Then we transform the `pcollection` to contain only serialized `record`s and remove the `_key` as the `_key` is no longer needed,
* Then we use `_CountAndSizeCombineFn` to count the number of elements in the pcollection(which is equal to the number of examples in the particular split) and the total size of the split,
* Then we calculate the number of shards needed from the size of the split obtained above,
* Then we follow the code as in `tfrecord_writer`, by re-using `_TFRecordWriter` class.

@pierrot0 
Please tell me how should I use the writer object in line 262.


pierrot0:
Hi Chanchal,

Thanks for the update.
I was hoping you'd find the answer.

One possible way would be to use beam partition.
Once your pcollection have been sorted, partition accordingly (using same
shard boundary logic as in _TFRecordWriter).
https://beam.apache.org/documentation/transforms/python/elementwise/partition/#example-3-partition-with-multiple-arguments

You should then end-up with shard_number pcollections, which you can dump
individually.

Conchylicultor:
TFDS now supports new writer for Beam. I think we can close this.

PR #1031, opened 2019-09-25
Add B_All_White_Blood_Cancer_Microscopic_Image Dataset

yihangx:


tensorflow_datasets/image/B_All_White_Blood_Cancer_Microscopic_Image_test.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from tensorflow_datasets import testing
+from tensorflow_datasets.image \
+    import B_ALL_White_Blood_Cancer_Microscopic_Image
+
+
+class BAllWhiteBloodCancerMicroscopicImageTest(testing.DatasetBuilderTestCase):
```
Ouwen: Thoughts on shortening the name?

Ouwen:
Overall looks good, I think the name length may need to be reconsidered but that's a nit pick.

Issue #552, opened 2019-05-10
Unregister Dataset for problem ValueError: Dataset with name x already registered.


NumesSanguis:
**Is your feature request related to a problem? Please describe.**
I want to create a new dataset using `DatasetBuilder` and I'm using Jupyter Notebook to interactively write my class for `tfds.core.GeneratorBasedBuilder`. It's frustrating that if I execute the cell with that class more than 1 time, I will get the error: `ValueError: Dataset with name my_dataset already registered.` This slows down the creation of a new Dataset.

**Describe the solution you'd like**
I would like to have an `unregister_dataset()` function, which can be used like:

```
ds_name = "my_dataset"
try:
    tfds.core.unregister_dataset(ds_name)
except:
    print(f"Dataset with name '{ds_name}' not registered")
```

**Describe alternatives you've considered**
Restart kernel of the Jupyter Notebook each time and run all cells again. However, this completely removes the interactive aspect of developing with Jupyter Notebook, and therefore slowing down the development of a new Dataset.

**Additional context**
DatasetBuilder class:

```
# data access class
class MyDataset(tfds.core.GeneratorBasedBuilder):
  """TODO(my_dataset): Short description of my dataset."""

  # TODO(my_dataset): Set up version.
  VERSION = tfds.core.Version('0.1.0')

  def _info(self):
    # TODO(my_dataset): Specifies the tfds.core.DatasetInfo object
    return tfds.core.DatasetInfo(
        builder=self,
        # This is the description that will appear on the datasets page.
        description=_DESCRIPTION,
        # tfds.features.FeatureConnectors
        features=tfds.features.FeaturesDict({
            # These are the features of your dataset like images, labels ...
        }),
        # If there's a common (input, target) tuple from the features,
        # specify them here. They'll be used if as_supervised=True in
        # builder.as_dataset.
        supervised_keys=('audio', 'label'),
        # Homepage of the dataset for documentation
        urls=[],
        citation=_CITATION,
    )

  def _split_generators(self, dl_manager):
    """Returns SplitGenerators."""
    # TODO(my_dataset): Downloads the data and defines the splits
    # dl_manager is a tfds.download.DownloadManager that can be used to
    # download and extract URLs
    return [
        tfds.core.SplitGenerator(
            name=tfds.Split.TRAIN,
            # TODO(my_dataset): Tune the number of shards such that each shard
            # is < 4 GB.
            num_shards=10,
            # These kwargs will be passed to _generate_examples
[...]

PR #1049, opened 2019-10-02
T552

dhirensr:
related to: https://github.com/tensorflow/datasets/issues/552

not fully fixed ,sending a PR for review


ChanchalKumarMaji:
@Conchylicultor this PR is not a `dataset request`.

dhirensr:
@Conchylicultor @ChanchalKumarMaji : can you'll still review this?

Conchylicultor:
Thank you very much for your contribution. This looks good (your test are simply failing because you set `SKIP_REGISTERING= False` instead of `True`).
However after internal discussion, we decided an alternative implementation (https://github.com/tensorflow/datasets/pull/1092) which will be more transparent for the user.

Issue #1053, opened 2019-10-04
Install fails MACOS virtualenv python3


giulbia:
Hi,

I am trying to build a virtualenv with tensorflow-datasets, it fails on `psutil`: `Building wheels for collected packages: psutil  Building wheel for psutil (setup.py) ... error`

**Environment information**
* Operating System: macOS 10.14.6
* Python version: 3.7 & 3.6
* `tensorflow-datasets`/`tfds-nightly` version: tensorflow-datasets 1.2.0
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: tensorflow 1.14

**Reproduction instructions**

requirements.txt:
```
jupyter==1.0.0
matplotlib==3.0.2
numpy==1.15.4
scipy==1.2.0
Pillow==5.3.0

# CPU version of TensorFlow 1.14:
tensorflow==1.14

# Tensorflow Privacy 0.1.0 released 2 oct 2019 https://pypi.org/project/tensorflow-privacy/
tensorflow-privacy==0.1.0

# tensorflow-datasets
tensorflow-datasets==1.2.0

```
Create virtualenv:
```
virtualenv --python=python3.7 tfp
source ./tfp/bin/activate
pip3 install --upgrade -r requirements.txt
```

**Link to logs**
I followed up on a [psutil open issue](https://github.com/giampaolo/psutil/issues/1572#issuecomment-538379700). All the previous packages are correctly installed.

Here is [full stack trace](https://gist.github.com/giulbia/bc97a7a82a6160d4bec0cc2dd159dade)

Thanks

us:
Hello @giulbia, thanks for reporting.
Please send all error message. You can use gist.

giulbia:
Looks like it!

`Successfully installed tfds-nightly-1.2.0.dev201910180106`

Thx

PR #1057, opened 2019-10-07
Use shutil.disk_usage instead of psutil.disk_usage for Python 3.3+

guillaumekln:
I would like to propose this change to avoid the `psutil` dependency if possible. It requires a C compiler and Python development files which are not always readily available.

Related to #1053.


guillaumekln:
Any update on this one?

Conchylicultor:
Looks good to me. Thank you for your contribution. Merging this soon.

PR #1061, opened 2019-10-07
Fix import error

lijunzh:
At for now, TensorFlow 2.0 (stable) does not allow import `io` module. Only reference it after import the `tensorflow` package.


lijunzh:
@googlebot I fixed

Conchylicultor:
Thank you for fixing this. Merging this soon.

PR #1069, opened 2019-10-10
Fix 'label' in panoptic_objects

netw0rkf10w:
`panoptic_objects` have incorrect `label` IDs (which are IDs of the semantic classes).

### DIFF
--- a/tensorflow_datasets/image/coco.py
+++ b/tensorflow_datasets/image/coco.py
@@ -401,7 +401,7 @@ class Coco(tfds.core.GeneratorBasedBuilder):
               'id': instance['id'],
               'area': instance['area'],
               'bbox': build_bbox(*instance['bbox']),
-              'label': categories_id2name[instance['category_id']],
+              'label': instance['category_id'],
               'is_crowd': bool(instance['iscrowd']),
           } for instance in instances]
       }


netw0rkf10w:
@googlebot I signed it!

Conchylicultor:
Could you provide more context about this ? I'm not sure about this one.
Our `ClassLabel` feature connector only accept labels ids between `0 <> num_class`. If the `instance['category_id']` do not capture all ids (ex: there are gaps), then the labels<>id mapping won't match anymore.
There is a plan to support multi-name for a single label to allow use case like this one (`features["label"].int2names(2)  # {'name': 'door', 'coco_id': 12}`).

netw0rkf10w:
@Conchylicultor Thanks for your message.
In the `panoptic_objects`, the `label` attribute is an array containing the semantic classes (i.e. the categories defined by COCO) of the objects, is that correct? If that's the case then I think the values of `label` in the current implementation are not correct. Let me show this with an example.

Read the dataset and display the `panoptic_objects` information for some image (here image id 68459):

```
dataset, info = tfds.load('coco/2017_panoptic:1.1.0', with_info=True)
# get the second sample (this looks stupid but I don't know a better way)
for sample in dataset['train'].take(2):
    sample = sample
# sample now corresponds to the image 68459
# pretty print the objects
pprint(sample['panoptic_objects'])
```

Results:

> {'area': <tf.Tensor: id=1313, shape=(6,), dtype=int64, numpy=array([  6694,  11550,  17862,   1138, 127406, 141939])>,
>  'bbox': <tf.Tensor: id=1314, shape=(6, 4), dtype=float32, numpy=
> array([[0.15416667, 0.803125  , 0.55833334, 0.8859375 ],
>        [0.13125   , 0.0375    , 0.6041667 , 0.228125  ],
>        [0.15625   , 0.1984375 , 0.78541666, 0.3875    ],
>        [0.11875   , 0.2375    , 0.25416666, 0.29375   ],
>        [0.51458335, 0.        , 1.        , 1.        ],
>        [0.        , 0.        , 0.6354167 , 1.        ]], dtype=float32)>,
>  'id': <tf.Tensor: id=1315, shape=(6,), dtype=int64, numpy=array([ 5596759,  5790298,  8092289, 10131586, 10392688,  7884593])>,
>  'is_crowd': <tf.Tensor: id=1316, shape=(6,), dtype=bool, numpy=array([False, False, False, False, False, False])>,
>  'label': <tf.Tensor: id=1317, shape=(6,), dtype=int64, numpy=array([  0,   0,   0,  38,  97, 131])>}

For clarity below are the `id` and `label`:

> 'id': [ 5596759,  5790298,  8092289, 10131586, 10392688,  7884593]
> 'label': [  0,   0,   0,  38,  97, 131]

First observation is that `0` is not a category defined in COCO, so this doesn't seem to be good. 

Now let's check the original annotation json:

```
import os
import json
json_file = os.path.expanduser('~/tensorflow_datasets/downloads/extracted/ZIP.image.cocod.org_annot_panop_annot_trainwF920hKba1Yetw7-FucAbfYvc_uSiJEy03O52Q4xo3A.zip/annotations/panoptic_train2017.json')
with open(json_file) as f:
    annotations = json.load(f)
# Now get the annotations of the image 68459
for ann in annotations['annotations']:
    if ann['image_id'] == 68459:
        segments_info = ann['segments_info']
        break
# Get the category id for each object 
ids = []
category_ids = []
for info in segments_info:
    ids.append(info['id'])
    category_ids.append(info['category_id'])
print('ids:', ids)
print('category_ids:', category_ids)
```

Results:

> ids: [5596759, 5790298, 8092289, 10131586, 10392688, 7884593]
> category_ids: [1, 1, 1, 43, 145, 199]

The category ids are completely different from  `sample['panoptic_objects']['label']`.

So if this is not a bug, then I misunderstood about `sample['panoptic_objects']['label']`. In this case, what does this array represent, and how to retrieve the original COCO category IDs?

Thank you in advance for your responses.

Conchylicultor:
For label being `[ 0, 0, 0, 38, 97, 131]`, can you try to convert them into string ?
```
_, ds_info = tfds.load('coco', with_info=True)

for label in [ 0, 0, 0, 38, 97, 131]:
  ds_info.features['label'].int2str(label)
```
You should see the correct string label. As explain above, Our `ClassLabel` feature connector only accept labels ids in the range `[0, num_class[`, so we are force to map the original coco mapping to our label.
One reason for having a `[0, num_class]` mapping is that it match the softmax ids of the last layer of a neural network.
You can get the full label list using: `ds_info.features['label'].name`.

To resolve this use case, we plan to add support for multi name mapping, to have name contains both the original coco id, and our remapping.

netw0rkf10w:
@Conchylicultor Thanks. It should be `ds_info.features['panoptic_objects']['label']` instead of `ds_info.features['label']`. Now I understand how it works. However, I think there are 2 issues with the current mapping:

1. It's confusing. This can be resolved by including coco_id in the mapping as you have proposed, and of course by better documentation. A solution I thought of is to add missing labels and give them unique names. For example:
```
0: void0
1: person
...
95: bridge
96: void96
97: void97
...
```
but maybe this is too cumbersome.

2. It ignores the `void` label. This is *critical* because it forces the model to predict undefined pixels (e.g. background) with one of the given labels, which will harm the performance. Moreover, it requires additional hack to deal with the unlabeled pixels (with black color) since they are not handled by the mapping.

```
# when defining the network
void_label = info.features['panoptic_objects']['label'].num_classes # 133
output_dim = void_label + 1
model = SomeModel(output_dim=output_dim)

# when processing the data
mask_panoptic = datapoint['panoptic_image']
mask = tf.dtypes.cast(mask_panoptic, tf.int64)
mask = mask[:, :, 0] + 256 * mask[:, :, 1] + (256 * 256) * mask[:, :, 2] # convert colors to IDs
# here map the panoptic IDs to category IDs, and in addition: map 0 to 133
...
```
To resolve this issue I would propose to add `0: void` to the mapping as well.

Conchylicultor:
Closing this as the solution seems to add multi-label mapping.

lucasdavid:
This dataset is very difficult to use, and I'm facing this issue as well! Some `info.features['panoptic_objects']['label']` are correctly mapped into category_ids, whereas others are plain wrong. For example: umbrella and person are correct, while the chair `category_id` is being assigned `toilet`.

image

---
If this is of any help to someone, here is how I fixed it:

1. I downloaded the train/val 2017 panoptic annotations from the official site ([cocodataset.org/#download](http://images.cocodataset.org/annotations/panoptic_annotations_trainval2017.zip)) and placed it inside the `coco17_panoptic` folder.
2. I wrote a function that loads both TFDS categories and COCO 2017 panoptic categories, and instantiates a map that translates the later to the former:
    ```py
    ANNOTATIONS = None
    CATEGORIES = None

    def load_annotations(config):
      global ANNOTATIONS, CATEGORIES

      data_dir = dig(config, 'load.data_dir')

      print('  loading category decode map...', end=' ')
      tfds_cats_file = os.path.join(data_dir, 'coco/2017_panoptic/1.1.0/panoptic_objects-label.labels.txt')
      coco_cats_file = os.path.join(data_dir, 'coco/2017_panoptic/panoptic_coco_categories.json')
      annotations_dir = os.path.join(data_dir, 'coco/2017_panoptic/annotations')

      with open(tfds_cats_file) as f:
        tfds_cats = [c.strip() for c in f.readlines()]
        tfds_cats = dict(zip(tfds_cats, range(len(tfds_cats))))

      with open(coco_cats_file) as f:
        coco_cats = json.load(f)
        coco_cats = {p['id']: p['name'] for p in coco_cats}

      CATEGORIES = {_id: tfds_cats[name] for _id, name in coco_cats.items()}

      with open(os.path.join(annotations_dir, 'panoptic_train2017.json')) as f:
        data = json.load(f)
        ANNOTATIONS = {e['image_id']: e for e in data['annotations']}
      with open(os.path.join(annotations_dir, 'panoptic_val2017.json')) as f:
        data = json.load(f)
        ANNOTATIONS.update({e['image_id']: e for e in data['annotations']})
    ```
3. And a function that converts the panoptic image into a segmentation map:
    ```py
    def instance_segmentation_from_coco17_panoptic(entry):
      iid = entry['image/id']
      image = entry['image']
      label = entry['panoptic_objects']['label']
      bbox = entry['panoptic_objects']['bbox']
      panoptic_image = entry['panoptic_image']

      segm = tf.numpy_function(panoptic_to_segm, [iid, panoptic_image], tf.int32)
      segm.set_shape([None, None])

      return image, label, bbox, segm

    def panoptic_to_segm(iid, panoptic_image):
      global ANNOTATIONS, CATEGORIES

      pan = rgb2id(panoptic_image)
      segm = -1*np.ones(panoptic_image.shape[:2], dtype=np.int32)

      for segm_info in ANNOTATIONS[iid]['segments_info']:
        cat_id = segm_info['category_id']
        segm[pan == segm_info['id']] = CATEGORIES[cat_id]

      return segm

    def rgb2id(color):
      color = color.astype(np.uint32)
      return color[..., 0] + 256 * color[..., 1] + 256**2 * color[..., 2]
    ```
4. Finally, this can just be added to the pipeline:
    ```py
    (train, valid), info = tfds.load('coco/2017_panoptic', ...)

    train.map(instance_segmentation_from_coco17_panoptic).padded_batch(...).prefetch(...)
    ```

PR #1187, opened 2019-11-10
Update file_format_adapter.py

nikochiko:
Switch deprecated `tensorflow.python.lib.io.tf_record.tf_record_iterator` with `tf.data.TFRecordDataset`.


Conchylicultor:
Thank you for your contribution. Unfortunately, this won't be compatible for people using graph mode.
I think you could try with:
```
ds = tf.data.TFRecordDataset(path)
ds = dataset_utils.as_numpy(ds, graph=tf.Graph())
```
and check if the tests still pass.

Conchylicultor:
Closing this as it seems obsolete. The `file_format_adapter` class is gonna be removed soon in favor of the new reader class.

PR #1190, opened 2019-11-12
Fix error: replace min with max

ageron:
`min` will always return 1. `max` will return the correct value (I think).


Conchylicultor:
Thanks, merging now.

PR #1197, opened 2019-11-15
Add the ARC dataset

ageron:
This dataset is located at https://github.com/fchollet/ARC and described in François Chollet's paper [The Measure of Intelligence](https://arxiv.org/pdf/1911.01547.pdf).

**Edit: what follows is outdated, see the comments below:**

Notes about this implementation:
* In the original ARC dataset, image sizes vary from 1x1 to 30x30. However, the `Example` protobuf format does not support such variation, so I stored all images as 30x30, filling up with 0, and I added a separate field for the image size, for each image. For example:

```
Original ARC image pair:
 
{
  "input": [[1,2],[3,4],[5,6]],  # 3x2
  "output": [[1,2,3,4],[5,6,7,8]]  # 2x4
}
 
TFDS ARC image pair:
 
{
  "input": [[1,2,0,0,...,0],[3,4,0,0,...,0],[5,6,0,0,...,0], [0,0,...,0], ..., [0,0,...,0]],  # 30x30
  "input_size": [3,2],
  "output": [[1,2,3,4,0,0,...,0],[5,6,7,7,0,0,...,0],[0,0,...,0], ..., [0,0,...,0]]  # 30x30
  "output_size": [2,4],
}
```

Here's a [Colab notebook](https://colab.research.google.com/drive/141ceaZhdps3JQj7QvK5Piibasdt7LZNy) where you can try out this TFDS dataset.


tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,158 @@
...
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TEST,
+            num_shards=10,
+            gen_kwargs={
+                "directory": test_dir,
+            }),
+    ]
+
+  def _to_image(self, data):
+      img = np.zeros(shape=[30, 30], dtype=np.uint8)
```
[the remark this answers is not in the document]
cyfra: Also mention what is the type of data - both input and output.

tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,158 @@
...
+      height = len(data)
+      width = len(data[0])
+      for index, row in enumerate(data):
+          width = max(width, len(row))
+          # Temporary hack to replace null values with 5 (gray).
+          # See https://github.com/fchollet/ARC/pull/2
+          #                              vvvvvvvvvvvvvvvvvvvvvvvvvvv
+          img[index, :len(row)] = [value if value is not None else 5
+                                   for value in row]
```
cyfra: Could you do this using np arrays ?

Something like: 

img[:height,:width] = data
img = np.where(img, img, 5)

tensorflow_datasets/image/arc_test.py
```diff
@@ -0,0 +1,20 @@
+"""TODO(arc): Add a description here."""
```
cyfra: please fix.

cyfra:
The tests are failing on Python2:

  File "<string>", line 1, in <module>
  File "tensorflow_datasets/__init__.py", line 52, in <module>
    from tensorflow_datasets import image
  File "tensorflow_datasets/image/__init__.py", line 20, in <module>
    from tensorflow_datasets.image.arc import ARC
  File "tensorflow_datasets/image/arc.py", line 49
    _DL_URL = _BASE_URL + f"zipball/{_COMMIT}/"
                                              ^
SyntaxError: invalid syntax

ageron:
Thanks @cyfra , great code review. I'll try to answer & fix everything today.

tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,158 @@
...
+_DL_RESOURCE = tfds.download.Resource(
+    url=_DL_URL,
+    extract_method=tfds.download.ExtractMethod.ZIP)
+_EXTRACT_SUBDIR = f"fchollet-ARC-{_COMMIT}"
+
+class ARC(tfds.core.GeneratorBasedBuilder):
+  """The Abstraction and Reasoning Corpus (ARC)."""
+
+  VERSION = tfds.core.Version(
+      "1.0.0", experiments={tfds.core.Experiment.S3: False})
```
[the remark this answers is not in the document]
ageron: I followed the [documentation on adding a new dataset](https://www.tensorflow.org/datasets/add_dataset#use_the_default_template), and the generated code has S3 disabled. I'm honestly didn't even notice, I just thought that it was there for a good reason. I'll turn it back on.

tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,158 @@
...
+  SUPPORTED_VERSIONS = [
+      tfds.core.Version(
+          "3.0.0", "New split API (https://tensorflow.org/datasets/splits)"),
```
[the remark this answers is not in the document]
ageron: Mmh, I guess I copy/pasted an existing dataset. I'll revert back down to 1.0.0.

tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,158 @@
...
+    # download and extract URLs
+    extracted_dir = dl_manager.download_and_extract(_DL_RESOURCE)
+    data_dir = os.path.join(extracted_dir, _EXTRACT_SUBDIR, "data")
+    train_dir = os.path.join(data_dir, "training")
+    test_dir = os.path.join(data_dir, "evaluation")
+
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=10,
```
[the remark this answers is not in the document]
ageron: Sounds good.

tensorflow_datasets/image/arc.py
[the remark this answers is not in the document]
ageron: Done

tensorflow_datasets/image/arc_test.py
ageron: Done

tensorflow_datasets/image/arc.py
ageron: Sure

ageron:
I fixed the code according to @cyfra's feedback (thanks again!):

* use the default S3 support and remove `num_shards`
* reverted version back to 1.0.0
* added missing docstrings
* simplified the function that converts lists of lists of ints to padded NumPy arrays, by using NumPy array primitives instead of Python loops
* replaced `f"foo{bar}"` with `"foo{}".format(bar)` to support Python 2 (after all it's still alive for a few more weeks).

I also made the following changes:
* updated the dataset to the latest version from fchollet/ARC (commit bd9e2c9)
* stop replacing None values with 5s: grids are not supposed to contain any
* stop supporting grids containing rows with different lengths: that's not supposed to happen
* instead, catch exceptions when loading individual tasks and just log an error message.
* rename "image" (and "img") to "grid", since this is the official term used in the ARC dataset
* moved a couple functions up to the module level, just in case someone needs them: `pad_grid()` (formerly named `_to_image()`) and `pad_grid_pairs()` (formerly named `_prepare_pairs()`).
* added a `crop_padded_grid()` function since most people will probably want to do this.

That said, since `tf.data.Dataset` now has support for ragged tensors, I'm wondering whether TFDS also has support? If so, it may be best to represent lists of grids as ragged tensors, rather than going the padding + size route. I can't find any documentation about ragged tensor support in TFDS, but the source code does contain a few references to ragged tensors, so I'm guessing it doesn't yet but it's on the roadmap?

Conchylicultor:
tfds supports ragged tensor if you're nesting sequences.

```
{
    'nested': tfds.features.Sequence(tfds.features.Sequence(tf.int32))
}
```

```
yield {
    'nested': [
        [1, 2, 3],
        [],
        [1, 2, 3, 4, 5]
    ],
}
```

Would this satisfy your use case ? Otherwise, how would you like to declare your features ?

ageron:
Reimplemented to use ragged tensors. Works great! :)

tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,126 @@
...
+_DESCRIPTION = """
+ARC can be seen as a general artificial intelligence benchmark, as a program
+synthesis benchmark, or as a psychometric intelligence test. It is targeted at
+both humans and artificially intelligent systems that aim at emulating a
+human-like form of general fluid intelligence.
+"""
+
+
+_BASE_URL = "https://github.com/fchollet/ARC/"
+_COMMIT = "bd9e2c934c83d00251b7b4781ffc38cd167c885f"
```
Conchylicultor: As the data seems to be evolving, the user should be able to select the version he wants.
Please use `BuilderConfig` which contains the commit, so that users is not forced to use the default commit.

tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,126 @@
...
+            }),
+            "test": tfds.features.Sequence({
+                "input": tfds.features.Sequence(tfds.features.Sequence(tf.int32)),
+                "output": tfds.features.Sequence(tfds.features.Sequence(tf.int32)),
+            })
+        }),
+        # If there's a common (input, target) tuple from the features,
+        # specify them here. They'll be used if as_supervised=True in
+        # builder.as_dataset.
+        supervised_keys=None,
```
Conchylicultor: `('train', 'test')`

tensorflow_datasets/image/arc.py
```diff
@@ -0,0 +1,126 @@
...
+            with tf.io.gfile.GFile(json_path) as f:
+                task = json.load(f)
+            task_id = os.path.basename(json_path)[:-len(".json")]
+            yield task_id, {
+                "task_id": task_id,
+                "train": task["train"],
+                "test": task["test"],
+            }
+        except Exception as ex:
+            logging.exception("Could not load task {}.".format(task_id))
```
Conchylicultor: Why is it required ? Which tasks cannot be loaded ?

It be much more preferable to have an explicit filter for those tasks rather than catch any exception.

Having broad exception catching is bad practice, as it would remove exception which shouldn't have been. (ex: no more disk space, file not found.)

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -359,6 +359,9 @@ def _bytes_flatten(element):
         element = element.decode("latin-1")
       element = element.encode("utf-8")
       ret += element
+    elif isinstance(element, (tf.RaggedTensor,
+                              tf.compat.v1.ragged.RaggedTensorValue)):
+      ret += str(element).encode("utf-8")
```
Conchylicultor: Could you add a TODO:
```
# TODO(tfds): Checksum should be computed from the `ragged.to_list()` value.
```

or try with
```
try with `str(element.to_list())
```
instead ?

tensorflow_datasets/image/arc.py
ageron: It's a bit more complicated, I think: input would be all pair inputs + all train pair outputs, and test would be all test outputs. I'm not sure this can be done with TFDS, unless I restructure the dataset entirely.

tensorflow_datasets/image/arc.py
ageron: The dataset is constantly evolving, and the first versions had some errors (null values). If the user can select the commit they want, then this is not needed anymore, as the user can always pick a commit that works.

tensorflow_datasets/testing/dataset_builder_testing.py
ageron: Sure

On a separate note: I noticed that `_bytes_flatten()` returns a different value every time I call it with the same string tensor (e.g., `tf.constant([b'foo', b'bar'])`). Even calling `numpy()` on the tensor doesn't help.

tensorflow_datasets/image/arc.py
Conchylicultor: Sounds good, in this case, please remove the try/except block, and use a commit which works for the default `BuilderConfig()`

tensorflow_datasets/testing/dataset_builder_testing.py
Conchylicultor: `_bytes_flatten` shouldn't be called directly on tensor, because nothing guarantee that the __str__ representation is constant.
This is why it should first be converted to values which always produce deterministic output.

`ret += tensor.numpy().tobytes()` or `ret += _bytes_flatten(tensor.numpy())` should works.

tensorflow_datasets/image/arc.py
Conchylicultor: I see, current version looks good, no need to change it, thanks.

tensorflow_datasets/image/arc.py
ageron: I just made this change, works fine. However, I'm not sure how to handle the checksums, since each commit will have a different checksum. Will the user have to run the `tensorflow_datasets.scripts.download_and_prepare` script with the `--register_checksums` option the first time they use a new commit? How exactly would that work? Would they have to define a new dataset?

tensorflow_datasets/image/arc.py
ageron: I looked into this a bit. Currently, the structure of each task is:

```python
{
    "train": {
        "input": input_grids,
        "output": output_grids,
    },
    "test": {
        "input": input_grids,
        "output": output_grids,
    }
}
```

In order to make this dataset compatible with `as_supervised=True`, I would have to restructure it to something like this:

```python
{
    "input": {
        "train_input": train_input_grids,
        "train_output": train_output_grids,
        "test_input": test_input_grids,
    },
    "target": {
        "test_output": test_output_grids,
    }
}
```

It's doable, but a bit ugly IMHO, and it doesn't respect the original dataset's structure, plus I'm not sure it really adds much value for the user, since this dataset is so different from regular image tasks (in fact, I'm not even sure it really belongs in the `image` category, perhaps there should be another category for special datasets like this one, that don't really fit in any category, or a category for abstract images, or reasoning).

tensorflow_datasets/image/arc.py
ageron: I removed the exception handling.

tensorflow_datasets/testing/dataset_builder_testing.py
ageron: That's what I thought, too, but I think there must be a bug in NumPy, because it is not the case. Try this:

```python
import numpy as np
import tensorflow as tf

for _ in range(5):
    obj = np.array([b"foo", b"bar"])
    print(bytes(obj))
print()

for _ in range(5):
    obj = tf.constant([b"foo", b"bar"])
    print(bytes(obj))
print()

for _ in range(5):
    obj = tf.constant([b"foo", b"bar"]).numpy()
    print(bytes(obj))
print()

for _ in range(5):
    obj = tf.constant([b"foo", b"bar"]).numpy()
    print(obj.tobytes())
print()
```

ageron:
Hi @Conchylicultor ,
Thanks for your great code review. 👍 
I updated the code:
* Added a BuilderConfig to allow the user to select any commit
* Updated `_bytes_flatten()` to use `ret += str(element.to_list()).encode("utf-8")`
* Removed the excessive exception handling
* Moved test data to a directory with a clearly fake commit id, to avoid any confusion

However, I did not set `supervised_keys` to `('train', 'test')`, since the test input grids are part of the input, so I would have to restructure tasks completely, which would make it too different from the original dataset.

tensorflow_datasets/image/arc.py
Conchylicultor: Yes, it's better to keep the original dataset structure. Keeping the following is better
```
{
    "train": {
        "input": input_grids,
        "output": output_grids,
    },
    "test": {
        "input": input_grids,
        "output": output_grids,
    }
}
```

tensorflow_datasets/testing/dataset_builder_testing.py
Conchylicultor: Interesting. Thanks for catching this. It's good to know.
I think the issue is that `tf.constant([b"foo", b"bar"]).numpy().dtype == object` while `np.array([b"foo", b"bar"]).dtype = 'S3'`.
Opening a bug, even if it's probably not a huge priority.

ageron:
Hi @Conchylicultor , is every good to go now? Github still says "1 change requested" but I can't see what change that is.

Conchylicultor:
> Hi @Conchylicultor , is every good to go now? Github still says "1 change requested" but I can't see what change that is.

Don't worries about what Github say, it should be merged soon. Thanks again for your contribution.

ageron:
Awesome, thanks for your help @cyfra and @Conchylicultor !

Conchylicultor:
Actually after reading the description more, it seems that the inputs are variable dim images. So `tf.RaggedTensor` do not seems to be the best input format.
Nested sequence should be used when each sequence has a different variable length (ex: `[[1, 2, 3], [], [4]]`).
Here the best format should be variable length image:
`tfds.features.Image(shape=(None, None, 1))`

Could you update the dataset ?
```
yield {
    'train': {
        # [..., np.newaxis] adds a new dimension at the end.
        'input': np.array([[1,2],[3,4],[5,6]], dtype=np.uint8)[..., np.newaxis],
    }
}
```

ageron:
Hi @Conchylicultor ,

I tried to make the grids actual images, as you suggested, but I couldn't get it to work. Each task actually contains a *sequence* of grid pairs, which may all have different sizes. So instead of...

```python
yield {
    'train': {
        'input': np.array([[1,2],[3,4],[5,6]], dtype=np.uint8)[..., np.newaxis],
    }
}
```

...it's actually more like this:

```python
yield {
    'train': [
        {
            'input': np.array([[1,2],[3,4],[5,6]], dtype=np.uint8)[..., np.newaxis],
            'output': np.array([[1],[2],[3],[4]], dtype=np.uint8)[..., np.newaxis],
        },
        {
            'input': np.array([[1,2,3],[4,5,6]], dtype=np.uint8)[..., np.newaxis],
            'output': np.array([[1]], dtype=np.uint8)[..., np.newaxis],
        },
    ],
    'test': [
        {
            'input': np.array([[1,2,3,4,5],[6,7,8,9,0]], dtype=np.uint8)[..., np.newaxis],
            'output': np.array([[1,2],[3,4]], dtype=np.uint8)[..., np.newaxis],
        },
    ]
}
```

Therefore, I tried this:

```python
features=tfds.features.FeaturesDict({
    "train": tfds.features.Sequence({
        "input": tfds.features.Image(shape=(None, None, 1), dtype=tf.uint8),
        "output": tfds.features.Image(shape=(None, None, 1), dtype=tf.uint8),
    }),
    "test": tfds.features.Sequence({
        "input": tfds.features.Image(shape=(None, None, 1), dtype=tf.uint8),
        "output": tfds.features.Image(shape=(None, None, 1), dtype=tf.uint8),
    })
}),
```

However, I got an `InvalidArgumentError` error when running the tests:

```
ERROR: test_download_and_prepare_as_dataset (__main__.ArcTest)
test_download_and_prepare_as_dataset (__main__.ArcTest)
Run the decorated test method.
----------------------------------------------------------------------
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/testing/test_utils.py", line 198, in decorated
    f(self, *args, **kwargs)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/testing/dataset_builder_testing.py", line 251, in test_download_and_prepare_as_dataset
    self._download_and_prepare_as_dataset(builder)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/testing/dataset_builder_testing.py", line 286, in _download_and_prepare_as_dataset
    builder.download_and_prepare(download_config=download_config)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/api_utils.py", line 52, in disallow_positional_args_dec
    return fn(*args, **kwargs)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_builder.py", line 339, in download_and_prepare
    self.info.compute_dynamic_properties()
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_info.py", line 282, in compute_dynamic_properties
    self._compute_dynamic_properties(self._builder)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_info.py", line 295, in _compute_dynamic_properties
    builder, split_name)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_info.py", line 503, in get_dataset_feature_statistics
    for example in utils.tqdm(np_dataset, unit=" examples", leave=False):
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tqdm/std.py", line 1081, in __iter__
    for obj in iterable:
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_utils.py", line 160, in _eager_dataset_iterator
    for item in dataset:
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 622, in __next__
    return self.next()
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 666, in next
    return self._next_internal()
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 651, in _next_internal
    output_shapes=self._flat_output_shapes)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/ops/gen_dataset_ops.py", line 2673, in iterator_get_next_sync
    _six.raise_from(_core._status_to_exception(e.code, message), None)
  File "<string>", line 3, in raise_from
tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __inference_Dataset_map_<class 'functools.partial'>_895}} PartialTensorShape: Incompatible shapes during merge: [7,18,1] vs. [19,29,1]
	 [[{{node sequence_decode_2/TensorArrayV2Stack/TensorListStack}}]] [Op:IteratorGetNextSync]
```

So I tried putting the sequence at a deeper level in the structure:

```python
   features=tfds.features.FeaturesDict({
       "train": {
           "input": tfds.features.Sequence(tfds.features.Image(
                        shape=(None, None, 1), dtype=tf.uint8)),
           "output": tfds.features.Sequence(tfds.features.Image(
                        shape=(None, None, 1), dtype=tf.uint8)),
       },
       "test": {
           "input": tfds.features.Sequence(tfds.features.Image(
                        shape=(None, None, 1), dtype=tf.uint8)),
           "output": tfds.features.Sequence(tfds.features.Image(
                        shape=(None, None, 1), dtype=tf.uint8)),
       },
   }),
```

I obviously adjusted the way I yield examples:

```python
```python
yield {
    'train': {
        'input': [
            np.array([[1,2],[3,4],[5,6]], dtype=np.uint8)[..., np.newaxis],
            np.array([[1,2,3],[4,5,6]], dtype=np.uint8)[..., np.newaxis],
        ],
        'output': [
            np.array([[1],[2],[3],[4]], dtype=np.uint8)[..., np.newaxis],
            np.array([[1]], dtype=np.uint8)[..., np.newaxis],
        ],
    },
    'test': {
        'input': [
            np.array([[1,2,3,4,5],[6,7,8,9,0]], dtype=np.uint8)[..., np.newaxis],
        ],
        'output': [
            np.array([[1,2],[3,4]], dtype=np.uint8)[..., np.newaxis],
        ]
    },
}
```

However, again I got an `InvalidArgumentError`:

```
ERROR: test_download_and_prepare_as_dataset (__main__.ArcTest)
test_download_and_prepare_as_dataset (__main__.ArcTest)
Run the decorated test method.
----------------------------------------------------------------------
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/testing/test_utils.py", line 198, in decorated
    f(self, *args, **kwargs)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/testing/dataset_builder_testing.py", line 251, in test_download_and_prepare_as_dataset
    self._download_and_prepare_as_dataset(builder)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/testing/dataset_builder_testing.py", line 286, in _download_and_prepare_as_dataset
    builder.download_and_prepare(download_config=download_config)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/api_utils.py", line 52, in disallow_positional_args_dec
    return fn(*args, **kwargs)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_builder.py", line 339, in download_and_prepare
    self.info.compute_dynamic_properties()
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_info.py", line 282, in compute_dynamic_properties
    self._compute_dynamic_properties(self._builder)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_info.py", line 295, in _compute_dynamic_properties
    builder, split_name)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_info.py", line 503, in get_dataset_feature_statistics
    for example in utils.tqdm(np_dataset, unit=" examples", leave=False):
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tqdm/std.py", line 1081, in __iter__
    for obj in iterable:
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_utils.py", line 160, in _eager_dataset_iterator
    for item in dataset:
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 622, in __next__
    return self.next()
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 666, in next
    return self._next_internal()
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 651, in _next_internal
    output_shapes=self._flat_output_shapes)
  File "/Users/ageron/miniconda3/envs/tf2/lib/python3.7/site-packages/tensorflow_core/python/ops/gen_dataset_ops.py", line 2673, in iterator_get_next_sync
    _six.raise_from(_core._status_to_exception(e.code, message), None)
  File "<string>", line 3, in raise_from
tensorflow.python.framework.errors_impl.InvalidArgumentError: {{function_node __inference_Dataset_map_<class 'functools.partial'>_895}} PartialTensorShape: Incompatible shapes during merge: [7,18,1] vs. [19,29,1]
	 [[{{node sequence_decode_2/TensorArrayV2Stack/TensorListStack}}]] [Op:IteratorGetNextSync]
```

I haven't dug into why this fails, but perhaps this is simply a bug, since I really feel like these methods should both work. Wdyt?

Conchylicultor:
Oh, I see, so the final size has actually 3 unknown dimensions `(None, None, None)` for `(seq_len, img_h, img_w)`. Yes, this case won't be supported unfortunately. It should be fine to keep nested `Sequence` in the meantime. Thanks for investigating.

ageron:
Got it! Thanks for your help on this dataset, @Conchylicultor . 👍

PR #1209, opened 2019-11-20
BugFix: Wrong labels in "DeepWeeds" dataset

hannesrichter:
Labels are now parsed from the author-supplied file.


hannesrichter:
@googlebot I signed it!

tensorflow_datasets/image/deep_weeds.py
```diff
@@ -98,19 +101,38 @@ def _split_generators(self, dl_manager):
             url=_URL,
             extract_method=tfds.download.ExtractMethod.ZIP))
 
+
+    path_lbl = dl_manager.download_and_extract(
+        tfds.download.Resource(
```
Conchylicultor: Please update the `fake_data/` dir to add this new file.

https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/testing/test_data/fake_examples/deep_weeds

Make sure to re-run the `deep_weed_test.py` to make sure the tests are passing.
You may have to move the fake images to a separate `images/` folder and add a `DL_EXTRACT_RESULT` to the `DeepWeedsTest `:

```python
DL_EXTRACT_RESULT = {
      "image": "images/",
      "label": "labels.csv",
}
```

tensorflow_datasets/image/deep_weeds.py
```diff
@@ -98,19 +101,38 @@ def _split_generators(self, dl_manager):
...
+  def _generate_examples(self, data_dir_path, label_dir_path):
     """Generate images and labels for splits."""
-
+    # parse the csv-label data
+    csv = np.loadtxt(label_dir_path,
+            dtype={'names': ('Filename', 'Label', 'Species'), 'formats': ('S21', 'i4', 'S1')},
+            skiprows=1,
+            delimiter=',')
+    
+    label_dict = {}
```
Conchylicultor: The labels that you are adding won't be correct, as the `self.info.features['label'].names` order do not match the ones from the file (another bug from the person which added the dataset).

Instead, you can dynamically assign the labels from the csv file:

In `_info`, only define the number of classes
```python
tfds.features.ClassLabel(num_classes=xx)
```

In `_generate_examples`, update the features labels, after having loaded the csv file:
```python
with tf.io.gfile.GFile(label_path) as f:
  reader = csv.DictReader(f)

  # Extract the mapping int -> str and save the label name string to the feature
  label_id_to_name = {
      row['Label']: row['Species'] for row in reader
  }
  self.info.features['label'].names = [v for k, v in sorted(label_id_to_name.items())]
```

Then you can yield the label string directly, without converting them to int first
```python
filename_to_label = {
    rows['Filename']: row['Species'] for row in reader
}
for file_name in tf.io.gfile.listdir(data_dir_path):
  yield file_name, {
      "image": os.path.join(data_dir_path, file_name), 
      "label": filename_to_label[file_name]
  }
```

Conchylicultor:
Thanks for your contribution. Looks good. Merging this now.

PR #1224, opened 2019-11-24
NIH Deeplesion, auto download version

jason-zl190:
correct user.name and user.email in the commit history


tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,264 @@
...
+            'zipfile49':'https://nihcc.box.com/shared/static/kisfbpualo24dhby243nuyfr8bszkqg1.zip',
+            'zipfile50':'https://nihcc.box.com/shared/static/rs1s5ouk4l3icu1n6vyf63r2uhmnv6wz.zip',
+
+            'zipfile51':'https://nihcc.box.com/shared/static/7tvrneuqt4eq4q1d7lj0fnafn15hu9oj.zip',
+            'zipfile52':'https://nihcc.box.com/shared/static/gjo530t0dgeci3hizcfdvubr2n3mzmtu.zip',
+            'zipfile53':'https://nihcc.box.com/shared/static/7x4pvrdu0lhazj83sdee7nr0zj0s1t0v.zip',
+            'zipfile54':'https://nihcc.box.com/shared/static/z7s2zzdtxe696rlo16cqf5pxahpl8dup.zip',
+            'zipfile55':'https://nihcc.box.com/shared/static/shr998yp51gf2y5jj7jqxz2ht8lcbril.zip',
+            'zipfile56':'https://nihcc.box.com/shared/static/kqg4peb9j53ljhrxe3l3zrj4ac6xogif.zip',
+            'ann_file':'https://raw.githubusercontent.com/anir16293/Deep-Lesion/master/DL_info.csv'
```
Ouwen: Why pull this from github? After extraction the zipfiles should create the annotation file

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,264 @@
...
+
+  VERSION = tfds.core.Version('0.1.0')
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            "image/name":tfds.features.Text(),
+            "image":tfds.features.Image(shape=(None, None, 1), dtype=tf.uint16, encoding_format='png'),
```
Ouwen: This feature description only contains the annotated slices, not the surrounding slices (which are quite useful to know the 3d context).

Consider making a sequence of images corresponding to the bounding box annotation.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: Are you saying the CSV file? they provide it on the Box, without static link and not compressed with the images.

jason-zl190: > Why pull this from github? After extraction the zipfiles should create the annotation file

Are you saying the CSV file? they provide it on the Box, without static link and not compressed with the images.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > This feature description only contains the annotated slices, not the surrounding slices (which are quite useful to know the 3d context).
> 
> Consider making a sequence of images corresponding to the bounding box annotation.

That's a good idea. I will provide this feature next.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: This CSV file is provided on [Box](https://nihcc.app.box.com/v/DeepLesion). However, the author does not provide a static link so I cannot download it directly from the original path.

cyfra:
The tests are failing with:
E       tensorflow.python.framework.errors_impl.NotFoundError: Could not find directory /tmpfs/src/github/datasets/tensorflow_datasets/testing/test_data/fake_examples/deeplesion/zipfile03/Images_png

jason-zl190:
> The tests are failing with:
> E tensorflow.python.framework.errors_impl.NotFoundError: Could not find directory /tmpfs/src/github/datasets/tensorflow_datasets/testing/test_data/fake_examples/deeplesion/zipfile03/Images_png

fixed

tensorflow_datasets/image/__init__.py
```diff
@@ -87,3 +87,4 @@
 from tensorflow_datasets.image.the300w_lp import The300wLp
 from tensorflow_datasets.image.uc_merced import UcMerced
 from tensorflow_datasets.image.visual_domain_decathlon import VisualDomainDecathlon
+
```
cyfra: remove empty line

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+
+            'zipfile51':'https://nihcc.box.com/shared/static/7tvrneuqt4eq4q1d7lj0fnafn15hu9oj.zip',
+            'zipfile52':'https://nihcc.box.com/shared/static/gjo530t0dgeci3hizcfdvubr2n3mzmtu.zip',
+            'zipfile53':'https://nihcc.box.com/shared/static/7x4pvrdu0lhazj83sdee7nr0zj0s1t0v.zip',
+            'zipfile54':'https://nihcc.box.com/shared/static/z7s2zzdtxe696rlo16cqf5pxahpl8dup.zip',
+            'zipfile55':'https://nihcc.box.com/shared/static/shr998yp51gf2y5jj7jqxz2ht8lcbril.zip',
+            'zipfile56':'https://nihcc.box.com/shared/static/kqg4peb9j53ljhrxe3l3zrj4ac6xogif.zip',
+            'ann_file':'https://raw.githubusercontent.com/anir16293/Deep-Lesion/master/DL_info.csv'
+    }
+    dl_manager._register_checksums = True
```
cyfra: This line should not be needed. (as you checksums should be registered only once)

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+        ymin = coords[i*4+1] / size[1]
+        xmin = coords[i*4] / size[0]
+        ymax = coords[i*4+3] / size[1]
+        xmax = coords[i*4+2] / size[0]
+        bbox_list.append(tfds.features.BBox(ymin=ymin, xmin=xmin, ymax=ymax, xmax=xmax))
+    
+    return bbox_list[0]
+    
+    
+class Split:
```
cyfra: any reason why this is a class rather than just a single function?

also - calling it 'Split' is too generic.

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+        self.ann_path = ann_path
+    
+    def split_generator(self):
+        """wrapper of the split
+        """
+        return self._ann_parser()
+    
+    def _ann_parser(self):
+        """parsers the csv file and returns the splits
```
cyfra: please add more explanation on what is inside the CSV file and what you return.

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+    paths = dl_manager.download_and_extract(resources)
+    ann_path = paths['ann_file']
+    del paths['ann_file']
+    lut = _build_lut(paths)
+    train_split, val_split, test_split = Split(ann_path).split_generator()
+
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=100,
```
cyfra: num_shards is no longer needed (here and below)

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+                "split": test_split,
+            },
+        ),
+    ]
+
+
+  def _generate_examples(self, lut, split):
+    """Returns examples
+    :type lut: dict, lookup dictionary to lookup whole paths of series_folder
+    :type split: string, path to read ann files
```
cyfra: split seems to be a list of tuples. Please update the comment

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+    :rtype: yield idx, example
+    """
+    for idx, value in enumerate(split.values):
+      fn, bboxs, size, sp = value
```
cyfra: nit: fn usually means 'function' - here I guess you mean the 'file_name' right? please update the variable name

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+    return os.path.join(lut[fd], slice_fn)
+    
+    
+def _format_bboxs(bboxs, size):
+    """Return bbox feature
+    :type bboxs: string, "xxx,xxx,xxx,xxx"
+    :type size: string, "xxx,xxx"
+    :rtype: tfds.features.BBox
+    """
+    size = list(map(lambda x: float(x), size.split(',')))
```
cyfra: nit: 

you can do (which is shorter):
[float(x) for x in size.split(',')]

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+    coords = list(map(lambda x: float(x), bboxs.split(',')))
+    coords = list(map(lambda x: x if x>0 else 0.0, coords))
+    coords = list(map(lambda x: x if x<size[0] else size[0], coords))
```
cyfra: If I understood this code correctly, you can replace it with:

coords = np.clip([float(x) for x in bbox.split(',')], 0.0, size[0])

which brings the question whether this is correct approach?
is size[0] == size[1]? if not - shouldn't you clip each dimension separately?

The alternative is to clip the xmin, ymin, xmax, ymax to 0.0, 1.0 after the division.

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,272 @@
...
+
+    bbox_list = []
+    for i in range(cnt):
+        ymin = coords[i*4+1] / size[1]
+        xmin = coords[i*4] / size[0]
+        ymax = coords[i*4+3] / size[1]
+        xmax = coords[i*4+2] / size[0]
+        bbox_list.append(tfds.features.BBox(ymin=ymin, xmin=xmin, ymax=ymax, xmax=xmax))
+    
+    return bbox_list[0]
```
cyfra: that's surprising that you return only a single bounding box. Is it on purpose?
Does dataset have multiple ?

tensorflow_datasets/image/__init__.py
jason-zl190: empty line removed

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: I planed to add more functions here to give more detailed annotation information. However, I replace it with a function for now. It should be more clear now. Thanks for the review.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > please add more explanation on what is inside the CSV file and what you return.

added it.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > num_shards is no longer needed (here and below)

Thanks for the notice. these lines have been removed.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > split seems to be a list of tuples. Please update the comment

updated

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > nit: fn usually means 'function' - here I guess you mean the 'file_name' right? please update the variable name

updated. Thanks for the suggestion.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: Thanks for the advice. I replaced this line with your suggestion. Glad to know the shorter and clearer usage.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > If I understood this code correctly, you can replace it with:
> 
> coords = np.clip([float(x) for x in bbox.split(',')], 0.0, size[0])
> 
> which brings the question whether this is correct approach?
> is size[0] == size[1]? if not - shouldn't you clip each dimension separately?
> 
> The alternative is to clip the xmin, ymin, xmax, ymax to 0.0, 1.0 after the division.

Thanks for the advice. I updated it. the sizes of two dimensions are always the same for this dataset. so it should works here.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > that's surprising that you return only a single bounding box. Is it on purpose?
> Does dataset have multiple ?

Thanks for the notice. I double-checked and fixed it in the last commit. For most of the images, they have only one annotation. However, some of the images do have multiple annotated bboxs.

tensorflow_datasets/image/__init__.py
```diff
@@ -90,4 +90,4 @@
 from tensorflow_datasets.image.svhn import SvhnCropped
 from tensorflow_datasets.image.the300w_lp import The300wLp
 from tensorflow_datasets.image.uc_merced import UcMerced
-from tensorflow_datasets.image.visual_domain_decathlon import VisualDomainDecathlon
```
cyfra: nit: revert this file

tensorflow_datasets/object_detection/__init__.py
```diff
@@ -20,3 +20,5 @@
 from tensorflow_datasets.object_detection.open_images import OpenImagesV4
 from tensorflow_datasets.object_detection.voc import Voc
 from tensorflow_datasets.object_detection.wider_face import WiderFace
+from tensorflow_datasets.object_detection.deeplesion import Deeplesion
```
cyfra: please sort

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,257 @@
...
+      yield idx, record
+
+    
+def _build_lut(paths):
+    """Returns lookup dictionary to lookup whole paths of series_folder
+    :type paths: dict, {<zipfn>:<unzip_path>}
+    :rtype: dict, {<series_folder>:<unzip_path>/Images_png/<series_folder>}
+    """
+    lut = {}
+    for k, v in paths.items():
```
cyfra: nit: you never use 'k' - replace it with _

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,257 @@
...
+    
+    
+def _format_bboxs(bboxs, size):
+    """Return bbox feature
+    :type bboxs: string, "xxx,xxx,xxx,xxx"
+    :type size: string, "xxx,xxx"
+    :rtype: tfds.features.BBox
+    """
+    size = [float(x) for x in size.split(',')]
+    coords = np.clip([float(x) for x in bboxs.split(',')], 0.0, size[0])
```
cyfra: please comment that you assume that sizes are equal. Or preferably do:
if len(size) != 2 or size[0] != size[1]: 
  raise AssertionError(...)

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,257 @@
...
+        xmax = coords[i*4+2] / size[0]
+        bbox_list.append(tfds.features.BBox(ymin=ymin, xmin=xmin, ymax=ymax, xmax=xmax))
+    
+    return bbox_list
+    
+    
+def _ann_parser(ann_path):
+    """parsers the annotation file and returns the splits in dataframes
+    :type ann_path: string, path of the annotation file
+    :rtype: pandas.dataframes
```
cyfra: tuple with 3 dataframes: for train, validation and test.

tensorflow_datasets/image/__init__.py
jason-zl190: > nit: revert this file

fixed

tensorflow_datasets/object_detection/__init__.py
jason-zl190: > please sort

fixed

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > nit: you never use 'k' - replace it with _

fixed

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > please comment that you assume that sizes are equal. Or preferably do:
> if len(size) != 2 or size[0] != size[1]:
> raise AssertionError(...)

fixed

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > tuple with 3 dataframes: for train, validation and test.

fixed

jason-zl190:
@cyfra Hi cyfra, Thanks for your review. I learned a lot. I create a new pull request #1342 from my local master branch, which is the same as here #1224 but I rebased the code on the newest official repository. I hope it can simplify the merge. Let me know if it's not. I will close that one.

jason-zl190:
@cyfra Like @Ouwen suggested this dataset contains contextual data that might help for some research but it's huge. I'm working on it to provide the whole data in another configuration. Should I include all images in records or should I include just the paths as strings and let users read them when necessary? I'm wondering which one is better. I appreciated for any advice.

tensorflow_datasets/object_detection/deeplesion.py
jason-zl190: > This feature description only contains the annotated slices, not the surrounding slices (which are quite useful to know the 3d context).
> 
> Consider making a sequence of images corresponding to the bounding box annotation.

Hi Ouwen, I finished the feature to provide the 3d context. Should I commit it here (to replace this one), or should I provide the 3d context feature using multiply configs or versioning? I'm a little confused about the usage of multiple configs or versioning.

jason-zl190: Also, I create a new pull request #1342 hoping to clean the commit history and keep updated to the newest official master branch. Let me know should I close that one.

jason-zl190: You can see the code here 128f0cf7c9ae0a31dac5f0c73212546a40a7d390

jason-zl190:
Thanks for the review! @cyfra

cyfra:
Hey @jason-zl190  - unfortunately this dataset turns out to be quite large - so in such cases, we'd like to have it use 'iter_archive' approach, rather than 'download_and_extract'.

The iter_archive, makes it more efficient (as you 'read' data only once).

tensorflow_datasets/object_detection/deeplesion.py
```diff
@@ -0,0 +1,259 @@
...
+    lut = {}
+    for _, v in paths.items():
+        for fd in tf.io.gfile.listdir(os.path.join(v,'Images_png')):
+            lut[fd] = os.path.join(v,'Images_png', fd)
+    return lut
+        
+
+def _lookup_image_path(lut, file_name):
+    """Returns whole path of an image
+    :type lut: dict, lookup dictionary
```
cyfra: for the function comments please use the same pattern as in other files:

""" Some short description.

Longer description

Args: 
  foo: something something
  bar: something else

Returns:
  some result
"""

jason-zl190:
> Hey @jason-zl190 - unfortunately this dataset turns out to be quite large - so in such cases, we'd like to have it use 'iter_archive' approach, rather than 'download_and_extract'.
> 
> The iter_archive makes it more efficient (as you 'read' data only once).

Hi, @cyfra Thanks for the review and notice. My concern is that whole zip archives will be iterated three times if they are not allowed to be extracted out. This is because those zip archives aren't compressed separately on splits. I need to iterate all zip archives once for collecting training data, once for validation data, and once for test data. However, I can refactor the code using `iter_archives` If you still think that's the better solution. Also, any suggestions are welcome if there's a way to avoid the three times iteration of the whole zip archives.

Another concern is that I plan to provide contextual images in the next version. However, I got `ResourceExhaustedError` when I'm trying to including the whole series images into a record. The reason is that a series, `Images_png/<series_folder> /xxx.png`, might includes more than 200 images. Although most of the images have no lesions on them, as @Ouwen suggested, they are surrounding slices and might contain contextual information, which is critical for medical diagnose. I'm thinking of including only the image paths in the records and let users decide how many contextual images they need to read. However, That means the data will be read frequently. Let me know if the paths-only plan is not a good idea. I appreciate getting any advice on how to include those mightly large surrounding images.

Ouwen:
@jason-zl190 does NIH predefine a validation/test split?

jason-zl190:
> @jason-zl190 does NIH predefine a validation/test split?

They do. They provided a CSV file, named "DL_info.csv," in which each key slice belongs to a split. However, the zip files they provided were compressed and separated orderly. For each split, images are randomly scattered in those archives.

cyfra:
@jason-zl190 - first, thanks a lot for your contribution - this is quite a large dataset, so it is bound to be challenging ;-)

What I'd suggest is:
a) let's start with reading the data 3 times. The other approach (download and extract) still has to read data twice (and write once) - but it has to read from many small files - which is usually slower (especially on distributed storage systems). That's why we prefer the iterate approach for larger datasets.

b) for the contextual images part: I'd tackle it in the separate PR. As you've mentioned, there are couple challenges (including the size of the records). One thing that you could do now - is to make sure that you include the 'series' name somewhere in the record (currently you include only file name from what I see).

jason-zl190:
> @jason-zl190 - first, thanks a lot for your contribution - this is quite a large dataset, so it is bound to be challenging ;-)
> 
> What I'd suggest is:
> a) let's start with reading the data 3 times. The other approach (download and extract) still has to read data twice (and write once) - but it has to read from many small files - which is usually slower (especially on distributed storage systems). That's why we prefer the iterate approach for larger datasets.
> 
> b) for the contextual images part: I'd tackle it in the separate PR. As you've mentioned, there are couple challenges (including the size of the records). One thing that you could do now - is to make sure that you include the 'series' name somewhere in the record (currently you include only file name from what I see).

@cyfra Hi, Thanks for the suggestions. I changed the code to read data directly from zip files. However, I didn't use `iter_archive` because the zip files are not compressed on splits. Instead, I look up and read necessary files using `zipfile` module. Also, I changed the comments pattern. I kept the structure of the example because file_name contains the series information inclusively. Please let me know if there is anything else I need to change.

jason-zl190:
@cyfra Thanks for your suggestion and review for this PR. However, this PR is a little outdated because of the long time window. I create a new PR #1769 contains the most recent changes, including different builder configs and better data filter rules. Would you mind taking it a look? Thanks a lot.

cyfra:
@jason-zl190 will do.

PR #1252, opened 2019-12-03
Adding the Opinosis dataset

Alex-Fabbri:
Adding the [Opinosis dataset](https://www.aclweb.org/anthology/C10-1039.pdf).


tensorflow_datasets/summarization/opinosis.py
```diff
@@ -0,0 +1,84 @@
...
+
+  def _generate_examples(self, path=None):
+    """Yields examples."""
+    topics_path = os.path.join(path, "topics")
+    filenames = tf.io.gfile.listdir(topics_path)
+    for i, filename in enumerate(filenames):
+      file_path = os.path.join(topics_path, filename)
+      topic_name = filename.split(".txt")[0]
+      with tf.io.gfile.GFile(file_path, "rb") as src_f:
+        lines = str(src_f.readlines())
```
cyfra: do you try to get read of "end-of-line" symbols ?

Doesn't it cause the words to merge ?

tensorflow_datasets/summarization/opinosis.py
```diff
@@ -0,0 +1,84 @@
...
+        input_data = "".join(lines)
+      summaries_path = os.path.join(path, "summaries-gold", topic_name)
+      summary_lst = []
+      for summ_filename in tf.io.gfile.listdir(summaries_path):
+        file_path = os.path.join(summaries_path, summ_filename)
+        file_path = os.path.join(summaries_path, summ_filename)
```
cyfra: duplicate line

tensorflow_datasets/summarization/opinosis.py
```diff
@@ -0,0 +1,84 @@
...
+        with tf.io.gfile.GFile(file_path) as tgt_f:
```
cyfra: do these files have UTF-8 chars ?

tensorflow_datasets/summarization/opinosis.py
Alex-Fabbri: Yeah I was just trying to get everything and pass it to the user to decide, so I changed it to read()

tensorflow_datasets/summarization/opinosis.py
Alex-Fabbri: Thanks for pointing this out.

tensorflow_datasets/summarization/opinosis.py
Alex-Fabbri: When I tried to open the source side data without "rb" it gave a decoding error, so I just changed both to open with "rb."

tensorflow_datasets/summarization/opinosis.py
```diff
@@ -0,0 +1,81 @@
+"""Opinosis Opinion Dataset"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+
+import tensorflow as tf
```
cyfra: import tensorflow.compat.v2 as tf

tensorflow_datasets/summarization/opinosis.py
```diff
@@ -0,0 +1,81 @@
...
+      with tf.io.gfile.GFile(file_path, "rb") as src_f:
+        input_data = src_f.read()
+      summaries_path = os.path.join(path, "summaries-gold", topic_name)
+      summary_lst = []
+      for summ_filename in tf.io.gfile.listdir(summaries_path):
+        file_path = os.path.join(summaries_path, summ_filename)
+        with tf.io.gfile.GFile(file_path, "rb") as tgt_f:
+          data = tgt_f.read()
+          summary_lst.append(data)
+      summary_data = b"[SEP_SUM]".join(summary_lst)
```
Conchylicultor: Rather than concatenating all different summaries together, do you think it would make sense to use `_SUMMARIES: tfds.features.Sequence(tfds.features.Text())` instead ? The user could still merge the strings together with `tf.string` or `tf.Text`.

We can do this modification internally when merging the dataset. So you can just answer here what was the rational to merge all summaries togethers.

tensorflow_datasets/summarization/opinosis.py
```diff
@@ -0,0 +1,81 @@
...
+            name=tfds.Split.TEST,
+            gen_kwargs={"path": extract_path},
+        ),
+    ]
+
+  def _generate_examples(self, path=None):
+    """Yields examples."""
+    topics_path = os.path.join(path, "topics")
+    filenames = tf.io.gfile.listdir(topics_path)
+    for i, filename in enumerate(filenames):
```
Conchylicultor: `tf.io.gfile.listdir` is not deterministic, so we shouldn't use the order as key. Instead the filename could be used

Alex-Fabbri: I modified this to use the filename as the key.

tensorflow_datasets/summarization/opinosis.py
Alex-Fabbri: Yes, that makes sense, and I modified it. I had been thinking of merging them into a master summary but this is much neater.

PR #1261, opened 2019-12-04
Add Imagenette to TensorFlow datasets

cking100:
This PR adds the [imagenette](https://github.com/fastai/imagenette) dataset as part of the tensorflow datasets.


tensorflow_datasets/image/imagenette.py
```diff
@@ -0,0 +1,139 @@
...
+import itertools
+import os
+
+import numpy as np
+import tensorflow as tf
+
+import tensorflow_datasets.public_api as tfds
+
+# TODO(imagenette): BibTeX citation
+_CITATION = """
```
cking100: I still have to add citation as I couldn't find one on the website.

cyfra: You can contact the author of the dataset and ask if they have one.

tensorflow_datasets/image/imagenette.py
```diff
@@ -0,0 +1,139 @@
...
+    configs.append(
+        ImagenetteConfig(
+            name='%s' % (size),
+            size=size,
+            description=_DESCRIPTION.format(size=size)))
+  return configs
+
+
+class Imagenette(tfds.core.GeneratorBasedBuilder):
+  """Imagenette:A smaller subset of 10 easily classified classes from
```
cyfra: please run the lint on your code:

https://github.com/tensorflow/datasets/blob/38189d93462b6e4f8fab820d416d8959c11d85e4/docs/add_dataset.md#5-check-your-code-style

tensorflow_datasets/image/imagenette.py
```diff
@@ -0,0 +1,139 @@
...
+        supervised_keys=('image', 'label'),
+        homepage='https://github.com/fastai/imagenette',
+        citation=_CITATION,
+    )
+
+  def _split_generators(self, dl_manager):
+    """Returns SplitGenerators."""
+    size = self.builder_config.size
+
+    if size in ["320 px","160 px"]:
```
cyfra: rather than repeating these lines 3 times, you could compute the "file-prefix" instead.

(file_prefix == "imagenette-160" or "imagenette-320" or "imagenette")

and then use it in the lines below.

tensorflow_datasets/image/imagenette.py
```diff
@@ -0,0 +1,139 @@
...
+      path = dl_manager.download_and_extract('%s/imagenette.tgz' % (_URL_PREFIX))
+      train_path = os.path.join(path, "imagenette", "train")
+      val_path = os.path.join(path, "imagenette", "val")
+    else:
+      raise ValueError('Size not implemented!')
+
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=10,  # Ignored when using a version with S3 experiment.
```
cyfra: you should be able to remove this line (as S3 should be enabled).

tensorflow_datasets/image/imagenette.py
```diff
@@ -0,0 +1,139 @@
...
+            gen_kwargs={
+                "datapath": train_path,
+            },
+        ),
+        tfds.core.SplitGenerator(
+            name=tfds.Split.VALIDATION,
+            num_shards=1,  # Ignored when using a version with S3 experiment.
```
cyfra: same here.

tensorflow_datasets/image/imagenette_test.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import print_function
+
+from tensorflow_datasets import testing
+from tensorflow_datasets.image import imagenette
+
+
+class ImagenetteTest(testing.DatasetBuilderTestCase):
+  DATASET_CLASS = imagenette.Imagenette
+  SPLITS = {
+      "train": 4,  # Number of fake train example
```
cyfra: your test images have fake random data (good) - but this also means that they don't really compress well (and some take over 3MB).

You can use function from testing/fake_data_utils.py 
(I see that png one is doing the "smart" thing, but the jpeg one is still generating large random images - we should fix that).

cking100: I have used the function from testing/fake_data_utils.py to generate fake random data. The images are large cuz the size I choose was large something like 2304 x 1728. The only reason for choosing a large size because I have seen similar size images in the original dataset.

tensorflow_datasets/image/imagenette.py
cking100: Okay, I will do it.

tensorflow_datasets/image/imagenette.py
cking100: Good idea. Will do it.

cking100:
@cyfra I have made the requested changes. Please have a look. I have also contacted the author for the citation. Now, I'm just left with https://github.com/tensorflow/datasets/pull/1261#discussion_r354002048. How should I fix it any suggestions?

cking100:
I have also removed the big jpeg files from fake data.

tensorflow_datasets/image/imagenette.py
cyfra: Did it run ? (as normally it complains that the comment should fit in a single line (with a "dot" at the end).

tensorflow_datasets/image/imagenette.py
```diff
@@ -0,0 +1,131 @@
...
+      path = dl_manager.download_and_extract(url)
+      train_path = os.path.join(path, "imagenette", "train")
+      val_path = os.path.join(path, "imagenette", "val")
+    else:
+      raise ValueError('Size not implemented!')
+
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            num_shards=10,
```
cyfra: please remove num_shards

tensorflow_datasets/image/imagenette.py
```diff
@@ -0,0 +1,131 @@
...
+  """BuilderConfig for Imagenette."""
+
+  def __init__(self, **kwargs):
+    super(ImagenetteConfig, self).__init__(
+        version=tfds.core.Version('0.1.0'), **kwargs)
+
+
+def _make_builder_configs():
+  configs = []
+  for variant in _VARIANTS:
```
cyfra: BTW -  having the config name "imagenette-160.tgz" doesn't look nice 
(as people will have to say:

tfds.load("imagenette/imagenette-160.tar")

Could you have the config named 160px instead ?

cking100: So use 160px instead of imagenette-160.tgz as the `name` parameter in `ImagenetteConfig`?

tensorflow_datasets/image/imagenette.py
cking100: Sure.

tensorflow_datasets/image/imagenette.py
cking100: I haven't run it on the whole tensorflow_dataset/image directory but I have checked it with the files that I have added and it is given 10/10 on both the files that I have added.

cking100: image

cking100:
I have made the changes. Please have a look. Thanks

tensorflow_datasets/image/imagenette.py
rickwierenga: https://github.com/fastai/imagenette/issues/16

guillaumelorre28:
Hello everyone,
the Imagenette dataset have changed:
"The Imagenette and Imagewoof datasets have recently (Dec 6 2019) changed. They now have a 70/30 train/valid split." (from https://github.com/fastai/imagenette)
The imagenette tensorflow dataset has still the small validation set.
It is impossible to compare to the fastai leaderboard using this dataset.

Would it be possible to update it using the new version?

Conchylicultor:
@guillaumelorre28 created #1679 to track. Please don't hesitate to send a PR with the new version. Do you think we should keep the two versions or only the new one ?
We could keep the two versions by using `BuilderConfig`.

PR #1265, opened 2019-12-05
GCI Task adding FLIC dataset

boronhub:
'flic': 'E:/Dhruv/Projects_19-20/Google_Code_In/Tensorflow/flic_dataset/FLIC',
TypeError: 'str' object is not callable

i am not able to extract the flic dataset


boronhub:
@googlebot I signed it!

PR #1266, opened 2019-12-05
GCI Task adding FLIC dataset

boronhub:
'flic': 'E:/Dhruv/Projects_19-20/Google_Code_In/Tensorflow/flic_dataset/FLIC',

TypeError: 'str' object is not callable

i am not able to extract the flic dataset


boronhub:
@googlebot I signed it!

WilliamHYZhang:
Have you thought about getting rid of 'flic' and just passing the url to the download_and_extract function by itself, as there is only one path? Otherwise, if you wish to keep 'flic' then try using dl_path['flic'].

PR #1267, opened 2019-12-05
Use static python_version dependency configuration in setup.py

ExpectationMax:
Allows installation of tensorflow-datasets using tools such as poetry and pipenv.

This pull request is very analogous to https://github.com/tensorflow/tensorflow/pull/32758 in the tensorflow main codebase.


ExpectationMax:
@googlebot I signed it!

Conchylicultor:
Nice, thanks for your contribution

Issue #547, opened 2019-05-08
[data request] FLIC Dataset


captain-pool:
* Name of dataset: FLIC
* URL of dataset: https://bensapp.github.io/flic-dataset.html
* License of dataset: MIT
* Short description of dataset and use case(s): The FLIC-full dataset is the full set of frames we harvested from movies and sent to Mechanical Turk to have joints hand-annotated. While the annotations between 5 turkers were almost always very consistent, many of these frames proved difficult for training / testing our MODEC pose model: occluded, non-frontal, or just plain mislabeled. 

Folks who would also like to see this dataset in `tensorflow/datasets`, please thumbs-up so the developers can know which requests to prioritize.

And if you'd like to contribute the dataset (thank you!), see our [guide to adding a dataset](https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md).

sgdgp:
@Conchylicultor I would like to work on this. Can this be assigned to me?

sgdgp:
@Conchylicultor @rsepassi 
The zip of the dataset is hosted/shared in google drive ([link](https://drive.google.com/file/d/0B4K3PZp8xXDJN0Fpb0piVjQ3Y3M/view?usp=sharing)). The data starts downloading only after manually triggering Download option in drive.

How to give the path in dl_manager.download_and_extract ?

Conchylicultor:
@sgdgp `dl_manager.download_and_extract` should support Google Drive url. Is it not the case ?

captain-pool:
@sgdgp can you provide the error you are facing?

PR #1290, opened 2019-12-11
Add FLIC Dataset

WilliamHYZhang:
Hi,

I'm currently working on adding the FLIC Dataset (https://bensapp.github.io/flic-dataset.html).

Currently, the plan is to have the annotations as closely resemble the original examples provided in the examples.mat file as we can.

Here is a basic outline of what I did already:
- citations, description, urls, all loaded
- example.mat file is loaded with scipy.io (will change import to tfds.core.lazy_imports in a future commit) (note: example.mat file is analyzed as a multidimensional array, e.g. data["example"][index_of_example_image][feature], where feature corresponds with the values in the FeatureDict)
- established FeatureDict in _info with features corresponding to the values given by .dtype (attribute names, mostly either tdfs.features.Image(), tfds.features.Text(), or tfds.features.Tensor() with corresponding shapes, types such as tf.uint16 and tf.float64 have also been found and implemented)
- established train/test splits in _split_generators and passed parameters to _generate_examples, done in the form of two lists (test and train) containing indexes of examples in data (note: no validation split from what I observed, booleans for istest/istrain are never both true or false)
- have a skeletal _generate_example function that simply yield the index of the attributes in data for each example in split

Some issues (most important to least important)
- download_and_prepare is failing as attributes with type tfds.features.Sequence are unable be yielded in the _generate_example function, returns error that the shapes must have the same length (note: these attributes are "poselet_hit_idx" and "coords", both of which are np arrays that have variable length for each example, see lines 99 and 101.)
- the "idx" values in the "poselet_hit_idx" array are nowhere in the range [1, 29] as the name-key mapping might suggest, I have absolutely no clue whatsoever what these numbers are supposed to represent
- "torsobox" attribute is in form of of 4 float Tensor instead of BBox, don't know if should change, if yes, then also need to implement normalize_bbox function and include PIL library (also probably from tfds.core.lazy_imports) to determine image sizes

Thanks, look forward to continuing with this project and contributing to Tensorflow.


tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,106 @@
...
+    ]
+
+  def _generate_examples(self, extract_path, data, split):
+    """Yields examples."""
+    for index in split:
+      yield index, {
+        "image": os.path.join(extract_path, "FLIC", "images", data["examples"][index][3]),
+        "poselet_hit_idx": {"idx": list(data["examples"][index][0])},
+        "moviename": data["examples"][index][1],
+        "coords": {"coord": list(data["examples"][index][2])},
```
WilliamHYZhang: This line should be rewritten to have "coords" return a dict of "x" and "y" tf.float64's. How should i use the zip() function for this?

WilliamHYZhang:
update: pushed with the fix, see output .json file [here](https://gist.github.com/WilliamHYZhang/a91ba59e867acfc00b8d773dae257938)

tensorflow_datasets/image/flic.py
WilliamHYZhang: Disregard, fixed with push

captain-pool:
Oops, I actually forgot to click Start Review Button. You should have got this review much earlier.😅

WilliamHYZhang:
@captain-pool Haha, thanks for the feedback, already fixed the issue with the coords, working on the updated split.

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,105 @@
...
+                                       mat_dtype=True)
+
+    train_list = []
+    test_list = []
+
+    for i in range(len(data["examples"])):
+      if data["examples"][i][7]:
+        train_list.append(i)
+      elif data["examples"][i][8]:
+        test_list.append(i)
```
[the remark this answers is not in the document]
WilliamHYZhang: @captain-pool thanks for the suggestion, pushed update with fix.

captain-pool:
Issue reference: #547

WilliamHYZhang:
@captain-pool thank you for the reviews! Will work on implementing changes and then testing later today.

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,101 @@
...
+      tfds.core.SplitGenerator(
+        name=tfds.Split.TEST,
+        gen_kwargs={
+          "extract_path": extract_path,
+          "data": data,
+          "istrain": False,
+          "istest": True,
+        },
+      ),
+    ]
```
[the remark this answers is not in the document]
WilliamHYZhang: I originally had a split for validation, however I believe in the annotations that none of the examples have both is true false and is test false, one of them is always true. This leads me to think that there was only intended to be too splits for the data.

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,101 @@
...
+  """Frames Labeled In Cinema (FLIC)"""
+
+  VERSION = tfds.core.Version('1.0.0')
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            "image": tfds.features.Image(),
```
captain-pool: As far as I remember the shape of all the images are more or less same. So, you can put that shape here as:
`tfds.features.Image(shape=(height, width, 3))`

tensorflow_datasets/image/flic.py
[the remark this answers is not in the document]
captain-pool: > 
> 
> I originally had a split for validation, however I believe in the annotations that none of the examples have both is true false and is test false, one of them is always true. This leads me to think that there was only intended to be too splits for the data.

I see.

WilliamHYZhang:
@captain-pool pushed new update with updated main file, new testing and fake examples also added. All tests are passing. Thanks so much!

captain-pool:
@cyfra, @rsepassi please take a look

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,95 @@
...
+detector on every tenth frame of 30 movies. People detected with high confidence
+(roughly 20K candidates) were then sent to the crowdsourcing marketplace Amazon
+Mechanical Turk to obtain groundtruthlabeling. Each image was annotated by five
+Turkers for $0.01 each to label 10 upperbody joints. The median-of-five labeling
+was taken in each image to be robust to outlier annotation. Finally, images were
+rejected manually by us if the person was occluded or severely non-frontal. We
+set aside 20% (1016 images) of the data for testing. 
+"""
+
+_URL = "https://drive.google.com/uc?id=0B4K3PZp8xXDJN0Fpb0piVjQ3Y3M&export=download"
```
captain-pool: Hey, I just suddenly remembered there are two versions of this right? One full dataset and another partial? And this is the partial right?

Will it be possible for you to write a heavy configuration to provide both versions of the datasets?
For example: "flic/full" and "flic/small".

WilliamHYZhang: Of course! On it.

WilliamHYZhang:
@captain-pool All done! Implemented the builderconfig and modified the testing and fake data for the new structure. All tests passing.

WilliamHYZhang:
Is there anything I should do before we can merge this and my other approved PRs in this repo? Thanks!

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,132 @@
...
+  """BuilderConfig for FLIC."""
+
+  @tfds.core.disallow_positional_args
+
+  def __init__(self, data, **kwargs):
+    """Constructs a FlicConfig."""
+    if data not in _DATA_OPTIONS:
+      raise ValueError("data must be one of %s" % _DATA_OPTIONS)
+
+    name = kwargs.get("name", data)
```
cyfra: nit: I'd put name=data in line 63 - and remove lines 48&49.

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,132 @@
...
+    configs.append(FlicConfig(
+        version=tfds.core.Version("2.0.0"),
+        data=data))
+  return configs
+
+class Flic(tfds.core.GeneratorBasedBuilder):
+  """Frames Labeled In Cinema (FLIC)"""
+
+  BUILDER_CONFIGS = _make_builder_configs()
+  VERSION = tfds.core.Version("2.0.0")
```
cyfra: can be removed (as you have versions in your configs)

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,132 @@
...
+        }),
+        citation=_CITATION,
+    )
+
+  def _split_generators(self, dl_manager):
+    """Returns SplitGenerators."""
+    extract_path = dl_manager.download_and_extract(
+        self.builder_config.url)
+    
+    data = scipy.io.loadmat(os.path.join(extract_path,
```
cyfra: please use tfds.core.lazy_imports.scipy.io.loadmat and remove the direct import of scipy.io

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,132 @@
...
+                                         self.builder_config.dir, "examples.mat"),
```
cyfra: Please change the code to use tf.io.gfile to read the data (this makes the code work on non-local filesystems too):

    with tf.io.gfile.GFile(filepath, "rb") as f:
      data = tfds.core.lazy_imports.scipy.io.loadmat(f)

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,132 @@
...
+                            struct_as_record=True, squeeze_me=True, mat_dtype=True)
+
+    return [
+      tfds.core.SplitGenerator(
+        name=tfds.Split.TRAIN,
+        gen_kwargs={
+          "extract_path": extract_path,
+          "data": data,
+          "istrain": True,
```
cyfra: nit:
having two separate bools, that are never both 'true' or both 'false' is not a great design.
Consider passing an enum or simply the 'name' argument instead.

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,132 @@
...
+          "istrain": False,
+          "istest": True,
+        },
+      ),
+    ]
+
+  def _generate_examples(self, extract_path, data, istrain, istest):
+    """Yields examples."""
+    for u_id, example in enumerate(data["examples"]):
+      if (example[7] and istrain) or (example[8] and istest):
```
cyfra: avoid 'magic numbers' - what is 7? what is 8?

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,132 @@
...
+Mechanical Turk to obtain groundtruthlabeling. Each image was annotated by five
+Turkers for $0.01 each to label 10 upperbody joints. The median-of-five labeling
+was taken in each image to be robust to outlier annotation. Finally, images were
+rejected manually by us if the person was occluded or severely non-frontal. We
+set aside 20% (1016 images) of the data for testing. 
+"""
+
+_DATA_OPTIONS = ["small", "full"]
+
+_URL_SUBSET = "https://drive.google.com/uc?id=0B4K3PZp8xXDJN0Fpb0piVjQ3Y3M&export=download"
```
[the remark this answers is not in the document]
WilliamHYZhang: They are the official links the dataset has provided.

WilliamHYZhang:
>nit:
having two separate bools, that are never both 'true' or both 'false' is not a great design.
Consider passing an enum or simply the 'name' argument instead.

>avoid 'magic numbers' - what is 7? what is 8?

I agree with this, however the dataset itself is structured like this, where for each example the 7th and 8th index respectively are booleans, `istrain` and `istest`. I was trying to reflect this in the splits.

WilliamHYZhang:
Thank you @cyfra for the time taken to review and your comments, appreciated very much.

cyfra:
> I agree with this, however the dataset itself is structured like this, where for each example the 7th and 8th index respectively are booleans, `istrain` and `istest`. I was trying to reflect this in the splits.


What you could do: is pass the 'int' argument (for example 'selection_column').
And then in the code you do:

if example[selection_column]:
  ....

WilliamHYZhang:
@ updated to reflect your suggestion.

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+  def _split_generators(self, dl_manager):
+    """Returns SplitGenerators."""
+    extract_path = dl_manager.download_and_extract(
+        self.builder_config.url)
+    
+    mat_path = os.path.join(extract_path, self.builder_config.dir, "examples.mat")
+    with tf.io.gfile.GFile(mat_path) as f:
+    data = tfds.core.lazy_imports.scipy.io.loadmat(f, struct_as_record=True,
```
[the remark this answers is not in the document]
WilliamHYZhang: Sorry, didn't catch that, should pass now.

tensorflow_datasets/image/flic.py
```diff
@@ -0,0 +1,134 @@
...
+          "data": data,
+          "selection_column": 7,
+        },
+      ),
+      tfds.core.SplitGenerator(
+        name=tfds.Split.TEST,
+        gen_kwargs={
+          "extract_path": extract_path,
+          "data": data,
+          "selection_column": 8,
```
Conchylicultor: Could you add a comment about what this selection_column represent ?

WilliamHYZhang:
Sure, no problem.

Conchylicultor:
Thanks, almost good to go. 

I think for consistency with other datasets, torsobox should be a `BBoxFeature()` instead of a `Tensor(shape=(4,))`, otherwise it's not clear for the user what are the coordinate order `(x, y, xmin, ymin)` ? `(x, xmin, y, ymin)`. What do you think ?

WilliamHYZhang:
Oh yeah, good point. Thanks for the suggestion.

WilliamHYZhang:
@Conchylicultor Implemented BBox attribute, fixed some bugs, uploaded checksums.

PR #1320, opened 2019-12-30
update dataset_info.json instructions

WilliamHYZhang:
Previously, the instructions for adding a dataset instructed users to create a Github Gist for the ```dataset_info.json``` files. However  I don't see the use of this as they should be added to the ```tensorflow_datasets/testing/metadata``` subdirectory for the dataset.  Updated the instructions to reflect this.

### DIFF
--- a/docs/add_dataset.md
+++ b/docs/add_dataset.md
@@ -506,7 +506,7 @@ python -m tensorflow_datasets.scripts.download_and_prepare \
 
 Note that the `--register_checksums` flag must only be used while in development.
 
-Copy in the contents of the `dataset_info.json` file(s) to the ```testing/metadata``` subdirectory for the dataset.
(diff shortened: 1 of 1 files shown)


anigasan:
Good that you had followed my recommendation. I was about to update!

Conchylicultor:
Thanks for you contribution, however both serve two distinct purpose.

* The gist is to ensure that contributors have actually tried to generate the full dataset (you'll be surprised by the number of people which send us code that they never tested)
* The `tensorflow_datasets/testing/metadata` is automatically created by us from the generated datasets at head. We want to keep it automatically generated and not rely on the user PR for multiple reason (ex: avoid incompatibilities if we update the dataset_info format but the PR was created in an old version).

Maybe we could improve the doc, but keeping it this way seems the best option for now.

WilliamHYZhang:
Thanks for the feedback! I was confused why some PR's had the .json in metadata and some didn't.

PR #1322, opened 2019-12-30
Updated EMNIST documentation

AND2797:
Replaced broken link - https://www.nist.gov/node/1298471/emnist-dataset by https://www.nist.gov/itl/products-and-services/emnist-dataset


AND2797:
@googlebot I signed it!

rsepassi:
Thank you. These docs are actually auto-generated, so the change should only go in mnist.py. I've gone ahead and done that and will send out shortly. Thank you for reporting!

PR #1323, opened 2019-12-31
[tensorflow_datasets/core/dataset_builder.py] Add no GCS support

SamuelMarks:
Was working on the train en route to the hospital, and noticed that this error:
> requests.exceptions.ConnectionError: HTTPConnectionPool(host='storage.googleapis.com', port=80): Max retries exceeded with url: /tfds-data/?prefix=dataset_info/DS%20NAME/1.0.0/ (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x1428de7d0>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))

Even though my dataset is found locally on my laptop and `try_download_gcs is False`.

This PR checks to see if the highest parent directory of `data_dir` exists, and if it does assumes that it's not a GCS directory.

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -197,6 +197,11 @@ class DatasetBuilder(object):
     if tf.io.gfile.exists(self._data_dir):
       logging.info("Overwrite dataset info from restored data version.")
       self.info.read_from_directory(self._data_dir)
+    elif os.path.isdir((lambda drive_directory: os.path.join(drive_directory[0],
+                                                             drive_directory[1][:drive_directory[1].find(os.path.sep,
+                                                                                                         1)])
+                       )(os.path.splitdrive(self._data_dir))):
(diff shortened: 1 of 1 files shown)


Conchylicultor:
Hi, thank you for your contribution. However, this fix really look like a hack. I don't see how checking for directory existence indicates if it is a GCS directory or not. GCS paths can certainly be a directory.

Could you provide more context about your issue, the error and what you're trying to achieve exactly ?

SamuelMarks:
Yes it is a bit of a hack, not sure a good non hacky way of handling this, but regardless it is a bug because it doesn't work offline.

I.e.: I can't use TensorFlow Datasets without an active Internet connection. It needs to query GCS, realise that it's not there, then grab it locally. This PR fixes that, but is—as you say—hacky.

Conchylicultor:
I'm closing this PR. If this is something you'll like to fix, could you open a bug with a small reproducible code example and the full stacktrace that you get when running the code.

PR #1331, opened 2020-01-04
Add 3.0.0 to fashion_mnist & kmnist

jvmancuso:
The documentation currently claims that both `fashion_mnist` and `kmnist` support the S3 API through version 3.0.0, since they inherit directly from the builder for MNIST. However, neither includes an S3 compatible version in their `SUPPORTED_VERSIONS`.  This PR adds the 3.0.0 version to both, fixing the discrepancy.

Happy to add a test or two, but felt that the existing tests were sufficient.  Works on Ubuntu with TF 2.0.0 & Python 3.5.

### DIFF
--- a/tensorflow_datasets/image/mnist.py
+++ b/tensorflow_datasets/image/mnist.py
@@ -174,6 +174,10 @@ class FashionMNIST(MNIST):
   VERSION = tfds.core.Version("1.0.0",
                               experiments={tfds.core.Experiment.S3: False})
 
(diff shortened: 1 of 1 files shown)


Conchylicultor:
As `FashionMNIST` inherit from `Mnist`, they are also inheriting from the `SUPPORTED_VERSIONS` property. Is it not working for you ?

jvmancuso:
ah, nvm different error, false flag!

PR #1342, opened 2020-01-06
Add NIH Deeplesion dataset, (same as #1224 but a updated and clean one)

jason-zl190:


jason-zl190:
this request contains the same change as #1224 but it updated with the official code and it's cleaner. I hope this one pull request could simplify the merge work.

Issue #910, opened 2019-08-14
Add a tfds.features.LabeledImage() feature connector


Conchylicultor:
For some use cases, it could be good to have a `LabelImage` or `LabeledImage` (name to define) which combines `ClassLabel` with `Image`. This could be used for per-instance pixel-masks.

API ideally should combine the two class kwargs and methods.
```python
features = {
    'mask': tfds.features.LabelImage(
        names=['background', 'people', 'road'], shape=(256, 256, 1))
}

info.features['mask'].num_classes  # 3
info.features['mask'].int2str(1)  # "people"
info.features['mask'].str2int('people')  # 1
info.features['mask'].encode_example('/path/to/img.png')

info.features['mask'].shape # (256, 256, 1)
info.features['mask'].dtype # tf.int64
```

* shape: (h,, w, 1)
* dtype: np.uint16

`Image(shape=(h,, w, 1), dtype=np.uint16)`

During `_generate_examples`:
```python
yield {
   'mask': '/path/to/image.png',  # Or np.array
}
```

Image could be stored as 1 channel int16 png, and converted to tf.int64 at decode time.
Implementation wise, not sure yet what would be the best option to reuse both Image and ClassLabel code.

rodrigob:
> This could be used for per-instance pixel-masks.
Note that it is important to distinguish between semantic labelling (e.g. Pascal VOC12), instance segmentation (e.g. COC 2017), and panoptic annotations case (e.g. Cityscapes).

One possible approach could be to have two channels int16 png.
One channel with the classes, another with the instance id (and just use by convention that instance id 0 is used for classes that are not supposed to have multiple instances).

Conchylicultor:
@rodrigob In case of instance segmentation, maybe it would make more sense to keep two separate image rather than trying to merge two semantic information in a single tensor (the user can still concatenate the two values).

```
features={
    'mask_label': tfds.features.LabelImage(names=['dog', 'cat']),
    'instance_id_mask': tfds.features.Image(dtype=tf.int16),
}
```

rodrigob:
Agreed. Then LabeImage focuses solely on the semantic labelling part of things, which is better.

rodrigob:
Toying with this feature, I noticed that I am unsure what is the recommended strategy to handle "empty features" for the test set.
I guess for class labels, the labels list is left empty. Unclear for `LabelImage` what is recommended way to express "no label ground truth available for this image".

Conchylicultor:
[...]

PR #1379, opened 2020-01-14
Added UCI Tic-Tac-Toe Endgame Data Set

boronhub:
Binary classification task on possible configurations of tic-tac-toe game

More about the dataset [here](https://archive.ics.uci.edu/ml/datasets/Tic-Tac-Toe+Endgame).

Link to dataset_info.json [here](https://gist.github.com/boronhub/eeb261f08908c2f789dd3ac346386641).

### DIFF
--- a/tensorflow_datasets/structured/tic_tac_toe.py
+++ b/tensorflow_datasets/structured/tic_tac_toe.py
@@ -1,0 +1,73 @@
+"""Binary classification task on possible configurations of tic-tac-toe game"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow_datasets.public_api as tfds
+
+_CITATION = """
+@misc{Dua:2019 ,
+author = "Dua, Dheeru and Graff, Casey",
+year = "2017",
+title = "{UCI} Machine Learning Repository",
+url = "http://archive.ics.uci.edu/ml",
+institution = "University of California, Irvine, School of Information and Computer Sciences" }
+"""
+
+_DESCRIPTION = """
+This database encodes the complete set of possible board configurations at the end of tic-tac-toe games, where "x" is assumed to have played first. The target concept is "win for x" (i.e., true when "x" has one of 8 possible ways to create a "three-in-a-row").
+
+Interestingly, this raw database gives a stripped-down decision tree algorithm (e.g., ID3) fits. However, the rule-based CN2 algorithm, the simple IB1 instance-based learning algorithm, and the CITRE feature-constructing decision tree algorithm perform well on it.
+"""
+
+def convert(d):
+  if d == "o":
+    return 0
+  if d == "x":
+    return 1
+  if d == "b":
+    return 2
+
+import csv
+import tensorflow as tf
+import tensorflow_datasets.public_api as tfds
+
+class TicTacToe(tfds.core.GeneratorBasedBuilder):
+  """Binary classification task on possible configurations of tic-tac-toe game"""
+
+  VERSION = tfds.core.Version('2.0.0')
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            "label": tfds.features.ClassLabel(names=["positive", "negative"]),
+            "features": tfds.features.Tensor(shape=(9,), dtype=tf.string)
+        }),
+        supervised_keys=("label", "features"),
+        homepage='https://archive.ics.uci.edu/ml/datasets/Tic-Tac-Toe+Endgame',
+        citation=_CITATION,
+    )
+
+  def _split_generators(self, dl_manager):
+    data = dl_manager.download("https://archive.ics.uci.edu/ml/machine-learning-databases/tic-tac-toe/tic-tac-toe.data")
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            gen_kwargs={
+
+                "file_path": data
+            },
+        ),
+    ]
+
+  def _generate_examples(self, file_path):
+    with tf.io.gfile.GFile(file_path) as f:
+      raw_data = csv.reader(f)
+      for i, row in enumerate(raw_data):
+        yield i, {
+            "label": str(row[-1]),
+            "features": [convert(x) for x in row[:-1]]
+        }
--- a/tensorflow_datasets/structured/__init__.py
+++ b/tensorflow_datasets/structured/__init__.py
@@ -19,4 +19,5 @@ from tensorflow_datasets.structured.amazon_us_reviews import AmazonUSReviews
 from tensorflow_datasets.structured.higgs import Higgs
 from tensorflow_datasets.structured.iris import Iris
 from tensorflow_datasets.structured.rock_you import RockYou
+from tensorflow_datasets.structured.tic_tac_toe import TicTacToe
 from tensorflow_datasets.structured.titanic import Titanic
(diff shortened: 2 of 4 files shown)


tensorflow_datasets/structured/tic_tac_toe.py
```diff
@@ -0,0 +1,73 @@
...
+def convert(d):
+  if d == "o":
+    return 0
+  if d == "x":
+    return 1
+  if d == "b":
+    return 2
+
+import csv
+import tensorflow as tf
```
WilliamHYZhang: Import should be `import tensorflow.compat.v2 as tf`

tensorflow_datasets/structured/tic_tac_toe.py
```diff
@@ -0,0 +1,73 @@
...
+
+  VERSION = tfds.core.Version('2.0.0')
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            "label": tfds.features.ClassLabel(names=["positive", "negative"]),
+            "features": tfds.features.Tensor(shape=(9,), dtype=tf.string)
```
WilliamHYZhang: Each feature should be a `ClassLabel(names=["x", "o", "b"])` I think.

boronhub: No, as they are strings in the dataset, and I have converted them into integers using the `convert() ` function, so they have to be a Tensor which I can assign a `dtype ` to.

WilliamHYZhang: Would it not be better to preserve the data as it is in the dataset, rather than converting it to integers?

boronhub: Not always, as sometimes adding this functionality makes the dataset much more intuitive. You can look at [titanic.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/structured/titanic.py) for an example of this.

WilliamHYZhang: Yes, observe that the classes are all labeled or converted to labels. Only the attributes that are originally quantitative values are of type `int` or `float`.

boronhub: Any use case would likely involve conversion to integers, hence I added this feature.

WilliamHYZhang: Hm, I understand your point of view, however I do think it's better to leave the labels the way they are for better understanding.

WilliamHYZhang:
I previously had another comment informing about the fact that you were passing a `dict` as `DL_EXTRACT_RESULT`, but removed it after looking at the documentation [here](https://github.com/tensorflow/datasets/blob/master/docs/api_docs/python/tfds/testing/DatasetBuilderTestCase.md), where it indeed does say that it's supposed to be a `dict` of `str`. However I that this is referring to the paths of the different files, (ie "name": "path") and not just the files themselves :joy:

tensorflow_datasets/structured/tic_tac_toe.py
boronhub: Alright,but I would like to wait for a mentor/reviewer's opinion. I do think what you're saying makes sense as well in that we stick to presenting the data as is, but I want to know if one extra function could potentially make importing and using this dataset easier.

WilliamHYZhang: Sounds good :+1:

boronhub:
> I previously had another comment informing about the fact that you were passing a `dict` as `DL_EXTRACT_RESULT`, but removed it after looking at the documentation [here](https://github.com/tensorflow/datasets/blob/master/docs/api_docs/python/tfds/testing/DatasetBuilderTestCase.md), where it indeed does say that it's supposed to be a `dict` of `str`. However I that this is referring to the paths of the different files, (ie "name": "path") and not just the files themselves 😂

In retrospect, I don't think `DL_EXTRACT_RESULT` is even needed, as there is only one file.

WilliamHYZhang:
Yeah I didn't think so either but apparently some other guy got an error without it...

tensorflow_datasets/structured/tic_tac_toe.py
Conchylicultor: This feature correspond to the disposition of labels (x, o, b) on a 3x3 grid.

Ideally, the feature correspond the most would be the LabeledImage proposal from this feature request (https://github.com/tensorflow/datasets/issues/910):
`LabeledImage(shape=(3, 3, 1), names=["b", "x", "o"])`

Unfortunately this is not available yet. The next best choices would be to choose between the `Label` or `Image` feature:
1. `Sequence(ClassLabel(names=["b", "x", "o"]), length=3*3)`
2. `Image(shape=(3, 3, 1))`

I'm in favour of 1., as it is can be converted trivially into image by the user with a `tf.reshape(labels, (3, 3, 1))`. With 2, the labels are lost for the users, which makes the example more difficult to interpret the image.

The user should be able to understand what each number represent. So I think it is important to keep the labels. The conversion str -> int is done automatically by TFDS, but the user can still recover the conversion int -> str with `info.features['labels'].int2str('x')`

The order of the labels is also important. The first label will have the value `0`, which in Python is associated with `False`:
1. tfds.features.ClassLabel(names=["positive", "negative"])  # positive=0, negative=1
2. tfds.features.ClassLabel(names=["negative", "positive"])  # positive=1, negative=2

2. is a much better choice because it match the user expectation that: `bool(str2int('negative')) == False` and `bool(str2int('positive')) == True`, which is not the case here.

Similarly, for the grid, `x` and `o` correspond to play, while `b` means "background"/"default"/"empty". In this case it make sense for the first label to be `b`, as an empty grid would just be a grid of 0.

WilliamHYZhang: Thanks for the insight @Conchylicultor

boronhub: Thank you @Conchylicultor, I'll make sure to implement the changes ASAP.

tensorflow_datasets/structured/tic_tac_toe.py
```diff
@@ -42,8 +69,38 @@ def _info(self):
...
+        citation=_CITATION,
+    )
+
+  def _split_generators(self, dl_manager):
+    data = dl_manager.download("https://archive.ics.uci.edu/ml/machine-learning-databases/tic-tac-toe/tic-tac-toe.data")
+    return [
+        tfds.core.SplitGenerator(
+            name=tfds.Split.TRAIN,
+            gen_kwargs={
+
```
jedlimlx: Great Job on the PR. One minor change would be to remove this unnecessary blank line.

tensorflow_datasets/structured/tic_tac_toe.py
```diff
@@ -0,0 +1,62 @@
...
+  VERSION = tfds.core.Version('0.0.1')
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            "label": tfds.features.ClassLabel(names=["negative", "positive"]),
+            "features": tfds.features.Sequence(ClassLabel(names=["b", "x", "o"]), length=3*3)
```
ChanchalKumarMaji: use `tfds.features.ClassLabel` not `ClassLabel`.

tensorflow_datasets/structured/tic_tac_toe.py
```diff
@@ -0,0 +1,62 @@
...
+        ),
+    ]
+
+  def _generate_examples(self, file_path):
+    with tf.io.gfile.GFile(file_path) as f:
+      raw_data = csv.reader(f)
+      for i, row in enumerate(raw_data):
+        yield i, {
+            "label": str(row[-1]),
+            "features": [convert(x) for x in row[:-1]]
```
ChanchalKumarMaji: `NameError: name 'convert' is not defined`.

PR #1393, opened 2020-01-17
Adding MSCOCO Captioning Dataset

Qwerty71:
This is a pull request to add the COCO Captioning Dataset. Currently, there exists a COCO dataset under `object_detection`, however, it does not contain the captions - only the bounding boxes for object detection. I believe that adding this captioning dataset will be of great benefit to the TensorFlow community.

Currently, these files only load the validation data from 2014 (which is 6GB - this is due to Colab's disk space not being enough to download the training set - which is 13GB - at least for now). A change to add the training set as well can be easily implemented in the near future.

This code has passed all `pylint`, `download_and_prepare`, and testing checks.

The dataset_info.json file can be found [here](https://gist.github.com/Qwerty71/e03d590c8f1ac84608ab2b4288eef9df).


Qwerty71:
Thanks! Nice to meet another GCI participant! 

Adding a new dataset gives the users options of only loading in one or the other (bounding boxes for object detection OR the captions for the images), whereas putting it all in one loads all of them, regardless of which the user wishes to use. In addition, I asked the mentors, and they said to create a new dataset for now.

WilliamHYZhang:
Cool, sounds great! I'll add some suggestion comments in the review. Perhaps in the future we could merge this to one and the user could have the option of specifying if desires bounding boxes/annotations/captions.

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
+"""MSCOCO Captioning Dataset from 2014."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import json
+import tensorflow as tf
```
WilliamHYZhang: This import should be `import tensorflow.compat.v2 as tf`

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+Superpixel stuff segmentation
+330K images (>200K labeled)
+1.5 million object instances
+80 object categories
+91 stuff categories
+5 captions per image
+250,000 people with keypoints
+
+In this dataset, there are 414113 captions, but 82783 images. Each image-caption
+pair is represented once in the dataset - thus, each image appears 5 times.
```
WilliamHYZhang: Could you specify that this is specifically only the captions?

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+"""
+
+_HOMEPAGE = "http://cocodataset.org/"
+
+class CocoCaptions(tfds.core.GeneratorBasedBuilder):
+  """MSCOCO Image Captioning Dataset - 2014"""
+
+  VERSION = tfds.core.Version('1.0.0')
```
WilliamHYZhang: Please update version to 2.0.0

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            'image': tfds.features.Image(encoding_format='jpeg'),
+            'caption': tfds.features.Text()
```
WilliamHYZhang: I'd actually suggest instead of yielding multiple of the same image with different captions, yield one image with the 5 captions. Let me know if you need help implementing this.

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+    Args:
+      images_path: `str`, directory containing the images
+      annotations_path: `str`, directory containing the annotations
+    Yields:
+      key (image id) and datapoint of (image, caption)
+    """
+    caption_file = '/captions_val2014.json'
+    with tf.io.gfile.GFile(annotations_path + caption_file) as f:
+      data = json.load(f)
+    path_head = images_path + '/COCO_val2014_'
```
WilliamHYZhang: Please use `os.path.join` for resolving the path

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+    ann = data['annotations']
```
WilliamHYZhang: Please make this variable a bit more descriptive :)

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+  """MSCOCO Image Captioning Dataset - 2014"""
+
+  VERSION = tfds.core.Version('1.0.0')
+
+  def _info(self):
+    return tfds.core.DatasetInfo(
+        builder=self,
+        description=_DESCRIPTION,
+        features=tfds.features.FeaturesDict({
+            'image': tfds.features.Image(encoding_format='jpeg'),
```
WilliamHYZhang: Could you specify the shape?

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+      annotations_path: `str`, directory containing the annotations
+    Yields:
+      key (image id) and datapoint of (image, caption)
+    """
+    caption_file = '/captions_val2014.json'
+    with tf.io.gfile.GFile(annotations_path + caption_file) as f:
+      data = json.load(f)
+    path_head = images_path + '/COCO_val2014_'
+    ann = data['annotations']
+    img_names = [path_head + '%012d.jpg' % i['image_id'] for i in ann]
```
WilliamHYZhang: Could you add a comment to explain what's going on in this list comprehension? Same for the next lines.

tensorflow_datasets/image/coco_captions.py
```diff
@@ -0,0 +1,111 @@
...
+    captions = ['<start> ' + i['caption'] + ' <end>' for i in ann]
+    ids = [i['id'] for i in ann]
+
+    for (i, name) in enumerate(img_names):
+      yield ids[i], {
+          'image': name,
+          'caption': captions[i]
```
WilliamHYZhang: Please see my previous comment about the caption yielding.

tensorflow_datasets/image/coco_captions.py
Qwerty71: I see where this is coming from, but I think there's benefit in yielding them separately. While training, implementing the 5 captions-per-image will require the user to iterate through each caption, just adding an extra step. On the contrary, yielding them separately will allow the user to implement one train step per datapoint in the dataset.

tensorflow_datasets/image/coco_captions.py
Qwerty71: I thought so as well, but I was getting an error about not using `os` from the `pylint` check.

Qwerty71:
Sure!

tensorflow_datasets/image/coco_captions.py
Qwerty71: Shouldn't the initial version be 1.0.0 (sorry if I'm misunderstanding, please let me know)?

tensorflow_datasets/image/coco_captions.py
WilliamHYZhang: I'm pretty sure you can ignore that error, looking at other datasets they all use `os.path.join` when needed, otherwise this will not work on all systems.

tensorflow_datasets/image/coco_captions.py
WilliamHYZhang: This is the tfds version, which should be 2.0.0 according to the mentors

tensorflow_datasets/image/coco_captions.py
WilliamHYZhang: Hm, interesting. I was thinking that yielding multiple images would result in unnecessary overhead and possibly some issues when shuffling/batching the data. I guess we can open this up to some further discussion down thr road.

Qwerty71: For sure. We'll see what others think - by the way, I'll try to make the other changes you mentioned later today.

tensorflow_datasets/image/coco_captions.py
ChanchalKumarMaji: No, `1.0.0` is Ok. Actually for new datasets we start with `0.0.1`.

ChanchalKumarMaji:
Add the `Training` set also. To increase the colab disk space, in the notebook do - `Edit` -> `Notebook settings` -> select `Hardware accelerator` to GPU -> `SAVE`.

tensorflow_datasets/image/coco_captions.py
ChanchalKumarMaji: It is recommended to use `tf.io.gfile`, instead of `os.path.join`.

ChanchalKumarMaji:
Make sure your dataset style is similar to [coco](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/object_detection/coco.py).

Qwerty71:
Hi!

I've tried adding the train dataset, but the Colab disk space just isn't compatible. I even switched to GPU (which only provides 69GB of storage, as compared to the 359GB it should) and TPU (which provides 107GB of storage - bug has been filed [here](https://github.com/googlecolab/colabtools/issues/919)) so I can't add it right now. Once this issue gets fixed, I will surely add it.

tensorflow_datasets/image/coco_captions.py
Qwerty71: How do we use `tf.io.gfile` for this? In the other datasets, they are always using `os.path.join`, and I don't see a similar function in the API docs for `tf.io.gfile`.

tensorflow_datasets/image/coco_captions.py
Qwerty71: Once we decide on which method to use, then I'll look into this 👍

tensorflow_datasets/image/coco_captions.py
Qwerty71: I'm not completely sure what you mean by this.

tensorflow_datasets/image/coco_captions.py
WilliamHYZhang: @captain-pool told me to change to 2.0.0 on my additions, so I'm a bit contradicted haha.

Qwerty71: That's fine. It does make much more sense for it to be `1.0.0` or `0.0.1` - I changed it to the latter but I initially thought the former made more sense, since it's a new addition of data.

> If MAJOR version is incremented, the existing data has been changed and/or the slicing API doesn't necessarily return the same set of records for a given slice.

WilliamHYZhang:
Hm, maybe try locally? It will take more time without GPU but if you have enough disk space it should work.

tensorflow_datasets/image/coco_captions.py
cyfra: As @WilliamHYZhang said: please use os.path.join for joining the paths:
path_head = os.path.join(images_path, "COCO_val2014_")

Same in line 99: instead of "annotations_path + caption_file" do:
os.path.join(annotations_path, caption_file).

we use the tf.io.gfile -- only for accessing the data (rather than constructing paths).

cyfra:
Hey @Qwerty71  - thanks for your efforts. Can you share more info about the mentor's decision to do it in separate file rather than extending existing coco code ?

The reason I'm asking: Coco already has multiple "Configs" (for example 2014, 2017, 2017_panoptic etc).
So my first approach, would have been to add "2014_with_captions" there.
This way we could keep things nicely together (both code & documentation wise).

sawyermade:
Hey guys, I figured out how to get the whole COCO-2017 dataset into Colab with Google Drive. Basically I broke train2017 and test2017 down into sub directories with a max of 5000 files (I noticed Colab could only read somewhere around 15k files from a directory, so 5000 seemed a safe bet). Here is the code for that:  

Then I used rclone to upload the whole damn dataset to Google Drive and shred with anyone who has a link can view: https://drive.google.com/drive/folders/1EVsLBRwT2njNWOrmBAhDHvvB8qrd9pXT?usp=sharing

Once you have the share in your google drive, create a shortcut for it so it can be accessed by Colab. Then I just create 118287 for train and 40670 for test symbolic links in the local directory. So far, it is working like a charm. I even save all my output to Google Drive so it can be resumed after the 12 hour kick. Here is the notebook for that: https://colab.research.google.com/drive/1OVStblo4Q3rz49Pe9-CJcUGkkCDLcMqP

I am training a mask rcnn now, will report results when finished but its looking pretty damn good so far.

Issue #1366, opened 2020-01-09
NonMatchingChecksumError for datasets "flores"


dojipkim:
**Short description**
Description of the bug.
tfds.load for "flores" fails due to NonMatchingChecksumError.

**Environment information**
* Operating System: Windows 10
* Python version: 3.7
* `tensorflow-datasets`/`tfds-nightly` version: tensorflow-datasets 1.2.0
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: tensorflow 2.0.0

**Reproduction instructions**

```
import tensorflow_datasets as tfds
datasets = tfds.load("flores", as_supervised=True)
```

**Link to logs**
```
NonMatchingChecksumError: Artifact https://github.com/facebookresearch/flores/raw/master/data/wikipedia_en_ne_si_test_sets.tgz, downloaded to C:\Users\dojip.kim\tensorflow_datasets\downloads\face_flor_raw_mast_wiki_en_ne_si_test_setsWS4PloHuIrs36CVNe1RvVy-NYIvZxGWik2ZmKbQjoUY.tgz.tmp.b8ce407eda6647d9a1129504da3c70c0\wikipedia_en_ne_si_test_sets.tgz, has wrong checksum.
```

**Expected behavior**
The datasets should be loaded without any issues

**Additional context**

Conchylicultor:
Do you still encounter this ? We do not host the data and sometimes the server hosting the data is down or reject the query if too many people try to download at the same time.
If you still get the issue, you can try to regenerate the checksums by running

This might indicates that the data has changed. In which case, we might want to regenerate the checksums with:

```
python -m tensorflow_datasets.scripts.download_and_prepare --registrer_checksums=True --datasets='flores'
```

dhirensr:
@Conchylicultor : even I am able to reproduce this on ubuntu!
```
import tensorflow_datasets as tfds
datasets = tfds.load("flores", as_supervised=True)
``` 
is not working.
however 
```
python -m tensorflow_datasets.scripts.download_and_prepare --registrer_checksums=True --datasets='flores'
``` 
with this it's working
can you describe how could we solve this? I would like to contribute to this issue

Conchylicultor:
Thanks for confirming. It seems the data has bee updated in https://github.com/facebookresearch/flores/commit/2f4a0e60bd7744304e87e456995fee5c61e87147, so the checksum has changed.

[...]

PR #1494, opened 2020-02-21
adding checksum for flores and also empty file check.

dhirensr:
related to https://github.com/tensorflow/datasets/issues/1366

@Conchylicultor : please review the PR.


tensorflow_datasets/translate/flores.py
```diff
@@ -174,6 +174,9 @@ def _generate_examples(self, source_file, target_file):
             len(source_sentences), len(target_sentences), source_file,
             target_file)
 
+    assert len(target_sentences) == 0, "Dataset file downloaded is empty"
+    assert len(source_sentences) == 0, "Dataset file downloaded is empty"
```
Conchylicultor: Have you try to generate the dataset with this change ?
If the dataset is generated without error, it means the dataset do not contains any examples, which is a bug.

dhirensr: @Conchylicultor : the dataset does get's generated but if I remove the assert. sorry for the issue.
the gists are https://gist.github.com/dhirensr/6de14800140e1780de3eaaa20a693c76, https://gist.github.com/dhirensr/46eca5ef6b389fee3e635738b1276200 
also the number of samples is the same so I should just update the patch right?

Conchylicultor: Are you sure you regenerated the dataset correctly ? The gist are both corresponding to v1.0.0, so it seems you haven't generated it with your changes

dhirensr: @Conchylicultor : it's the same ! I just generated the files without changing the version number,but when I did it afterwards also the data size is coming the same. so I am sure it's the same.
should I just remove the tests and update the version to 1.0.1?

Conchylicultor: I am not sure that you generated the dataset from the Github TFDS clone, because the dataset_info.json you copied do not contains some of the most recent additions (ex: dataset size).
Could you try to generate the data with your updates from the TFDS cloned version ?

dhirensr: @Conchylicultor : I updated the gist and I am using the TFDS Cloned version and on branch flores_checksum_update to generate the dataset_info.json.
As you can see in the gist only the version gets updated
also on the terminal print I can see one extra thing which is not there in dataset_info.json and that's the download_size. are you referring to that?

Conchylicultor: Thank you for the update. This looks good. I'll merge this soon.

Conchylicultor: Actually, I still do not understand how the generation can succeed if you haven't removed the
`assert len(target_sentences) == 0`.
Could you add a `print(len(target_sentences))` next to your assert and copy the full generation logs in a gist. This looks suspicious to me.

The only way generation succeed is if the number of examples is empty, which doesn't seems to be the case by looking at the gist.
Additionally your gist show `1.0.1`, but the version here is `1.1.0` ?

dhirensr: @Conchylicultor : I have pushed the code now in this branch! was waiting for you to review the gist files

Issue #1221, opened 2019-11-22
NonMachtingChecksumError for Kitti dataset


jrabary:
The following code fails with `NonMatchingChecksumError`

Code:

```
import tensorflow_datasets as tfds
import tensorflow as tf

kitti_builder = tfds.builder('kitti', data_dir='/data')

# Download the dataset
kitti_builder.download_and_prepare()
```

Output error:

```
NonMatchingChecksumError: Artifact https://s3.eu-central-1.amazonaws.com/avg-kitti/devkit_object.zip, downloaded to /data/downloads/s3.eu-central-1_avg-kitti_devkit_objectLwMAO1uIjU-eBJEp605N-Wtd3-25ncIu5j-rD605uW0.zip.tmp.523afd5a748243559a2ccd57ba87102b/devkit_object.zip, has wrong checksum
```

Conchylicultor:
Is the issue still here ? This might indicates that the file has been silently updated or is momentarily unavailable. If the issue persist, it means that we need to find a non-corrupted mirror of this file and update the url.
If it's the dataset which has been updated, the version should be increased with the new checksum (and should make sure the dataset still correctly generate).

jrabary:
The workaround I found is to run the script `download_and_prepare` with `register_checksums=True` as follow 

```
python -m tensorflow_datasets.scripts.download_and_prepare --registrer_checksums=True --datasets=kitti
```

cyfra:
This is a good workaround - but running with 'register_checksums=True' basically means that you 'ignore' the build-in checksum and simply download the file.

@jrabary - could you do us a favor and
a) verify that data is still correct and not corrupted (same number of elements etc)
b) send the pull request with the new checksum + version change?

Thanks!

tmensink:
I get the same error and used a similar workaround.

The resulting checksum is: 
`https://s3.eu-central-1.amazonaws.com/avg-kitti/devkit_object.zip 63794 ce0b76b69c0c5f89690a0d65b7302bbbdb962a0c7e8aba6efc7050d1b04b4cf1`

vijayphoenix:
Fixed by PR https://github.com/tensorflow/datasets/pull/1507

PR #1507, opened 2020-02-25
Update kitti dataset checksum files

vijayphoenix:
* Dataset Name: kitti dataset
* Issue Reference: https://github.com/tensorflow/datasets/issues/1221
* `dataset_info.json` Gist: [link](https://gist.github.com/vijayphoenix/cb778fc85a6f1fee3c1cf703ddd353c9)
  
  
## Checklist
* [X] Run `download_and_prepare` successfully
* [X] Add checksums file

CC: @Conchylicultor @cyfra, please review

### DIFF
--- a/tensorflow_datasets/object_detection/kitti.py
+++ b/tensorflow_datasets/object_detection/kitti.py
@@ -72,7 +72,7 @@ RawBoundingBox = collections.namedtuple("RawBoundingBox",
 class Kitti(tfds.core.GeneratorBasedBuilder):
   """Kitti dataset."""
 
-  VERSION = tfds.core.Version("3.1.0")
+  VERSION = tfds.core.Version("3.1.1")
   SUPPORTED_VERSIONS = [
       tfds.core.Version(
           "2.0.0", "New split API (https://tensorflow.org/datasets/splits)"),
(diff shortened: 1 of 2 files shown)


Conchylicultor:
Thank you very much for updating this. Could you just investigate a little why has the file changed ? (ex: has a new version of kitti been released ?).
In any case, we need to update the dataset Kitti version as this might have changed the underlying data.

It would be nice to link the change to this PR. No worries if you don't find anything.

vijayphoenix:
So, I found this on their site.

> Note 2: On 08.10.2019, we have followed the suggestions of the Mapillary team in their paper Disentangling Monocular 3D Object Detection and use 40 recall positions instead of the 11 recall positions proposed in the original Pascal VOC benchmark. 

This might have caused the changes.

vijayphoenix:
@Conchylicultor Updated the version to `3.1.1`

Conchylicultor:
Thanks for the quick update.
In the gist you copied, only the validation set has been generated (train and test are missing), this means that the new file has probably changed its structure.

The dataset and fake files also need to be updated to reflect those new changes.

vijayphoenix:
No, my computer ran out of space so I generated gist only for validation for now :-)

vijayphoenix:
The `dataset_info.json` now contains both validation and test set.
Unable to provide info for training set because of storage limitations on my side.

Conchylicultor:
Thank you for the update. I've merged this internally and it should appear in TFDS nightly in a few days

PR #1545, opened 2020-02-29
Added code to insert tfds dir into sys.path

### LINKED ISSUE #1544: ModuleNotFoundError: No module named 'tensorflow_datasets'
**Short description**
Interpreter is not able to import `tensorflow_datasets.core` in case of unstable versions of pythons. For example, code won't run in case of python 3.5 but it runs for 3.6. I have 3.7.6 in which it is not running. The problem is with `sys.path`. I think we need to insert tfds dir into `sys.path` . 

**Environment information**
* Operating System: Windows 10
* Python version: 3.7.6
* `tensorflow-datasets`/`tfds-nightly` version: 2.1
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: 2.0

**Error Stack**
```
  File "tensorflow_datasets/scripts/create_new_dataset.py", line 38, in <module>      
    from tensorflow_datasets.core import naming  # pylint: disable=g-import-not-at-top
ModuleNotFoundError: No module named 'tensorflow_datasets'
```

**Expected behavior**
The file should run if we execute following command from the tfds dir.
```
python tensorflow_datasets/scripts/create_new_dataset.py \
  --dataset my_dataset \
  --type image  # text, audio, translation,...
```

ashutosh1919:
Fixes #1544. @Conchylicultor , Please review this one.

### DIFF
--- a/tensorflow_datasets/scripts/create_new_dataset.py
+++ b/tensorflow_datasets/scripts/create_new_dataset.py
@@ -27,6 +27,9 @@ from __future__ import print_function
 
 import os
 
+import sys
+sys.path.insert(0,'.')
+
 from absl import app
 from absl import flags
 


dhirensr:
@ashutosh1919 : Path appending via sys is not needed actually and I am able to generate it without it.
maybe your $PYTHONPATH is not properly set.

ashutosh1919:
> @ashutosh1919 : Path appending via sys is not needed actually and I am able to generate it without it.
> maybe your $PYTHONPATH is not properly set.

That's what I mentioned. It depends upon the version of python you are using. I have checked my $PYTHONPATH. It's fine.

vijayphoenix:
I don't think this is necessary. Try to run this script installing the package in dev mode.
```
pip install -e .
```
See #1611 for more details.

ashutosh1919:
@vijayphoenix , Note here that if we are creating the new dataset files, it should not be necessary to install the code. Please read my initial comments on the parent issue #1544

vijayphoenix:
> @vijayphoenix, Note here that if we are creating the new dataset files, it should not be necessary to install the code

It is necessary to install the code because you are using `tfds core` API for creating the new dataset files.

> . Please read my initial comments on the parent issue #1544

`setup.py` is present for the very same reason.
https://github.com/tensorflow/datasets/blob/c034a193967d6d72152196708a5638e546e320f4/setup.py#L32
When you run 
```
python setup.py develop     or         pip install -e .
```
you add the current path to the list of all places interpreter needs to search for modules and packages.


> 2) If you install tensorflow_datasets using pip, then your new dataset files generated using

> python tensorflow_datasets/scripts/create_new_dataset.py --dataset my_dataset --type image

> will not be created in the current contribution dir of tensorflow_datasets, instead, they would be created in the site-packages/tensorflow_datasets dir which will be in your $PYTHONPATH because you are using tensorflow_datasets to import core from $PYTHONPATH and not the git cloned. 

This cmd
```
python setup.py develop     or         pip install -e .
```
will not add any new files to `site-packages/tensorflow_datasets` dir.
It will just install some third-party dependencies and it will add the current directory to your sys path

I hope this clarifies your doubt

ashutosh1919:
@vijayphoenix , Let's also take comments from @Conchylicultor

Issue #1528, opened 2020-02-27
[GSoC] Support more datasets format for visualization with tfds.show_examples()


Conchylicultor:
**Task**: Currently [tfds.show_examples()](https://www.tensorflow.org/datasets/overview#visualization) only works for supervised images datasets. It would be good to extend the heuristic to more dataset types, like:
* object detection
* 2-images side by side (ex: `div2k`, `cityscapes`)
* video (generated as gifs) ?
* texts ? 
* audio ?

**Instructions**:

* Update `show_examples` [visualization.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/visualization.py) to support new dataset types of your choice.
* Experiment on Colab to iterate with the visualization, link the Colab to the PR so reviewer can visualize the result
* Update the `visualization_test.py` with dummy data to test the visualization of the new dataset.
* Only add a single dataset type at a time

As multiple data-type exists, multiple people can work on this issue at the same time

dhirensr:
@Conchylicultor : working on translate type datasets.

vinayvr11:
Hello I would like to contribute on this issue. This is my first time so could you please guide me. So I only need to make a visualisation of these formats by implementing some updates in visualization_test.py.

vinayvr11:
@Conchylicultor : I am working on coco object detection dataset but this dataset was to big to load in my colab so can i use another dataset apart from tensorflow datasets because tensorflow object detection datasets are too big.

Conchylicultor:
@vinayvr11 Why don't you have a look at our catalog which display the download size ? For instance voc/2007 is about 1GB, https://www.tensorflow.org/datasets/catalog/voc

VaranRohila:
Hi, I was building code for support of object detection datasets but I got stuck at the values of the Bbox feature. They all are in between 0 and 1 and I checked in the source that it is meant to be that. I am just confused on what scale should I use while plotting the Bounding Boxes?

VaranRohila:
I figured it out. Here is the [notebook link](https://colab.research.google.com/drive/1801-vXCENw3w-ptNHh4eel2D1qHS9s9o#scrollTo=qFo0uHZ3CG11). @Conchylicultor Please review the changes.

Conchylicultor:
@VaranRohila nice, this looks great. Could you send a PR so I can see and review the code ?

VaranRohila:
I just did! [Link](https://github.com/tensorflow/datasets/pull/1639)

[...]

PR #1547, opened 2020-03-01
added support show_examples for translate data

dhirensr:
related to https://github.com/tensorflow/datasets/issues/1528

added support show_examples for translate type datasets

https://colab.research.google.com/drive/1LDXsE2tAxbn8qnhqpxUnXVpzltpSiYts

### DIFF
--- a/tensorflow_datasets/core/visualization.py
+++ b/tensorflow_datasets/core/visualization.py
@@ -27,11 +27,12 @@ from tensorflow_datasets.core import features as features_lib
 from tensorflow_datasets.core import lazy_imports_lib
 
 
-def show_examples(ds_info, ds, rows=3, cols=3, plot_scale=3., image_key=None):
-  """Visualize images (and labels) from an image classification dataset.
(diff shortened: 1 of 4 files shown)


Conchylicultor:
Thank you for your contribution, however, I'm not really sure what that users really want to have show_examples return a list of string.
Users can already get this using `list(tfds.as_numpy(ds.take(5)))`, so it's not clear to me what advantage this would bring vs the added complexity.
Unless we came out with a nice way of formatting the text I think we should focus on images types first (like object detection).

dhirensr:
@Conchylicultor : so for now I will send for the image datasets then like object detection or 2 image side by side

Conchylicultor:
Sounds good!

Conchylicultor:
Closing this as the user can already get the same feature with `list(tfds.as_numpy(ds.take(5)))`.
Note that I recently refactored the visualization to make it easier to add new extensions, in case you'll like to add new visualisation types.

PR #1580, opened 2020-03-05
Visualization support for images when as_supervised is set true

### LINKED ISSUE #1575: [GSoC] Add `as_supervised=True` compatibility for `tfds.show_examples`
It would be nice if `tfds.show_examples` was working with `as_supervised=True`. Currently it raise a not supported error:

```python
ds, ds_info = tfds.load('mnist', split='train', as_supervised=True)
tfds.show_examples(ds_info, ds)
```
Changes should only affect the `show_examples` function, so it should be fairly simple.

vijayphoenix:
Fix #1575

### DIFF
--- a/tensorflow_datasets/core/visualization/image_visualizer.py
+++ b/tensorflow_datasets/core/visualization/image_visualizer.py
@@ -78,6 +78,13 @@ def _add_image(ax, image):
   plt.xticks([], [])
   plt.yticks([], [])
 
+def _infer_ds_property(ds):  # TODO: batched
+  ex = list(dataset_utils.as_numpy(ds.take(1)))[0]
+  if isinstance(ex, tuple):
+    return "supervised"
+  if isinstance(ex, dict):
+    return "normal"
+  raise ValueError("No Visualization support for this type datasets")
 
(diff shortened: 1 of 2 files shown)


Conchylicultor:
Thank you very much. This would works however the implementation do not seem really scalable. For instance, every time someone want to add a new visualization capability, he/she would be forced to add `if is_supervised else` everywhere in his code.
Could you try to refactor this to make this more generic ?

vijayphoenix:
@Conchylicultor yeah, actually I was thinking of making a complete Visualization class which would work for all kinds of image datasets (for now, maybe extend this to other type datasets later).

Basically, this approach will address both  #1528 and #1575 at the same time.
But this will require a good amount of time, so this PR was just a temporary fix

Conchylicultor:
@vijayphoenix, yes, having a more flexible visualization looks good! Thank you for looking into this.

vijayphoenix:
@Conchylicultor Please review the changes.
Thank you

vijayphoenix:
@Conchylicultor Please review the changes.
Thank you

PR #1586, opened 2020-03-05
Test that new dataset has correctly set the checksums

### LINKED ISSUE #1397: Test that new dataset has correctly set the checksums
When adding new datasets in our repository, it is possible for developer to forgot the checksums.

We cannot test the actual checksums values in the unittest, but we should at least check the presence or not of the requested urls.

In `dataset_builder_testing.py`, should modify the mocking of the `dl_manager.download` and `dl_manager.download_and_prepare` to validate the checksum urls presence.
https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/testing/dataset_builder_testing.py#L218
Also add a constant `DatasetBuilderTestCase.SKIP_CHECKSUMS = False`, if external tests want to opt-out of checksum testing.

vijayphoenix:
Fix #1397

### DIFF
--- a/tensorflow_datasets/testing/dataset_builder_testing.py
+++ b/tensorflow_datasets/testing/dataset_builder_testing.py
@@ -277,14 +277,15 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
         self._test_checksums()
 
   def _test_checksums(self):
-    filepath = os.path.join(checksums._get_path(self.builder.name))
-    if tf.io.gfile.exists(filepath):
+    if self._download_urls:
+      filepath = os.path.join(checksums._get_path(self.builder.name))
+      if not tf.io.gfile.exists(filepath):
+        raise AssertionError("url checksums file not found at %s" % filepath)
+
       sizes_checksums = checksums._get_sizes_checksums(filepath)
       urls = sizes_checksums.keys()
-    else:
-      raise AssertionError("url checksums file not found at %s" % filepath)
-
-    self.assertTrue(self._download_urls.issubset(set(urls)), "url checksums don't match.")
+      
+      self.assertTrue(self._download_urls.issubset(set(urls)), "url checksums don't match.")
 
   def _download_and_prepare_as_dataset(self, builder):
     # Provide the manual dir only if builder has MANUAL_DOWNLOAD_INSTRUCTIONS
--- a/tensorflow_datasets/testing/dataset_builder_testing.py
+++ b/tensorflow_datasets/testing/dataset_builder_testing.py
@@ -29,6 +29,7 @@ from absl.testing import parameterized
 import numpy as np
 import tensorflow.compat.v2 as tf
 
+# from tensorflow_datasets.core.download import checksums
 from tensorflow_datasets.core import dataset_builder
 from tensorflow_datasets.core import dataset_info
 from tensorflow_datasets.core import dataset_utils
@@ -138,14 +139,13 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
   OVERLAPPING_SPLITS = []
   MOCK_OUT_FORBIDDEN_OS_FUNCTIONS = True
   SKIP_CHECKSUMS = False
-  URLS = []
-  CHECKSUMS_PATH = None 
   
   @classmethod
   def setUpClass(cls):
     tf.compat.v1.enable_eager_execution()
     super(DatasetBuilderTestCase, cls).setUpClass()
     name = cls.__name__
+    cls._download_urls = []
     # Check class has the right attributes
     if cls.DATASET_CLASS is None or not callable(cls.DATASET_CLASS):
       raise AssertionError(
@@ -219,7 +219,7 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
     self.assertEqual(self.builder.name, info.name)
 
   def _get_dl_extract_result(self, url):
-    self.URLS.append(url)
+    self._download_urls.append(url)
     del url
     if self.DL_EXTRACT_RESULT is None:
       return self.example_dir
@@ -233,12 +233,6 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
       return self._get_dl_extract_result(url)
     return utils.map_nested(lambda fname: os.path.join(self.example_dir, fname),
                             self.DL_DOWNLOAD_RESULT)
-
-  def _get_checksums_filepath(self):
-    if self.CHECKSUMS_PATH is None:
-      dir_path = os.path.normpath(os.path.join(os.path.dirname(__file__), "../url_checksums"))
-      return os.path.join(dir_path, self.builder.name + ".txt")
-    return self.CHECKSUMS_PATH
 
   def _make_builder(self, config=None):
     return self.DATASET_CLASS(  # pylint: disable=not-callable
@@ -270,19 +264,19 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
           self._download_and_prepare_as_dataset(builder)
     else:
       self._download_and_prepare_as_dataset(self.builder)
-
     if not self.SKIP_CHECKSUMS:
       with self._subTest("url_checksums"):
         self._test_checksums()
 
   def _test_checksums(self):
     urls = []
-    filepath = self._get_checksums_filepath()
-    with tf.io.gfile.GFile(filepath, "rb") as f:
+    filepath = os.path.join(download.checksums._CHECKSUM_DIRS[0], self.builder.name + ".txt")
+    if not tf.io.gfile.exists(filepath):
+      raise AssertionError("The urls checksums file does not exist")
+    with tf.io.gfile.GFile(filepath, "r") as f:
       for line in f.readlines():
-        urls.append(line.split()[0].decode("utf-8"))
-
-    self.assertEqual(set(self.URLS), set(urls))
+        urls.append(line.split()[0])
+    self.assertEqual(set(self._download_urls), set(urls), "The urls checksums don't match")
 
   def _download_and_prepare_as_dataset(self, builder):
     # Provide the manual dir only if builder has MANUAL_DOWNLOAD_INSTRUCTIONS
--- a/tensorflow_datasets/testing/dataset_builder_testing.py
+++ b/tensorflow_datasets/testing/dataset_builder_testing.py
@@ -139,7 +139,8 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
   MOCK_OUT_FORBIDDEN_OS_FUNCTIONS = True
   SKIP_CHECKSUMS = False
   URLS = []
-
+  CHECKSUMS_PATH = None 
+  
   @classmethod
   def setUpClass(cls):
     tf.compat.v1.enable_eager_execution()
@@ -233,6 +234,12 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
     return utils.map_nested(lambda fname: os.path.join(self.example_dir, fname),
                             self.DL_DOWNLOAD_RESULT)
 
+  def _get_checksums_filepath(self):
+    if self.CHECKSUMS_PATH is None:
+      dir_path = os.path.normpath(os.path.join(os.path.dirname(__file__), "../url_checksums"))
+      return os.path.join(dir_path, self.builder.name + ".txt")
+    return self.CHECKSUMS_PATH
+
   def _make_builder(self, config=None):
     return self.DATASET_CLASS(  # pylint: disable=not-callable
         data_dir=self.tmp_dir,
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +263,20 @@ def test_download_and_prepare_as_dataset(self):
     else:
       self._download_and_prepare_as_dataset(self.builder)
 
+    if not self.SKIP_CHECKSUMS:
+      with self._subTest("url_checksums"):
+        self._test_checksums()
+
+  def _test_checksums(self):
+    urls = []
+    path = ""  # How to get this path
```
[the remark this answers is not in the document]
Conchylicultor: Why don't you have a look at the checksums.py file to see how this is done ?

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -137,7 +137,10 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
   EXAMPLE_DIR = None
   OVERLAPPING_SPLITS = []
   MOCK_OUT_FORBIDDEN_OS_FUNCTIONS = True
-
+  SKIP_CHECKSUMS = False
+  URLS = []
+  CHECKSUMS_PATH = None
```
Conchylicultor: The checksum path should be automatically look-up, by checking all the registered checksum paths. This shouldn't be a property.

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -230,6 +234,12 @@ def _get_dl_download_result(self, url):
     return utils.map_nested(lambda fname: os.path.join(self.example_dir, fname),
                             self.DL_DOWNLOAD_RESULT)
 
+  def _get_checksums_filepath(self):
+    if self.CHECKSUMS_PATH is None:
```
Conchylicultor: You should use `checksums._CHECKSUM_DIRS` instead.

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -216,6 +219,7 @@ def test_info(self):
     self.assertEqual(self.builder.name, info.name)
 
   def _get_dl_extract_result(self, url):
+    self.URLS.append(url)
```
Conchylicultor: This might not be a string but a `tfds.download.Resource(`. Could you tried to run on `image/sun_test.py` to check this is working properly ?

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -137,7 +137,10 @@ class DatasetBuilderTestCase(parameterized.TestCase, test_utils.SubTestCase):
   EXAMPLE_DIR = None
   OVERLAPPING_SPLITS = []
   MOCK_OUT_FORBIDDEN_OS_FUNCTIONS = True
-
+  SKIP_CHECKSUMS = False
+  URLS = []
```
Conchylicultor: All class will share the same urls, which might create conflicts between multiple instances. This shouldn't be a global attribute.
I would move it `setUpClass` instead.
Also UPPER_CASE mean constant in Python, I would rename to `_download_urls`.

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +271,19 @@ def test_download_and_prepare_as_dataset(self):
...
       self._download_and_prepare_as_dataset(self.builder)
 
+    if not self.SKIP_CHECKSUMS:
+      with self._subTest("url_checksums"):
+        self._test_checksums()
+
+  def _test_checksums(self):
+    urls = []
+    filepath = self._get_checksums_filepath()
+    with tf.io.gfile.GFile(filepath, "rb") as f:
```
Conchylicultor: You should also check the file exists in the first place

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +271,19 @@ def test_download_and_prepare_as_dataset(self):
...
+      for line in f.readlines():
+        urls.append(line.split()[0].decode("utf-8"))
+
+    self.assertEqual(set(self.URLS), set(urls))
```
Conchylicultor: The user is likely to be confused about this error. Could you raise a more explicit error message ?

tensorflow_datasets/testing/dataset_builder_testing.py
Conchylicultor: Why `rb` ? and not just `r` ?

tensorflow_datasets/testing/dataset_builder_testing.py
vijayphoenix: Yeah, just noticed that url may not be a string. 
```
{'images': <tensorflow_datasets.core.download.resource.Resource object at 0x7fb04a11c790>, 'partitions': 'https://vision.princeton.edu/projects/2010/SUN/download/Partitions.zip'}
```
But when I run this file (without including any of my changes)
```
python tensorflow_datasets/image/sun_test.py
```
I am getting segmentation fault. So, I will trying to resolve it first

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +279,23 @@ def test_download_and_prepare_as_dataset(self):
...
       self._download_and_prepare_as_dataset(self.builder)
 
+    if not self.SKIP_CHECKSUMS:
+      with self._subTest("url_checksums"):
+        self._test_checksums()
+
+  def _test_checksums(self):
+    urls = []
+    # how to get the path to the checksum file corresponding to the testcase.
+    filepath = os.path.join(download.checksums._get_path(os.path.basename(self.example_dir)))
```
vijayphoenix: @Conchylicultor I am unable to get the file path to the correct checksum file.
The above code does not work for testcases like `translate/wmt_test.py`
and
`self.builder.name`, `self.info.name` does not work in for `sun_test.py`.

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -29,6 +29,7 @@
 import numpy as np
 import tensorflow.compat.v2 as tf
 
+# from tensorflow_datasets.core.download import checksums
```
vijayphoenix: will remove it...

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +279,23 @@ def test_download_and_prepare_as_dataset(self):
...
+    # how to get the path to the checksum file corresponding to the testcase.
+    filepath = os.path.join(download.checksums._get_path(os.path.basename(self.example_dir)))
+    if tf.io.gfile.exists(filepath):
+      with tf.io.gfile.GFile(filepath, "r") as f:
+        for line in f.readlines():
+          urls.append(line.split()[0])
+    else:
+      raise AssertionError("url checksums file not found at %s" % filepath)
+
+    self.assertTrue(self._download_urls.issubset(set(urls)), "url checksums don't match.")
```
vijayphoenix: All URLs are not encountered. So `_download_urls` is only a subset of all URLs in the url_checksum file.
Please correct me, if I am wrong.

vijayphoenix:
@Conchylicultor could please help me out with the issues.
Thank you

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -216,6 +219,21 @@ def test_info(self):
...
+        for _, v in url_or_urls.items():
+          add_url(v)
+      elif isinstance(url_or_urls, list):
+        map(add_url, url_or_urls)
+      elif isinstance(url_or_urls, download.resource.Resource):
+        self._download_urls.add(url_or_urls.url)
+      else:
+        self._download_urls.add(url_or_urls)
+
+    add_url(url)
```
Conchylicultor: You could use `tf.nest.map_structure` instead to recursively add the urls.

And you also need to modify `_get_dl_download_result `

tensorflow_datasets/testing/dataset_builder_testing.py
Conchylicultor: Yes, this is correct (ex:for builder config)

tensorflow_datasets/testing/dataset_builder_testing.py
Conchylicultor: Why don't you use `checksums._get_path(dataset_name)` ? It will already raise the error if it don't find the file

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +279,23 @@ def test_download_and_prepare_as_dataset(self):
...
+      with self._subTest("url_checksums"):
+        self._test_checksums()
+
+  def _test_checksums(self):
+    urls = []
+    # how to get the path to the checksum file corresponding to the testcase.
+    filepath = os.path.join(download.checksums._get_path(os.path.basename(self.example_dir)))
+    if tf.io.gfile.exists(filepath):
+      with tf.io.gfile.GFile(filepath, "r") as f:
+        for line in f.readlines():
```
Conchylicultor: You should reuse existing code whenever possible, here by loading the checksums with `checksums._get_sizes_checksums(path)`

tensorflow_datasets/testing/dataset_builder_testing.py
vijayphoenix: I agree we could this but for test files like sun_test.py, I get the following error.
```
File "/home/vijay/GitHub-Repos/Tensorflow/datasets/tensorflow_datasets/testing/dataset_builder_testing.py", line 289, in _test_checksums
filepath = os.path.join(download.checksums._get_path(self.dataset_name))
AttributeError: 'Sun397TfdsTest' object has no attribute 'dataset_name'

```

vijayphoenix: or if we use `checksums._get_path(self.builder.name)`. I got the following error
```
AssertionError: No checksums file could be found for dataset testable_sun397. Please create one in one of:
```
So, is there any proper way to handle all these errors and get the path to proper checksum file.

Conchylicultor: What is the content of `checksums._CHECKSUM_DIRS` ?
Note that when encountering a bug like this, you should try to look at the code which raise the error and try to debug from there.

Conchylicultor: Also it is expected that this fail if you haven't created a `testable_sun397.txt` file first. Why don't you try to test on an existing dataset test ?

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +272,20 @@ def test_download_and_prepare_as_dataset(self):
...
+
+  def _test_checksums(self):
+    filepath = os.path.join(checksums._get_path(self.builder.name))
+    if tf.io.gfile.exists(filepath):
+      sizes_checksums = checksums._get_sizes_checksums(filepath)
+      urls = sizes_checksums.keys()
+    else:
+      raise AssertionError("url checksums file not found at %s" % filepath)
+
+    self.assertTrue(self._download_urls.issubset(set(urls)), "url checksums don't match.")
```
Conchylicultor: I'm sure there is a `self.assertSubSet`, or a better function than `assertTrue `.

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +272,20 @@ def test_download_and_prepare_as_dataset(self):
     else:
       self._download_and_prepare_as_dataset(self.builder)
 
+    if not self.SKIP_CHECKSUMS:
+      with self._subTest("url_checksums"):
+        self._test_checksums()
+
+  def _test_checksums(self):
```
Conchylicultor:
```
filepath = os.path.join(checksums._get_path(self.builder.name))
if not tf.io.gfile.exists(filepath):
  raise AssertionError("url checksums file not found at %s" % filepath)

sizes_checksums = checksums._get_sizes_checksums(filepath)
urls = sizes_checksums.keys()
```

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -261,6 +272,20 @@ def test_download_and_prepare_as_dataset(self):
...
+    filepath = os.path.join(checksums._get_path(self.builder.name))
```
Conchylicultor: Note you should only check the file existence if the `dl_manager.download` has been called. If there is no check to do if `self._download_urls` is empty.

tensorflow_datasets/testing/dataset_builder_testing.py
vijayphoenix: I could only find [assertSetEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertSetEqual).
A similar subset function [assertDictContainsSubset](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertDictContainsSubset) is deprecated

Issue #1361, opened 2020-01-09
Better (a.k.a smalller) fake image generator


cyfra:
Currently our fake_image code (get_random_picture) in fake_data_utils, is creating a truly random image (by randomizing each pixel).

This unfortunately means, that images are non-compresable, which results in large images in test_data directory.

The goal is to:
a) change that method to generate random images that compress better (for example maybe by generating larger 'tiles' instead of single pixels etc)
b) replacing the largest images that we have in our 'test_data' with more 'compressable' equivalents.

Conchylicultor:
Yes I completely agree. I had plan to add a script to parse all images in the `fake_data/` directory and replace all images by single colour of the same shape (Something like: `np.ones_like(original_image) * np.random.randint(max(original_image))`).

It shouldn't be too difficult, but the tricky part is that some images are compressed inside `.zip`, `.tar`,... archives. So ideally the script should also extract archives, replace images within the archive (without modifying other files) and recompress the archive with the right format.

I hadn't the time to add the script yet. If someone can help with this it would be great.

Shashankjain12:
Yes i will be happy to work on this issue @Conchylicultor  and @cyfra so if you won't mind can you assign me the issue so that I can work on that.!

Shashankjain12:
hey @Conchylicultor  and @cyfra can we use np.compress() method to compress those images?

Shashankjain12:
I got a bit confused in this case, To solve this problem I was thinking to use ```np.ones_like(original_image) * np.random.randint(np.amax(original_image))```
and for the compressing part I got a bit confused do we need to replace all the previously generated fake image files by it's compressed equivalant generated from above or just replace with newly generated fake files?

Conchylicultor:
The goal of the issue is to reduce the size of the `.png` and `.jpg` images present in the folder. `.png` and `jpg` images are format which already compress the image.

The issue is that white noise images by definition cannot be compressed, as they do not have any patterns and contains the maximum information. They are too random to be compressed by the png and jpg compression algorithm, so the resulting file has a big size.

You should not use np.compress. Instead, we want to replace all random images by uniform color images, which will have a much smaller size, as they contains very few information.

Shashankjain12:
So @Conchylicultor  i need to write the script which will iterate over all the archive files can be either ```.tar``` or ```.zip``` format  and then find ```.jpg``` or ```.png``` files and decompress them and replace them with their compressed equivalant images  and then recompress that folder containing images.

[...]

PR #1593, opened 2020-03-06
Create random_img_generator.py

vinayvr11:
@Conchylicultor :

# Add fake image generator python script

* Issue Reference: <https://github.com/tensorflow/datasets/issues/1361#issuecomment-595463723>
  
## Description
Add a fake_img_generator.py file in scripts to generate random images that compress better. By replacing the larger images with more compressable equivalents.
<description>
  
## Checklist
- [x] Address all TODO's
- [x] Add alphabetized import to subdirectory's `__init__.py`
- [x] Run `download_and_prepare` successfully
- [x] Add checksums file
- [x] Properly cite in `BibTeX` format
* [x] Lint code

### DIFF
--- a/tensorflow_datasets/scripts/random_image_generator.py
+++ b/tensorflow_datasets/scripts/random_image_generator.py
@@ -1,3 +1,7 @@
+
+""" Generate random images that compress better.
+By replacing the larger images with more compressable equivalents."""
+
 import os
 from zipfile import ZipFile
 from PIL import Image
@@ -22,7 +26,7 @@ def image_process(root):
     """
     images_path.append(root)
     image = np.array(Image.open(root))
-    image = np.ones_like(image) * np.random.randint(np.amax(image))
+    image = np.ones_like(image%255) * np.random.randint(np.amax(image%255))
     image = Image.fromarray(image)
     image = image.convert('RGB')
     image.save(root)
@@ -36,20 +40,20 @@ def rewrite_zip(filepath, filename):
             filename: name of the older zip file that need to be compressed
 
     """
-    # creating a temporary file to store images 
-    temp = tempfile.TemporaryDirectory(dir=filepath)
+     # creating a temporary file to store images 
+    with tempfile.TemporaryDirectory(dir=filepath) as temp:
 
-    # Extraction of compressed .zip file
-    with ZipFile(filepath+filename, 'r') as zip:
-        zip.extractall(path=temp.name)
-        print(temp.name)
+      # Extraction of compressed .zip file
+      with ZipFile(filepath+filename, 'r') as zip:
+          zip.extractall(path=temp.name)
+          print(temp.name)
 
-        process_dir(temp.name+"/")  # Image Processing
+          process_dir(temp.name+"/")  # Image Processing
 
-    # Compressed the .zip file again
-    with ZipFile(filepath+filename, 'w') as zip:
-        for file in images_path:
-            zip.write(file)
+      # Compressed the .zip file again
+      with ZipFile(filepath+filename, 'w') as zip:
+          for file in images_path:
+              zip.write(file)
 
 
 def rewrite_tar(filepath, filename):
@@ -62,18 +66,19 @@ def rewrite_tar(filepath, filename):
 
     """
     # Create a tempfile to store the images with noise
-    temp = tempfile.TemporaryDirectory(dir=filepath)
+    with tempfile.TemporaryDirectory(dir=filepath) as temp:
 
-    # Extraction of .tar file
-    with tarfile.open(filepath+filename, 'r') as tar:
-        tar.extractall(path=temp.name)
-        # Image Process to decrease the size
-        process_dir(temp.name+"/")
+      # Extraction of .tar file
+      with tarfile.open(filepath+filename, 'r') as tar:
+          tar.extractall(path=temp.name)
+          # Image Process to decrease the size
+          process_dir(temp.name+"/")
 
-    # Converting into tarfile again to decrease the space taken by the file-
-    with tarfile.open(filepath+filename, 'w') as tar:
-        for file in images_path:
-            tar_handle.add(os.path.join(filepath, file))
+      # Converting into tarfile again to decrease the space taken by the file-
+      with tarfile.open(filepath+filename, 'w') as tar:
+          for file in images_path:
+              tar_handle.add(os.path.join(filepath, file))
+
 
 
 def process_dir(path):
@@ -88,7 +93,7 @@ def process_dir(path):
     images_path = []
     for root, dirs, files in os.walk(path):
         for file in files:
-            filepath = os.path.join(root, file)
+            filepath = os.path.join(root, file).lower()
             fileName = filepath.split('/')[-1]
 
             if filepath.endswith('.png') or filepath.endswith('.jpg')
--- a/tensorflow_datasets/scripts/random_image_generator.py
+++ b/tensorflow_datasets/scripts/random_image_generator.py
@@ -1,107 +1,132 @@
 
 """ Generate random images that compress better.
-By replacing the larger images with more compressable equivalents."""
+    By replacing the larger images with more compressable equivalents. """
 
 import os
-from zipfile import ZipFile
+import tempfile
+import zipfile
+import tarfile
 from PIL import Image
 import numpy as np
-import tarfile
-import zipfile
-import shutil
-import gzip
+from absl import app
+from absl import flags
 
 
-# images_path variable to store the path of images for later compression
-images_path = []
+FLAGS = flags.FLAGS
 
+flags.DEFINE_string("dir_path",
+                    "testing/test_data/fake_examples",
+                    "path to the directory which contains files")
 
-def image_process(root):
+def image_process(filepath):
     """ Generate random images and remove noise of the
         images to compress better.
 
           Args:
-           root: path of the images to get processed
+           filepath: path of the images to get processed
 
     """
-    images_path.append(root)
-    image = np.array(Image.open(root))
-    image = np.ones_like(image%255) * np.random.randint(np.amax(image%255))
+    image = np.array(Image.open(filepath))
+    grey = int(hash(filepath) % 255)
+    image = np.ones_like(image)*grey
     image = Image.fromarray(image)
     image = image.convert('RGB')
-    image.save(root)
+    image.save(filepath)
 
 
-def rewrite_zip(filepath, filename):
+
+def rewrite_zip(_dir, path):
     """ Rewrite the older .zip files into new better compressed one
 
-        Args:
-            filepath: path of the older zip file that need to be compressed
-            filename: name of the older zip file that need to be compressed
+            Args:
+                _dir: directory path which contain zip compressed file
+                path: path from directory to file
 
     """
-     # creating a temporary file to store images 
-    with tempfile.TemporaryDirectory(dir=filepath) as temp:
+    # creating a temporary file to store images
+    with tempfile.TemporaryDirectory(dir=_dir) as temp:
 
-      # Extraction of compressed .zip file
-      with ZipFile(filepath+filename, 'r') as zip:
-          zip.extractall(path=temp.name)
-          print(temp.name)
+        # Extraction of compressed .zip file
+        with zipfile.ZipFile(path, 'r') as _zp:
+            _zp.extractall(path=temp)
 
-          process_dir(temp.name+"/")  # Image Processing
+        process_dir(temp)  # Image Processing
 
-      # Compressed the .zip file again
-      with ZipFile(filepath+filename, 'w') as zip:
-          for file in images_path:
-              zip.write(file)
+        # Compressed the .zip file again
+        with zipfile.ZipFile(path, 'w') as _zp:
+            for _file_dir, _, files in os.walk(temp):
+                for _file in files:
+                    file_path = os.path.join(_file_dir, _file)
+                    _zp.write(file_path)
 
 
-def rewrite_tar(filepath, filename):
+def rewrite_tar(_dir, path):
     """ Rewrite the older .tar file into new better compressed one.
         Compression formats supports by this method (.tar.gz, .tgz, .tar.bz2)
 
-        Args:
-            filepath: path of the older .tar file that need to be compressed
-            filename: name of the older .tar file that need to be compressed
+            Args:
+              _dir: directory path which contain tar compressed file
+              path: path from directory to file
 
     """
-    # Create a tempfile to store the images with noise
-    with tempfile.TemporaryDirectory(dir=filepath) as temp:
+    # Create a tempfile to store the images contain noise
+    with tempfile.TemporaryDirectory(dir=_dir) as temp:
 
-      # Extraction of .tar file
-      with tarfile.open(filepath+filename, 'r') as tar:
-          tar.extractall(path=temp.name)
-          # Image Process to decrease the size
-          process_dir(temp.name+"/")
+        #Checking the extension of file to be extract
+        ext_list = ["gz", "tgz", "bz2"]
+        read_ext = path.split('.')[-1]
+        write_ext = path.split('.')[-1]
+        if read_ext in ext_list:
+            read_ext = "r:" + read_ext
+            write_ext = "w:" + write_ext
+        else:
+            read_ext = "r"
+            write_ext = "w"
 
-      # Converting into tarfile again to decrease the space taken by the file-
-      with tarfile.open(filepath+filename, 'w') as tar:
-          for file in images_path:
-              tar_handle.add(os.path.join(filepath, file))
+        # Extraction of .tar file
+        with tarfile.open(path, read_ext) as tar:
+            tar.extractall(path=temp)
+        # Image Process to decrease the size
+        process_dir(temp)
+
+        # Converting into tarfile again to decrease the space taken by the file-
+
+        with tarfile.open(path, write_ext) as tar:
+            for _file_dir, _, files in os.walk(temp):
+                for _file in files:
+                    file_path = os.path.join(_file_dir, _file)
+                    tar.add(file_path, recursive=False)
 
 
 
-def process_dir(path):
+def process_dir(dir_path):
     """ This method process the whole directory which contains the
-        compressed files.
+         compressed files.
 
-        Args:
+          Args:
             path: path of the directory which contains all compression files
 
     """
+    img_ext_list = ["jpg", "jpeg", "png"]
+    for _dir, _, _files in os.walk(dir_path):
+        for file in _files:
+            path = os.path.join(_dir, file)
+            _name = os.path.basename(file).split('.')[-1]
+            _name = _name.lower()
+            if _name in img_ext_list:
+                image_process(path)
 
-    images_path = []
-    for root, dirs, files in os.walk(path):
-        for file in files:
-            filepath = os.path.join(root, file).lower()
-            fileName = filepath.split('/')[-1]
+            elif zipfile.is_zipfile(path):
+                rewrite_zip(_dir, path)
 
-            if filepath.endswith('.png') or filepath.endswith('.jpg')
-            or filepath.endswith('.jpeg'):
-                image_process(filepath)
+            elif tarfile.is_tarfile(path):
+                rewrite_tar(_dir, path)
 
-            elif zipfile.is_zipfile(filepath):
-                rewrite_zip(root, fileName)
+def main(unused_argv):
+    """ This method calls the process_dir method to start
+    the process """
+    process_dir(FLAGS.dir_path)
 
-            elif tarfile.is_tarfile(filepath):
-                rewrite_tar(root, fileName)
+
+if __name__ == "__main__":
+    app.run(main)
--- a/tensorflow_datasets/scripts/random_image_generator.py
+++ b/tensorflow_datasets/scripts/random_image_generator.py
@@ -31,9 +31,9 @@ def image_process(root):
 def rewrite_zip(filepath, filename):
     """ Rewrite the older .zip files into new better compressed one
 
-            Args:
-                filepath: path of the older zip file that need to be compressed
-                filename: name of the older zip file that need to be compressed
+        Args:
+            filepath: path of the older zip file that need to be compressed
+            filename: name of the older zip file that need to be compressed
 
     """
     # creating a temporary file to store images 
@@ -56,9 +56,9 @@ def rewrite_tar(filepath, filename):
     """ Rewrite the older .tar file into new better compressed one.
         Compression formats supports by this method (.tar.gz, .tgz, .tar.bz2)
 
-            Args:
-              filepath: path of the older .tar file that need to be compressed
-              filename: name of the older .tar file that need to be compressed
+        Args:
+            filepath: path of the older .tar file that need to be compressed
+            filename: name of the older .tar file that need to be compressed
 
     """
     # Create a tempfile to store the images with noise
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/scripts/random_img_generator.py
```diff
@@ -0,0 +1,102 @@
...
+
+# images_path variable to store the path of images for later compression
+images_path = []
+
+
+def image_process(root):
+    """ Generate random images and remove noise of the
+        images to compress better.
+
+          Args:
```
ashutosh1919: Please take a note of indentation in descriptions. You should change that as follows:
```py
""" 
Generate random images and remove noise of the
images to compress better.

Args:
  root: path of the images to get processed
"""
```
Also, update the indentation in other methods too in the same way.

tensorflow_datasets/scripts/random_img_generator.py
```diff
@@ -0,0 +1,102 @@
...
+
+            if filepath.endswith('.png') or filepath.endswith('.jpg')
+            or filepath.endswith('.jpeg'):
+                image_process(filepath)
+
+            elif zipfile.is_zipfile(filepath):
+                rewrite_zip(root, fileName)
+
+            elif tarfile.is_tarfile(filepath):
+                rewrite_tar(root, fileName)
```
ashutosh1919: please leave one empty line at the end of file.

tensorflow_datasets/scripts/random_img_generator.py
vinayvr11: Thanks you ashutosh i will make the changes in indentations and this code is pylint resolved. I have already run pylint on this code.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,102 @@
...
+def rewrite_zip(filepath, filename):
+    """ Rewrite the older .zip files into new better compressed one
+
+        Args:
+            filepath: path of the older zip file that need to be compressed
+            filename: name of the older zip file that need to be compressed
+
+    """
+    # creating a temporary file to store images 
+    temp = tempfile.TemporaryDirectory(dir=filepath)
```
Conchylicultor: `with tempfile.TemporaryDirectory(dir=filepath)` to automatically cleanup the temporary file afterward

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,102 @@
+import os
```
Conchylicultor: Could you run pylint to format the file. You also need to add a top level docstring to explain what this file does.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,102 @@
...
+    """ Generate random images and remove noise of the
+        images to compress better.
+
+          Args:
+           root: path of the images to get processed
+
+    """
+    images_path.append(root)
+    image = np.array(Image.open(root))
+    image = np.ones_like(image) * np.random.randint(np.amax(image))
```
Conchylicultor: As explained previously: For the random generation, rather than pure randomness, we should make it deterministic such as running the script twice produce the same image. One way is to use the hash of the relative path % 255. As the relative path is constant across run, the images will be too.

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: > Thank you for the PR. Could you also send another PR of the script applied on our fake_data dir to see what the result is ?

@Conchylicultor  : Sorry but i want to know something more about this. Could you share me the link of fake_data dir.  And what i need to do in that directory. 
Thank you

Conchylicultor: How did you tested your script ? The goal of this issue is to compress the size of our fake data dir: https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/testing/test_data/fake_examples

You should test the script on this folder to make sure this works. Also could you report the directory size before/after image compression ?

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,102 @@
...
+
+    """
+
+    images_path = []
+    for root, dirs, files in os.walk(path):
+        for file in files:
+            filepath = os.path.join(root, file)
+            fileName = filepath.split('/')[-1]
+
+            if filepath.endswith('.png') or filepath.endswith('.jpg')
```
Conchylicultor: `.lower()` to also catch `.PNG`, `.JPEG`,... files

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: I have already format the file with pylint. Should i need to add the license and the copyright comment as well.

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: I have tested the script with my own data. But i will report you the results within some minutes.

Conchylicultor: Awesome thank you.

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: Could you use https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style ?

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: So i need to compress the whole directory or i just check my script on one file of that directory.

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: Done

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: No need to add the license. It will be added automatically. But thank you for asking.

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: Try to use the script on the full directory to make sure this works as expected. Thank you.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Args:
+            path: path of the directory which contains all compression files
+
+    """
+
+    images_path = []
+    for root, dirs, files in os.walk(path):
+        for file in files:
+            filepath = os.path.join(root, file).lower()
```
Conchylicultor: If you lower the path here, the `image_process(filepath)` will be invalid, no ?

vinayvr11: It works code works properly and decrease the size from 255mb to 152mb without any error. And also refined the code in pylint so now i am making an new PR with updated Fake_data_utils.

vinayvr11: So shall i remove the old data with the new and compressed one or just add the compressed data without removing the previous one.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+      # Extraction of compressed .zip file
+      with ZipFile(filepath+filename, 'r') as zip:
+          zip.extractall(path=temp.name)
+          print(temp.name)
+
+          process_dir(temp.name+"/")  # Image Processing
+
+      # Compressed the .zip file again
+      with ZipFile(filepath+filename, 'w') as zip:
+          for file in images_path:
```
Conchylicultor: * If there is multiple zip files, the second zip files will compress all recorded images, including the ones which do not belong to the zip file.
* Also if the zip file was containing other files, they are lost when compressing the zip files.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+
+      # Extraction of compressed .zip file
+      with ZipFile(filepath+filename, 'r') as zip:
+          zip.extractall(path=temp.name)
+          print(temp.name)
+
+          process_dir(temp.name+"/")  # Image Processing
+
+      # Compressed the .zip file again
+      with ZipFile(filepath+filename, 'w') as zip:
```
Conchylicultor: Use `os.path.join()` instead of + (here and bellow). And remove the `+"/"`

Otherwise, the script won't works on Windows.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+    """ Generate random images and remove noise of the
+        images to compress better.
+
+          Args:
+           root: path of the images to get processed
+
+    """
+    images_path.append(root)
+    image = np.array(Image.open(root))
+    image = np.ones_like(image%255) * np.random.randint(np.amax(image%255))
```
Conchylicultor: I'm not sure what you're trying to compute here. To get deterministic value, you shouldn't use `np.random.randint` anymore as it will compute a new value each time the script is executed.

We want a deterministic mapping `filepath -> integer` to generate the grey_scale between 0 and 255.
and then compress the image like:

1) Extract the relative tfds path from the `filepath` to relative_filepath.
Ex: `/home/user/xyz/tensorflow_dataset/some_path/file.jpg` -> `some_path/file.jpg`

2) Generate the integer by hashing the file path
```
grey_scale = hash_fn(relative_filepath) % 255
```
3) Create the new image.
```
image = np.ones_like(image) * `grey_scale`
```

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+
+            if filepath.endswith('.png') or filepath.endswith('.jpg')
+            or filepath.endswith('.jpeg'):
+                image_process(filepath)
+
+            elif zipfile.is_zipfile(filepath):
+                rewrite_zip(root, fileName)
+
+            elif tarfile.is_tarfile(filepath):
+                rewrite_tar(root, fileName)
```
Conchylicultor: Could you add `absl.app(main)` like other scripts/ ? And add a flag to control the root dir ?

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+import zipfile
+import shutil
+import gzip
+
+
+# images_path variable to store the path of images for later compression
+images_path = []
+
+
+def image_process(root):
```
Conchylicultor: I would rename this `filepath`.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+        Args:
+            path: path of the directory which contains all compression files
+
+    """
+
+    images_path = []
+    for root, dirs, files in os.walk(path):
+        for file in files:
+            filepath = os.path.join(root, file).lower()
+            fileName = filepath.split('/')[-1]
```
Conchylicultor: Please use snake_case.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,107 @@
...
+          process_dir(temp.name+"/")
+
+      # Converting into tarfile again to decrease the space taken by the file-
+      with tarfile.open(filepath+filename, 'w') as tar:
+          for file in images_path:
+              tar_handle.add(os.path.join(filepath, file))
+
+
+
+def process_dir(path):
```
Conchylicultor: By convention:

* `...name`: A file name, ex: `abc.jpg`
* `_dir`: A directory path, ex `/abc/def/`
* `...path`: A file path, ex `/abc/def/abc.jpg`

Could you rename all variables to follow this convention ? This helps the reader to understand the code.

vinayvr11: Done

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: Done

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: done

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: I have a doubt in this because i have never used this so could you please share me the link to learn about this. 
Thank you

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: Done

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: https://abseil.io/docs/python/quickstart#creating-and-running-a-script

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: @Conchylicultor : I make a new script with all the changes you suggested. Now just a last question shall i need to add the testing file also using abseil testing and if yes so what should i suppose to add in that file.
Thank you

vinayvr11: @Conchylicultor : I have updated the file with changes.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,132 @@
+
```
Conchylicultor: Thank you for the update. Could you format your code correctly with 
https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,132 @@
...
+    image = np.array(Image.open(filepath))
+    grey = int(hash(filepath) % 255)
+    image = np.ones_like(image)*grey
+    image = Image.fromarray(image)
+    image = image.convert('RGB')
+    image.save(filepath)
+
+
+
+def rewrite_zip(_dir, path):
```
Conchylicultor: `_` means private. And it's good to have explicit names so the user can understand the variable content just from the variable names.
I would rename the args `root_dir, zip_filepath`, and similarly below.

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,132 @@
...
+    """ Rewrite the older .zip files into new better compressed one
+
+            Args:
+                _dir: directory path which contain zip compressed file
+                path: path from directory to file
+
+    """
+    # creating a temporary file to store images
+    with tempfile.TemporaryDirectory(dir=_dir) as temp:
```
Conchylicultor: `as temp_dir:`

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,132 @@
...
+
+          Args:
+            path: path of the directory which contains all compression files
+
+    """
+    img_ext_list = ["jpg", "jpeg", "png"]
+    for _dir, _, _files in os.walk(dir_path):
+        for file in _files:
+            path = os.path.join(_dir, file)
+            _name = os.path.basename(file).split('.')[-1]
```
Conchylicultor: Function variables shouldn't start with trailing `_`. How about renaming this `img_ext`?

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: Rather that testing, could you send a new PR of the script applied to the fake data dir ? to show this is working properly ?

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,132 @@
...
+              _dir: directory path which contain tar compressed file
+              path: path from directory to file
+
+    """
+    # Create a tempfile to store the images contain noise
+    with tempfile.TemporaryDirectory(dir=_dir) as temp:
+
+        #Checking the extension of file to be extract
+        ext_list = ["gz", "tgz", "bz2"]
+        read_ext = path.split('.')[-1]
```
Conchylicultor: Rather than recomputing twice the extension. How about something like:

```
tar_ext = path.split('.')[-1]
with tarfile.open(path, _make_tar_mode('r', tar_ext)) as tar:
  ...

with tarfile.open(path, _make_tar_mode('w', tar_ext)) as tar:
  ...
```

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,132 @@
...
+        # Image Process to decrease the size
+        process_dir(temp)
+
+        # Converting into tarfile again to decrease the space taken by the file-
+
+        with tarfile.open(path, write_ext) as tar:
+            for _file_dir, _, files in os.walk(temp):
+                for _file in files:
+                    file_path = os.path.join(_file_dir, _file)
+                    tar.add(file_path, recursive=False)
```
Conchylicultor: Why not just have a single `tar.add(file_path, recursive=True)` instead ?

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: I already format the code using pylint and that shows 10/10 results.

Conchylicultor: We are not using the standard pytype, but the Google python style guide. Please use instructions at: https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style

vinayvr11: Sure

tensorflow_datasets/scripts/random_image_generator.py
```diff
@@ -0,0 +1,132 @@
...
+
+          Args:
+           filepath: path of the images to get processed
+
+    """
+    image = np.array(Image.open(filepath))
+    grey = int(hash(filepath) % 255)
+    image = np.ones_like(image)*grey
+    image = Image.fromarray(image)
+    image = image.convert('RGB')
```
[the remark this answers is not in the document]
vinayvr11: No actually i forgot to remove this.

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: Is this a method _make_tar_mode()

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: like this ->
tar.add(temp_dir , recursive=False)

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: So i need to use lint.sh in oss_scripts.

Conchylicultor: Yes please, as explained in https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: Yes, basically create a new private function `_make_tar_mode` to reuse the logic between `r` and `w`. As the function shouldn't be in other files, we mark it as private (starts with `_`)

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: I meant, rather than compressing each files individually, why don't you compress the full folder at once.

vinayvr11: Yes now this clicks in my mind this will also remove these loops. You are amazing

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: but instead of making a separate function for this we can also did this as ('r'+tar_ext) for read and ('w' + tar_ext) for write.

Conchylicultor: Yes, this works too

tensorflow_datasets/scripts/random_image_generator.py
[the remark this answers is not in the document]
vinayvr11: Actually if i remove this then it shows an error that image is not converting into jpeg.

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: If i compress the full folder by giving the folder path then it only compress the folder not images so i need to iterate over all the images to make them compress and that's how tarfile.add() works.

tensorflow_datasets/scripts/random_image_generator.py
vinayvr11: could you please help  me with lint.sh i got error while running this script like unexpected end of file this is just the last step that remains. Please help me with this i am using windows.
Thank you

Conchylicultor: Without knowing which error you're getting it's difficult to help. Can you copy the full message you're getting ?

vinayvr11: lint.sh: line 2: $'\r': command not found
lint.sh: line 4: $'\r': command not found
lint.sh: line 11: syntax error: unexpected end of file

tensorflow_datasets/scripts/random_image_generator.py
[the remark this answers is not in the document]
vinayvr11: cannot write mode RGBA as JPEG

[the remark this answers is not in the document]
vinayvr11: This is because some images contains alpha channel.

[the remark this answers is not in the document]
Conchylicultor: You should save the images the same way they were. If the images has 4 channels, you should save it as 4 channels, if the images has 1 channel, you should save it with 1 channel.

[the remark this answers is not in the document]
Conchylicultor: The question is how loading a jpeg channel load 4 channels. Something feels wrong to me.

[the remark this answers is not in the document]
vinayvr11: Ok so shall i remove this or not.

tensorflow_datasets/scripts/random_image_generator.py
Conchylicultor: The script may not be available on windows.

You'll have to manually install pylint and use it with this config file: https://raw.githubusercontent.com/tensorflow/tensorflow/master/tensorflow/tools/ci_build/pylintrc

tensorflow_datasets/scripts/random_image_generator.py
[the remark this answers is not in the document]
vinayvr11: Because pillow JPEG doesn't allow RGBA. Now what should i do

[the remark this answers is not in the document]
Conchylicultor: You should try to investigate why `np.array(Image.open('x.jpeg'))` shape is `(h, w, 4)`. For instance, for which path does it happen ? This might indicates that one of the images is a png saved as jpeg.

The script should save images the same way they are loaded, so if `x.jpeg` had 4 channels, then the saved images should have 4 channels too.

Just to get a first overview, could you send a PR of the fake_data dir modified with your script (without this fix) so I can get an overview of if the overall script is working or not ?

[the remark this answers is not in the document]
vinayvr11: Sure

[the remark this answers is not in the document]
vinayvr11: I have sended you a new PR. Please have a look.

Conchylicultor:
Closing in favor of #1661

PR #1606, opened 2020-03-08
change some download urls to https

Ir1d:
- I've also edited the TOC in README, but I think it's a bit weird, because the headings are not aligned to `<h2>`.. Maybe the headings in the README need some further alignment
- I've changed some url for download.tensorflow.org , because it's suffering from cert error at this moment
- changed some download urls to https for better user experience. I've checked the edited urls and they work perfectly.

### DIFF
--- a/tensorflow_datasets/image/caltech_birds.py
+++ b/tensorflow_datasets/image/caltech_birds.py
@@ -57,9 +57,9 @@ class CaltechBirds2010(tfds.core.GeneratorBasedBuilder):
   def _caltech_birds_info(self):
     return CaltechBirdsInfo(
         name=self.name,
-        images_url="https://www.vision.caltech.edu/visipedia-data/CUB-200/images.tgz",
-        split_url="https://www.vision.caltech.edu/visipedia-data/CUB-200/lists.tgz",
-        annotations_url="https://www.vision.caltech.edu/visipedia-data/CUB-200/annotations.tgz",
+        images_url="http://www.vision.caltech.edu/visipedia-data/CUB-200/images.tgz",
+        split_url="http://www.vision.caltech.edu/visipedia-data/CUB-200/lists.tgz",
+        annotations_url="http://www.vision.caltech.edu/visipedia-data/CUB-200/annotations.tgz",
     )
 
   def _info(self):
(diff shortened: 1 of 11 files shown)


tensorflow_datasets/image/caltech_birds.py
```diff
@@ -57,9 +57,9 @@ class CaltechBirds2010(tfds.core.GeneratorBasedBuilder):
   def _caltech_birds_info(self):
     return CaltechBirdsInfo(
         name=self.name,
-        images_url="http://www.vision.caltech.edu/visipedia-data/CUB-200/images.tgz",
-        split_url="http://www.vision.caltech.edu/visipedia-data/CUB-200/lists.tgz",
-        annotations_url="http://www.vision.caltech.edu/visipedia-data/CUB-200/annotations.tgz",
+        images_url="https://www.vision.caltech.edu/visipedia-data/CUB-200/images.tgz",
```
Conchylicultor: Are you sure you tested this correctly ? I'm getting "website can't be accessed" with this new URL. I haven't checked the other ones.

tensorflow_datasets/image/food101.py
```diff
@@ -25,7 +25,7 @@
 import tensorflow.compat.v2 as tf
 import tensorflow_datasets.public_api as tfds
 
-_BASE_URL = "http://data.vision.ee.ethz.ch/cvl/food-101.tar.gz"
+_BASE_URL = "https://data.vision.ee.ethz.ch/cvl/food-101.tar.gz"
```
Conchylicultor: If you modify the urls, could you re-generate the checksums files with `--register_checksums` ?
https://www.tensorflow.org/datasets/add_dataset#2_run_download_and_prepare_locally

tensorflow_datasets/image/caltech_birds.py
Ir1d: This is weird :smile: perhaps it was a mistake.. It's not working now, I'll revert this line

Ir1d:
@Conchylicultor Thanks. I didn't realize that needed regenerating the checksum, because some urls are exactly identical. Anyway, I found that due to network and disk limitation, I couldn't calc the checksum of some datasets. So I reverted those lines. Besides, I found that `python -m tensorflow_datasets.scripts.download_and_prepare --register_checksums --datasets=flower` was not successful for `tensorflow_datasets/image/flowers.py`.. It was a bit weird.

PR #1621, opened 2020-03-10
Remove python 2 legacy from download_manager.py

Eshan-Agarwal:
As python 2 use is been dropped. So Remove unnecessary code from `download_manager.py` .

@Conchylicultor please have a look

### DIFF
--- a/tensorflow_datasets/core/download/download_manager.py
+++ b/tensorflow_datasets/core/download/download_manager.py
@@ -16,9 +16,6 @@
 # Lint as: python3
 """Download manager interface."""
 
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
 
 import os
 import sys
@@ -388,11 +385,9 @@ class DownloadManager(object):
               self._manual_dir, self._manual_dir_instructions))
     return self._manual_dir
 
-  
-if sys.version_info[0] > 2:
 
-  def _wait_on_promise(p):
-    return p.get()
+def _wait_on_promise(p):
+  return p.get()
 
 def _map_promise(map_fn, all_inputs):
   """Map the function into each element and resolve the promise."""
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/core/download/download_manager.py
```diff
@@ -389,29 +389,7 @@ def manual_dir(self):
...
 
-# ============================================================================
-# In Python 2.X, threading.Condition.wait() cannot be interrupted by SIGINT,
-# unless it's given a timeout. Here we artificially give a long timeout to
-# allow ctrl+C.
-# This code should be deleted once python2 is no longer supported.
-if sys.version_info[0] > 2:
-
-  def _wait_on_promise(p):
-    return p.get()
```
Conchylicultor: You still need to resolve the promise. The code in `sys.version_info[0] > 2` is executed for Python3 so it should be kept.

Eshan-Agarwal: Okay will make changes. Can you please tell me what is p here I didn't get it.

tensorflow_datasets/core/download/download_manager.py
```diff
@@ -388,8 +388,14 @@ def manual_dir(self):
               self._manual_dir, self._manual_dir_instructions))
     return self._manual_dir
 
+  
+if sys.version_info[0] > 2:
```
Conchylicultor: Now that this is always true, the condition can be removed.

Eshan-Agarwal: Yeah, I was also thinking of it . Thanks

tensorflow_datasets/core/download/download_manager.py
Conchylicultor: `p` means `promise` from https://pypi.org/project/promise/

Issue #1534, opened 2020-02-27
[GSoC] Script to cleanup the testing/metadata dir


Conchylicultor:
**Task**: Our [metadata folder](https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/testing/metadata) is growing big, but some of the content is obsolete and do not match any existing version. We should add a script to cleanup the metadata versions which do not exists anymore.

**Instruction**

* Add an util to `registered.py` to get all available versions:

```python
def _get_all_versions(version_list):
  return set(str(v) for v in version_list)

def iter_dataset_full_names():
  """Yield all supported datasets full_names."""
  for builder_name, builder_cls in registered._DATASET_REGISTRY.items():  # pylint: disable=protected-access
    if builder_cls.BUILDER_CONFIGS:
      for config in builder_cls.BUILDER_CONFIGS:
        for v in _get_all_versions(
            [config.version] + config.supported_versions):
          yield os.path.join(builder_name, config.name, v)
    else:
      for v in _get_all_versions(
          [builder_cls.VERSION] + builder_cls.SUPPORTED_VERSIONS):
        yield os.path.join(builder_name, v)
```

* Add a script in `scripts/` which iterate over the all metadata directories and check if the metadata match a registered versions or not. If not, delete the metadata directory.
* The script should have a `--dry_run` option which print the folders to delete but without actually deleting the files.

**Dificulty:** Easy

Eshan-Agarwal:
Hey! I want to confirm dew things.
 The code snippet you given above is a util to `registered.py` which get all available versions of datasets ?
And the code which we have to add in `scripts/` have `--dry_run` option which only prints folders to delete ?

rohanreddych:
@Conchylicultor the `iter_dataset_full_names` is returning empty generator because `_DATASET_REGISTRY` is empty how should I test my script?

rohanreddych:
@Conchylicultor Github skips through the empty directories do you want the script to remove those directories.

Conchylicultor:
@Eshan-Agarwal yes, exactly

@rohanreddych To make `iter_dataset_full_names` return datasets, you have to import tensorflow_datasets as datasets are registered in __init__.py.
The goal of the script is to remove folders which correspond to old versions. Ex: `metadata/mnist/1.0.0/` should be deleted as `mnist` only exists in version `3.*.*`

rohanreddych:
Thank you will make the changes now.

ManishAradwad:
[...]

PR #1629, opened 2020-03-11
Cleanup script to delete older dataset versions

ManishAradwad:
## Description
Issue : #1534  

Hello @Conchylicultor the script you've provided is returning an empty generator object. I tried running `tfds.list_builders()` in cleanup_metadata.py and it is returning the list of all available datasets, but when I run `iter_dataset_full_names` from registered.py I'm still getting empty object. Could you plz help?

### DIFF
--- a/tensorflow_datasets/core/registered.py
+++ b/tensorflow_datasets/core/registered.py
@@ -377,3 +377,20 @@ def _cast_to_pod(val):
       return float(val)
     except ValueError:
       return tf.compat.as_text(val)
+
+
+def _get_all_versions(version_list):
+  return set(str(v) for v in version_list)
+
+def iter_dataset_full_names():
+  """Yield all supported datasets full_names."""
+  for builder_name, builder_cls in _DATASET_REGISTRY.items():  # pylint: disable=protected-access
+    if builder_cls.BUILDER_CONFIGS:
+      for config in builder_cls.BUILDER_CONFIGS:
+        for v in _get_all_versions(
+            [config.version] + config.supported_versions):
+          yield os.path.join(builder_name, config.name, v)
+    else:
+      for v in _get_all_versions(
+          [builder_cls.VERSION] + builder_cls.SUPPORTED_VERSIONS):
+        yield os.path.join(builder_name, v)
--- a/tensorflow_datasets/scripts/cleanup_metadata.py
+++ b/tensorflow_datasets/scripts/cleanup_metadata.py
@@ -1,0 +1,17 @@
+import tensorflow_datasets as tfds
+import os
+import sys
+
+# Returns metadata dir content
+metadata_dir = os.walk(os.path.join(os.pardir,"testing/metadata"))
+roots = [root for root,dir,file in metadata_dir]
+
+# Import registered.py
+reg_path = os.path.join(os.pardir, "core")
+sys.path.insert(1, os.path.abspath(reg_path))
+import registered
+
+# Get list of all available versions of datasets
+registered_versions = [i for i in registered.iter_dataset_full_names()]
+print(registered_versions)
+print(tfds.list_builders())


Conchylicultor:
To debug this, you could try to print the values, for instance what is the content of `_DATASET_REGISTRY` ? Is `list_builder()` working properly ? And then try to progressively find the source of the issue. If `list_builder()`is working but not `iter_dataset_full_names`, you can start from the `list_builder` code and gradually add feature until localising the issue.

rohanreddych:
@Conchylicultor the _DATASET_REGISTRY is empty `_DATASET_REGISTRY = {}` is there any init script which initializes this dict because there is no function in this file which initializes it

Conchylicultor:
I'm talking about printing `_DATASET_REGISTRY` inside the `iter_dataset_full_names` function, which is the function you're trying to debug.
Datasets are automcatically added to `_DATASET_REGISTRY` when they are imported, in your case in `import tensorflow_datasets as tfds`. If `list_builder` is not empty, then neither should `_DATASET_REGISTRY`.

ManishAradwad:
@Conchylicultor Actually, `list_builder` is also retuning an empty set.
I tried importing `tfds` in `registered.py` and then printing `_DATASET_REGISTRY` and `list_builders()`, they both are empty. 
I saw there is a function `__new__` under `RegisteredDataset` class in registered.py which is providing values for `_DATASET_REGISTRY`. Could it be of some help??

Conchylicultor:
@ManishAradwad the datasets are added on import. So importing TFDS (`import tensorflow_datasets as tfds`) should import all datasets.

Is `tfds.image.mnist.MNIST()` available ? If so, it should be registered in the `_DATASET_REGISTRY`, otherwise, you should investigate why it doesn't get imported.

Conchylicultor:
> I tried importing tfds in registered.py

You should import tfds inside your `scripts/cleanup_metadata.py`. Not inside `registered.py`.

And you should test the script through executing `cleanup_metadata.py`. And check there that the `list_builder` & cie are not empty.

ManishAradwad:
Yes, actually I've done the exact same thing. I imported `tfds` in `registered.py` to see if `_DATASET_REGISTRY` updates or not after running only `registered.py`(and eventually it doesn't).

> Is `tfds.image.mnist.MNIST()` available ? If so, it should be added in the `_DATASET_REGISTRY`, otherwise, you should investigate why it doesn't get imported.

I'll try doing this.

Conchylicultor:
Hey @ManishAradwad , any updates on this ?

ManishAradwad:
@Conchylicultor I tried using `tfds.image.mnist.MNIST()`, `cleanup_metadata.py` runs fine without any errors but I'm still getting empty `_DATASET_REGISTRY` and `list_builders()`

ManishAradwad:
To make sure I'm doing everything correctly here's the list of things I'm doing:
- In `registered.py`:
Added 3 functions - `_get_all_versions` , `iter_dataset_full_names` and `print_dataset_registry`(prints `_DATASET_REGISTRY` and `list_builders()`)

- In `cleanup_metadata.py`:
Imported `tfds`
Imported `registered.py`
Called `tfds.image.mnist.MNIST()`
Printed `iter_dataset_full_names` output as a list
Called `print_dataset_registry`

Conchylicultor:
Can anyone help on this ?

Conchylicultor:
@ManishAradwad you could look at the `download_and_prepare.py` script and start from there to see how this is done. This script correctly loads all dataset list.

ManishAradwad:
@Conchylicultor I tried going through `download_and_prepare.py` but couldn't get how to fetch the versions of all datasets. 

Also regarding our previous approach, is it possible to use `iter_dataset_full_names` like `tfds.list_builders()`(i.e. as `tfds.iter_dataset_full_names()`)?

I tried this approach by adding `from tensorflow_datasets.core.registered import iter_dataset_full_names` in `public_api.py` and running `tfds.iter_dataset_full_names()`. But I'm still getting no attribute error:
```AttributeError: module 'tensorflow_datasets' has no attribute 'iter_dataset_full_names'```  
Could you plz help?

vijayphoenix:
> Can anyone help on this?

@ManishAradwad If you want you can collaborate with #1730 as you both are working on the same issue.

Eshan-Agarwal:
@vijayphoenix  Its almost finish just have to change few things I don't think this have any work remaining wotk.

PR #1631, opened 2020-03-11
Better compression of fake_examples

vinayvr11:
# Add Dataset
@Conchylicultor 
* Dataset Name: <better_compresssed_fake_examples>
 
## Description
Better compressed fake_examples.
<description>


Conchylicultor:
That you very much for the update!
It seems that the script is modifying a lot of non-images files. This means there is something wrong in the script. Only images files should be modified.
Could you have a look ?

vinayvr11:
> That you very much for the update!
> It seems that the script is modifying a lot of non-images files. This means there is something wrong in the script. Only images files should be modified.
> Could you have a look ?

Sure

vinayvr11:
Could you send me the name or any refrence to that  file.

Conchylicultor:
Oh I see what you did. The script should replace the old images, not duplicate all data. Could you update the PR with the new script overwriting the data, rather than duplicating it ?

vinayvr11:
> Oh I see what you did. The script should replace the old images, not duplicate all data. Could you update the PR with the new script overwriting the data, rather than duplicating it ?

But i am saving the images at their old path.

Conchylicultor:
Then why is there a `better_compressed_fake_examples` in all paths in your PR ?

vinayvr11:
> Then why is there a `better_compressed_fake_examples` in all paths in your PR ?

Because that's the different directory which contains all updated files.

Conchylicultor:
> Because that's the different directory which contains all updated files.

Instead of a new directory, the script should replace the old images, not duplicate all data. Could you update the PR with the new script overwriting the data, rather than duplicating it ?

vinayvr11:
> > Because that's the different directory which contains all updated files.
> 
> Instead of a new directory, the script should replace the old images directly, and keep the same path. Not duplicate all data in a new directory. Could you update the PR with the overwritten data, rather than a duplicated folder ?

I have debug the error the error is in (chexpert/CheXpert-v1.0-small/train/patient0/study0/view1_frontal.jpg)  this is an 4 channel image and pillow doesn't supports the RGBA images so i made a condition to check the RGBA image and convert only the RGBA into RGB not all images. and it works fine.

Conchylicultor:
>  i made a condition to check the RGBA image

Thanks for the follow up. This is better!
I'm starting to wondering if Pillow is really the good library for this task as it seems limited (not support all image types, lose information about the image dtype,...). Do you think opencv would work better ?

Conchylicultor:
Closing this in favor of the new PR

PR #1639, opened 2020-03-12
visual support for object detection dataset

VaranRohila:
Added visualization support for object detection dataset.

Procedure:
1. Find out all the keys with the Sequence feature connector (Bbox are wrapped inside them).
2. Find the keys with BboxFeature in Sequence keys.
3. Find the labels if there are any in the Sequence keys.
4. Plot image, bboxes, and labels on them.

We only take the first Sequence feature, first bbox feature and first label feature for now.

Notebook demonstration: [Link](https://colab.research.google.com/drive/1801-vXCENw3w-ptNHh4eel2D1qHS9s9o#scrollTo=qFo0uHZ3CG11)


tensorflow_datasets/core/visualization.py
```diff
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Lint as: python3
```
Conchylicultor: Why remove this ?

tensorflow_datasets/core/visualization.py
```diff
@@ -28,6 +27,18 @@
 from tensorflow_datasets.core import lazy_imports_lib
 
 
+matplotlib = lazy_imports_lib.LazyImporter.matplotlib
```
Conchylicultor: The goal of lazy import is to avoid imports in the global scope, for people not using the feature. This should be moved inside the function, otherwise it will crash for users not having installed matplotlib

tensorflow_datasets/core/visualization.py
```diff
@@ -28,6 +27,18 @@
...
+
+def draw_text(ax, xy, txt, sz=14):
+    text = ax.text(*xy, txt, verticalalignment='top', color='white', fontsize=sz, weight="normal",bbox=dict(facecolor='black', alpha=0.7, pad=0.3))
```
Conchylicultor: Could you format your code using https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style

tensorflow_datasets/core/visualization.py
```diff
@@ -85,12 +96,37 @@ def show_examples(ds_info, ds, rows=3, cols=3, plot_scale=3., image_key=None):
...
+  sequence_keys = [
       k for k, feature in ds_info.features.items()
-      if isinstance(feature, features_lib.ClassLabel)
+      if isinstance(feature, features_lib.Sequence)
   ]
 
-  label_key = label_keys[0] if len(label_keys) == 1 else None
+  bbox_keys = []
+  sequence_key = None
+  for key in sequence_keys:
```
Conchylicultor: Add comment to give a high level description of what each sections of your code is doing

tensorflow_datasets/core/visualization.py
```diff
@@ -85,12 +96,37 @@ def show_examples(ds_info, ds, rows=3, cols=3, plot_scale=3., image_key=None):
 
     image_key = image_keys[0]
 
-  label_keys = [
+  sequence_keys = [
```
Conchylicultor: This functions starts having too many conditions/variables. Making the code harder to read. I think it would make sense to refactor it in smaller modules/functions. For instance:

```python
def show_examples(ds_info, ds, **options_kwargs):
    if visualizer.match(ds_info, ds):
      return visualizer.build(ds_info, ds, **options_kwargs)
  else:
    raise ValueError(f"Visualisation not supported for dataset `{ds_info.name}`")
```

tensorflow_datasets/core/visualization.py
VaranRohila: Sorry about that. That was by mistake

VaranRohila:
@Conchylicultor I have made some major changes in the structure of the module as you had suggested. There is one main file `visualization.py` and two visualizer modules `image_visualizer.py` and `object_visualizer.py`

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+          k for k, feature in ds_info.features[key].items()
+          if isinstance(feature, features_lib.BBoxFeature)
+      ]
+
+  if bbox_keys:
+    return False
+
+  return True
+
+def build(ds_info, ds, **kwargs):
```
Conchylicultor: You should replace `**kwargs` with actual argument names.

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Lint as: python3
+
+"""Visualizer for image datasets.
```
Conchylicultor: Rather than a module, it should be a class and have `SupervisedVisualizer` and `ObjectVisualizer` depend on it:

```
class Visualizer(object):

  def match(self,...)

   ...
```

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+from tensorflow_datasets.core import features as features_lib
+from tensorflow_datasets.core import lazy_imports_lib
+
+
+def match(ds_info):
+  """Checks if dataset belongs to image dataset category"""
+  image_keys = [
+      k for k, feature in ds_info.features.items()
+      if isinstance(feature, features_lib.Image)
+  ]
```
Conchylicultor: When you are re-using the same pattern multiple times, it's good to make a function and reuse it every-where. For instance:

`image_keys = _extract_keys(ds_info, features_lib.Image)`

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+
+  for key in kwargs:
+    if key not in args:
+      raise TypeError(
+          "show_example() got an unexpected keyword argument '{}'".format(key)
+          )
+
+  plt = lazy_imports_lib.lazy_imports.matplotlib.pyplot
+
+  if not image_key:
```
Conchylicultor: Rather than duplicating the code, you should factorise in function and reuse them.
Similarly below

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,129 @@
...
+
+  if not image_keys:
+    return False
+
+  sequence_keys = [
+      k for k, feature in ds_info.features.items()
+      if isinstance(feature, features_lib.Sequence)
+  ]
+
+  bbox_keys = []
```
Conchylicultor: You shouldn't test for bbox in the image visualizer. Instead, just make the bbox visualizer be checked before the image one.

tensorflow_datasets/core/object_visualizer.py
```diff
@@ -0,0 +1,178 @@
...
+  if not label_key:
+    logging.info("Was not able to auto-infer label.")
+  bbox_key = bbox_keys[0] if len(bbox_keys) > 0 else None
+
+  num_examples = rows * cols
+  examples = list(dataset_utils.as_numpy(ds.take(num_examples)))
+
+  fig = plt.figure(figsize=(plot_scale*cols, plot_scale*rows))
+  fig.subplots_adjust(hspace=1/plot_scale, wspace=1/plot_scale)
+  for i, ex in enumerate(examples):
```
Conchylicultor: Rather than duplicating all this between the visualizer, is there a way to reuse more the code ?

VaranRohila:
@Conchylicultor I have made changes according to your suggestions. Created an `ImageVisualizer` base class and two child classes, `SupervisedVisualizer` and `ObjectVisualizer`.  

Could you suggest some methods to reuse the code inside the build function? I was thinking of using python decorators but that would certainly increase the code complexity quite a bit.

VaranRohila:
Update: I checked my code on wider_faces dataset as well. It is working fine, added that to the colab notebook for demonstration. I have also added a `bbox_label` flag to the parameters so that the label is not plotted for the bbox (for the datasets that have too many bbox and labels associated with them).

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,222 @@
...
+from absl import logging
+
+from tensorflow_datasets.core import dataset_utils
+from tensorflow_datasets.core import features as features_lib
+from tensorflow_datasets.core import lazy_imports_lib
+
+class ImageVisualizer():
+  """Parent Class for support of Image based datasets"""
+
+  def _extract_keys(self, ds_info, feature_type, secondary_key=None):
```
Conchylicultor: Please use functions instead of methods when you're not using `self`. This make the code easier to read (as it makes it clear for the reader what the input/outputs are (no hidden `self._xyz` inputs/outputs).

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,222 @@
...
+    # for BBoxes as in voc dataset
+    label_key = label_keys[0] if len(label_keys) > 0 else None
+    if not label_key:
+      logging.info("Was not able to auto-infer label.")
+    bbox_key = bbox_keys[0] if len(bbox_keys) > 0 else None
+
+    num_examples = rows * cols
+    examples = list(dataset_utils.as_numpy(ds.take(num_examples)))
+
+    fig, plt = self._init_plot(rows, cols, plot_scale)
```
Conchylicultor: Only the drawing part seems to be different between the two class, The loop seems to be the same, so this is what should have been factorised out too. Maybe something like:

```python
def _make_fig(draw_fn):
  ...
  for i, ex in enumerate(examples):
    ...
    draw_fn(fig, ...)
```

VaranRohila:
@Conchylicultor Some structural changes in the code. The `ImageVisulaizer` now has a build function which acts as a binder for two functions, `_infer_object_keys()` and `_plot_objects()` which should be defined in the child classes.  

In the `_infer_object_keys()` we retrieve keys of the objects we need to plot on a single image like for `SupervisedVisualizer` we retrieve label keys and `ObjectVisualizer` we retrieve the bboxes and the labels keys and we pass them to the `_plot_objects()` function in the child classes.

In the `_plot_objects()` method of the child classes, we plot the objects on a single example. We plot labels for `SupervisedVisualizer`; labels and objects for `ObjectVisualizer`.

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,235 @@
...
+
+  def _plot_objects(self, plt, ax, ds_info, example, image_key, label_key):
+    """Plots objects for a single example"""
+    # Plot the label
+    if label_key:
+      label = example[label_key]
+      label_str = ds_info.features[label_key].int2str(label)
+      plt.xlabel("{} ({})".format(label_str, label))
+
+    return plt, ax
```
Conchylicultor: Why return plt here ? this doesn't seem consistent with `_plot_image`

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,235 @@
...
+
+    num_examples = rows * cols
+    examples = list(dataset_utils.as_numpy(ds.take(num_examples)))
+
+    plt = lazy_imports_lib.lazy_imports.matplotlib.pyplot
+    fig = plt.figure(figsize=(plot_scale*cols, plot_scale*rows))
+    fig.subplots_adjust(hspace=1/plot_scale, wspace=1/plot_scale)
+
+    for i, ex in enumerate(examples):
+      ax = self._plot_image(plt, fig, ex, rows, cols, i, image_key)
```
Conchylicultor: You could have something like:

```python
  keys = self._infer_keys(ds_info, **key_names)
  ...

  for i, ex in enumerate(examples):
    ax = self._plot(plt, fig, ex, rows, cols, i, keys)
```

And then use super in the child class ?

```python
def _plot(self, plt, ...):
  # Plot the grid
  super(ObjectVisulaizer, self)._plot(plt, ...)
  # Update the grid with bounding boxes
  plt.xyz
```

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,235 @@
...
+        if isinstance(feature, feature_type)
+    ]
+
+  return [
+      k for k, feature in ds_info.features.items()
+      if isinstance(feature, feature_type)
+  ]
+
+def _draw_text(ax, xy, txt, size=14):
+  """Draws text on the top-left corner of the bounding box"""
```
Conchylicultor: Please add docstring explaining what each arg does. (with shape/dtype)

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,235 @@
+# coding=utf-8
```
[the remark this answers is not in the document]
Conchylicultor: Actually, maybe let's not do this in this PR

tensorflow_datasets/core/image_visualizer.py
Conchylicultor: As a general rules, when the number of argument becomes too big, it's best to use keywords arguments:
```
_plot_image(
    plt=plt,
    ex=ex,
    ...
)
```
This make the code less error prone an more maintainable

tensorflow_datasets/core/image_visualizer.py
```diff
@@ -0,0 +1,235 @@
...
+    image_keys = _extract_keys(ds_info, features_lib.Image)
+
+    if not image_keys:
+      return False
+
+    sequence_keys = _extract_keys(ds_info, features_lib.Sequence)
+
+    bbox_keys = []
+    for key in sequence_keys:
+      if isinstance(ds_info.features[key].feature, features_lib.FeaturesDict):
```
Conchylicultor: Rather than this loop, how about a generic

```python
sequence_key, bbox_key = _extract_nested_keys(
    ds_info, [features_lib.Sequence, features_lib.FeaturesDict, features_lib.BBoxFeature])
```
It would return a new key for each `FeatureDict`, including the top-level one.

tensorflow_datasets/core/visualization.py
```diff
@@ -26,25 +28,24 @@
...
   a plot of (rows*columns) images from a tf.data.Dataset.
-
   Usage:
-
   ```python
   ds, ds_info = tfds.load('cifar10', split='train', with_info=True)
   fig = tfds.show_examples(ds_info, ds)
   ```
-
   Args:
```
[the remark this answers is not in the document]
VaranRohila: I think there is some problem with my editor. I'll resolve it.

tensorflow_datasets/core/visualization.py
```diff
@@ -26,25 +28,24 @@
 from tensorflow_datasets.core import dataset_utils
 from tensorflow_datasets.core import features as features_lib
 from tensorflow_datasets.core import lazy_imports_lib
+from tensorflow_datasets.core.image_visualizer import ObjectVisulaizer, SupervisedVisualizer
```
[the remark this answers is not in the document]
VaranRohila: I did. I am importing classes from the `image_visualizer.py`

VaranRohila:
@Conchylicultor I have also overridden `build` function in the child classes so that more arguments can be added with respect to different child classes like `bbox_label` flag in `ObjectVisualizer`.

VaranRohila:
Also added `DummyObjectDetection` class in the `test_utils.py` for usage in the `visualization_test.py` file. Ran the test and it is working for both.

Conchylicultor:
I ended up refactoring myself the visualization to make it more extensible: https://github.com/tensorflow/datasets/pull/1728/files
Could you refactor the code to use the new class ?

VaranRohila:
Since I encountered some git errors, I had to force push. Everything is ok now/

PR #1642, opened 2020-03-13
Added pytype in tfds/core/lazy_imports_lib.py patch-9

### LINKED ISSUE #1517: [GSoC] Add Python 3 typing annotations to TFDS
**Task:** Add Python 3 typing annotations to TFDS repository. Now that Python 2 support is officially dropped, we can add better typing to TFDS.

**Impact:** Everyone using TFDS with a static type checker

**Instructions:**

* Add typing annotations to our API: (Typing annotation guide: https://docs.python.org/3/library/typing.html)
* Check that the typing annotations are valid with https://github.com/google/pytype
* Only update only a few files at a time (send multiple PR).
* As TFDS has many files, many people can work on this issue at the same time, just coordinate with each other.

**Complexity:** Easy

ashutosh1919:
Fixes #1517 .


tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -48,106 +52,106 @@ class LazyImporter(object):
 
   @utils.classproperty
   @classmethod
-  def apache_beam(cls):
+  def apache_beam(cls: Type[T]) -> object:
```
Conchylicultor: There shouldn't be annotations for `cls` or `self`, as the annotations are implicit.
For the returned object, it should be Module type, instead of object.

ashutosh1919: @Conchylicultor , Can you please give me a hint about how to return Module type using pytype annotations because I was not able to find it anywhere?

Conchylicultor: From the first google page: https://www.google.com/search?q=typing+module+python

 https://stackoverflow.com/questions/48389157/python-typing-for-module-type

ashutosh1919: done

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -52,106 +50,106 @@ class LazyImporter(object):
...
   @utils.classproperty
   @classmethod
-  def crepe(cls: Type[T]) -> object:
+  def crepe(cls) -> ModuleType("crepe"):
     return _try_import("crepe")
 
   @utils.classproperty
   @classmethod
-  def cv2(cls: Type[T]) -> object:
+  def cv2(cls) -> ModuleType("cv2"):
```
Conchylicultor, line 63: This should be `ModuleType`, not `ModuleType(xyz)`. `ModuleType` is the `type(module)` while `ModuleType(xyz)` is an instance of module.

Typing annotation indicates the type.

ashutosh1919: Thanks for clarification. I understand the concept now. I will change that.

PR #1661, opened 2020-03-17
script update

vinayvr11:
@Conchylicultor 
#Update random_image_generator.py Script

* [x] Lint code

### DIFF
--- a/tensorflow_datasets/scripts/random_image_generator.py
+++ b/tensorflow_datasets/scripts/random_image_generator.py
@@ -1,0 +1,123 @@
+
+""" Generate random images that compress better.
+    By replacing the larger images with more compressable equivalents. """
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+import os
+import sys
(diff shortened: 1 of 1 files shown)


vinayvr11:
> Thank you!

@Conchylicultor : one last help i know that i troubled you a lot so sorry for that but could you please help me with my GSOC proposal i have some queries. 
1) Can i select 2 or more projects like i selected in my proposal Tensorflow-dataset and tf.image.
2) In tensorflow-dataset project we only need to implement the new datasets is this the only work to do in tensorflow-datasets.
3) They just mention that in our project work period we just need to add more functionalities to tf.image but they won't mention that what are those features so that's why i only mention the updates and enhancement words in my proposal. Does they are right or i should need add something more.
Thank you

Conchylicultor:
TFDS and TF are two different team, I would strongly suggest you to only pick one.
Also from a TFDS point of view, we currently have more datasets to review than we can manage it, so instead GSoC should focus on core features.

PR #1669, opened 2020-03-18
Better compression of .tar and .zip files

### LINKED ISSUE #1666: [GSoC] Better fake data compression script
**Context**: Since #1661, we have a script [replace_fake_images.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/replace_fake_images.py) which compress all images inside our fake data directory by replacing the random noise images by a uniform color image of the same shape/dtype.

**Goal**: However, when applied to our [fake directory](https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/testing/test_data/fake_examples), some files seems to increase size, instead of being smaller (For the list of modified files, see [#1634](https://github.com/tensorflow/datasets/pull/1634/files)).
For example: imagenet2012/ILSVRC2012_img_train.tar increase by +50KB.
Some JPEG images are also badly compressed by the script (e.g. diabetic_retinopathy_detection/sample/1_left.jpeg).

The goal of this feature request would be:
* Investigate why some archives/image size increase after being compressed.
* Find a way to better compress the images / archives (e.g. by playing with the compression parameters of PIL https://pillow.readthedocs.io/en/5.1.x/handbook/image-file-formats.html#jpeg)

Additionally, other improvement could be done in parallel by different persons:
* Update the script to also compress audio files
* Update the script to also compress video files.

Eshan-Agarwal:
Fixes #1666  @Conchylicultor  Please have a look. Below I mention changes I done in `replace_fake_images.py`

1. Add `compression=zipfile.ZIP_DEFLATED` in `rewrite_zip` function which re compresses files in efficient way than previous.

2. Add `'w:gz'` & `arcname=os.path.basename(temp_dir)` in `rewrite_tar` function 

It solves problem related to all type of `tar `and `zip `files.

### DIFF
--- a/tensorflow_datasets/scripts/replace_fake_images.py
+++ b/tensorflow_datasets/scripts/replace_fake_images.py
@@ -104,10 +104,9 @@ def rewrite_tar(root_dir, tar_filepath):
     # Checking the extension of file to be extract
     if tar_filepath.lower().endswith('gz'):
       extension = ':gz'
-    elif tar_filepath.lower().endswith('bz2'):
+    if tar_filepath.lower().endswith('bz2'):
       extension = ':bz2'
-    else:
-      extension = ''
+    extension = ''
 
     # Extraction of .tar file
     with tarfile.open(tar_filepath, 'r' + extension) as tar:
@@ -116,8 +115,8 @@ def rewrite_tar(root_dir, tar_filepath):
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
 
     # Converting into tarfile again to decrease the space taken by the file-
-    with tarfile.open(tar_filepath, 'w' + extension) as tar:
-      tar.add(temp_dir, recursive=True)
+    with tarfile.open(tar_filepath, 'w:gz') as tar:
+      tar.add(temp_dir, recursive=True,arcname=os.path.basename(temp_dir))
 
 
 def rewrite_dir(fake_dir):
@@ -134,11 +133,11 @@ def rewrite_dir(fake_dir):
       img_ext = os.path.basename(file).split('.')[-1].lower()
       if img_ext in img_ext_list:
         rewrite_image(path)
-      elif path.endswith('.npz'):  # Filter `.npz` files
+      if path.endswith('.npz'):  # Filter `.npz` files
         continue
-      elif zipfile.is_zipfile(path):
+      if zipfile.is_zipfile(path):
         rewrite_zip(root_dir, path)
-      elif tarfile.is_tarfile(path):
+      if tarfile.is_tarfile(path):
         rewrite_tar(root_dir, path)
 
 
--- a/tensorflow_datasets/scripts/replace_fake_images.py
+++ b/tensorflow_datasets/scripts/replace_fake_images.py
@@ -82,7 +82,7 @@ def rewrite_zip(root_dir, zip_filepath):
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
 
     # Compressed the .zip file again
-    with zipfile.ZipFile(zip_filepath, 'w') as zip_file:
+    with zipfile.ZipFile(zip_filepath, 'w',compression=zipfile.ZIP_DEFLATED) as zip_file:
       for file_dir, _, files in os.walk(temp_dir):
         for file in files:
           file_path = os.path.join(file_dir, file)
--- a/tensorflow_datasets/scripts/replace_fake_images.py
+++ b/tensorflow_datasets/scripts/replace_fake_images.py
@@ -52,7 +52,7 @@ def rewrite_image(filepath):
   image = np.array(PIL.Image.open(filepath))
 
   # Filter unsuported images
-  if image.mode == 'RGBA' or image.dtype == np.bool:
+  if PIL.Image.open(filepath).mode == 'RGBA' or image.dtype == np.bool:
     return
 
   # The color is a deterministic function of the relative filepath.


tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -52,7 +52,7 @@ def rewrite_image(filepath):
   image = np.array(PIL.Image.open(filepath))
 
   # Filter unsuported images
-  if image.mode == 'RGBA' or image.dtype == np.bool:
+  if PIL.Image.open(filepath).mode == 'RGBA' or image.dtype == np.bool:
```
Conchylicultor: You should re-use the pil image open above, rather than re-opening the image twice.

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -82,7 +82,7 @@ def rewrite_zip(root_dir, zip_filepath):
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
 
     # Compressed the .zip file again
-    with zipfile.ZipFile(zip_filepath, 'w') as zip_file:
+    with zipfile.ZipFile(zip_filepath, 'w',compression=zipfile.ZIP_DEFLATED) as zip_file:
```
Conchylicultor: Please format your file using https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -104,10 +104,9 @@ def rewrite_tar(root_dir, tar_filepath):
     # Checking the extension of file to be extract
     if tar_filepath.lower().endswith('gz'):
       extension = ':gz'
-    elif tar_filepath.lower().endswith('bz2'):
+    if tar_filepath.lower().endswith('bz2'):
```
Conchylicultor: Why change this ?

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -134,11 +133,11 @@ def rewrite_dir(fake_dir):
       img_ext = os.path.basename(file).split('.')[-1].lower()
       if img_ext in img_ext_list:
         rewrite_image(path)
-      elif path.endswith('.npz'):  # Filter `.npz` files
+      if path.endswith('.npz'):  # Filter `.npz` files
```
Conchylicultor: Why change this ?

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -116,8 +115,8 @@ def rewrite_tar(root_dir, tar_filepath):
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
 
     # Converting into tarfile again to decrease the space taken by the file-
-    with tarfile.open(tar_filepath, 'w' + extension) as tar:
-      tar.add(temp_dir, recursive=True)
+    with tarfile.open(tar_filepath, 'w:gz') as tar:
```
Conchylicultor: I don't think we should compress all tar files using gz. `.tar.gz` are compressed with `.gz`, but not `.tar` files.

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -116,8 +115,8 @@ def rewrite_tar(root_dir, tar_filepath):
...
+      tar.add(temp_dir, recursive=True,arcname=os.path.basename(temp_dir))
```
Conchylicultor: Can you add a comment explaining why `arcname` is required here in this context, as reader might not be familiar with it. (e.g. give an example of filepath)

tensorflow_datasets/scripts/replace_fake_images.py
Conchylicultor: Can you add a comment explaining how the compression method was choose ? `compression=zipfile.ZIP_DEFLATED`

tensorflow_datasets/scripts/replace_fake_images.py
Eshan-Agarwal: Actually compressing files with `.gz` compress files better, if we don't use it here it ends up in large files as previous. But I provided `arcname` so that final output file is of format `.tar`.

I tried out many things but it works finally is it okay to use this method ?

tensorflow_datasets/scripts/replace_fake_images.py
Eshan-Agarwal: I will change it as per previous just forgot to correct it after running locally

tensorflow_datasets/scripts/replace_fake_images.py
Eshan-Agarwal: Yeah sure its good for reader also, as I explains above why I use `arcname `here

tensorflow_datasets/scripts/replace_fake_images.py
Eshan-Agarwal: @Conchylicultor  is it ok if I compress files using .gz and saving it as .tar final file was saving as .tar not .gz

Conchylicultor: I think this may creates bug when the program expect tar while it would actually receive targz compressed files.

Conchylicultor: Is there a way to detect if the original `.tar` archive file was compressed with `gz` or not ? If so, we should apply the same compression as the original archive.

Eshan-Agarwal: @Conchylicultor  Yeah sure I fill look that, but I after compressing files with .gz I tried to decompress it using `tar -xzvf archive.tar` and it gives decompresses them successfully as original one, Is it make any sense ?

Conchylicultor:
Closing this as the other PR has been merged

PR #1672, opened 2020-03-18
Fixed zip/tar file sizes and image compression issues in replace_fake_images.py

MadhavEsDios:
In lieu of Issue No. 1666, I have made changes to the replace_fake_images.py script to fix the issues of
1) Increase in zip/ tar file size despite the contents being smaller in size. 
For this issue, I fix the directory structure which previously used the absolute path.
Additionally for tar files, the end compression objective is now a gztar.
2) Certain images not being compressed properly by PIL leading to a dramatic increase in file-size.
For this, I use the optimize functionality provided by PIL.Image.save which eliminates unused colors.


MadhavEsDios:
@googlebot I signed it!

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -114,10 +112,10 @@ def rewrite_tar(root_dir, tar_filepath):
...
 
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
-
-    # Converting into tarfile again to decrease the space taken by the file-
-    with tarfile.open(tar_filepath, 'w' + extension) as tar:
-      tar.add(temp_dir, recursive=True)
+    
+    os.remove(tar_filepath) # Remove original tar file as dupicates with the extension tar.gz will be created
+    
+    shutil.make_archive(tar_filepath[:-4], 'gztar', temp_dir) # Copies the directory structure of the extracted tar file and stores it back gztar
```
Conchylicultor: `[:-4]` won't work with `.tar.gz`, `.tgz` files.
Also, we should keep the same compression than the original archive. For instance, `.tgz` should use `gz` compression, but I'm not sure `.tar` should

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -80,14 +83,9 @@ def rewrite_zip(root_dir, zip_filepath):
...
-
-    # Compressed the .zip file again
-    with zipfile.ZipFile(zip_filepath, 'w') as zip_file:
-      for file_dir, _, files in os.walk(temp_dir):
-        for file in files:
-          file_path = os.path.join(file_dir, file)
-          zip_file.write(file_path)
-
+    
+    shutil.make_archive(zip_filepath[:-4], 'zip', temp_dir) #Copies the directory structure of the extracted file ans stores it back as zip
```
Conchylicultor: I think we should keep using the low level Zip file API as it allow to access more fine grain compression options in the zipfile API.
Is it possible to solve the absolute path bug while keeping the low level API ?

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -63,7 +66,7 @@ def rewrite_image(filepath):
 
   image = np.ones_like(image) * color
   image = PIL.Image.fromarray(image)
-  image.save(filepath)
+  image.save(filepath, optimize=True)
```
Eshan-Agarwal: @MadhavEsDios  yes you are right settings optimize will help but when I tried it I observed it decreases size of .jpeg images only but increases size for png format.

tensorflow_datasets/scripts/replace_fake_images.py
MadhavEsDios: 1. Sure, I will update the PR with the fake_data dir result!
2. The only benefit of using the Zipfile API which shutil uses as well is that you would have access to the Zip.deflated option.
3. Thank you for pointing out the problem with using [:-4]. My bad ! I will make the correction.
4. I understand that the reason for the fake_data dir is to replicate the original and thus it makes sense to not convert from tar to tar.gz. Will make that change as well.

tensorflow_datasets/scripts/replace_fake_images.py
MadhavEsDios: Thank you for pointing that out I will check file-sizes and confirm what you have observed.

Eshan-Agarwal: Thank you sure

tensorflow_datasets/scripts/replace_fake_images.py
MadhavEsDios: Hey @Conchylicultor, sorry I did not reply to your last comment. Sure, I think it can be done, the only problem here is that if I set the arcname to the filename, directory information would be lost. However, in my opinion, nothing some lengthy 'os' operations can't solve.
The low-level zipfile API can be used, albeit with a loss in readability and not too great of an advantage as far as I can tell. (Maybe you have some inputs regarding this)

MadhavEsDios:
@Conchylicultor 
Please refer to the latest version where I have changed the hardcoded file extension removal and also have posted the output of the fake_example dir at the same location.

tensorflow_datasets/scripts/replace_fake_images.py
MadhavEsDios: Hey, I checked a couple of file-sizes and it looked OK, there was no increase in size for the png format. If you still have concerns, please refer to the "fake_examples" dir in my latest commit which contains the output of the modified script.
Perhaps, you find some anomalies which we can then address sequentially!

Conchylicultor:
Thank you for updating the files. It seems that even `.tar.gz` file size grow bigger, which isn't expected as it should be compressed with `.gz`. Could you try to investigate what is happening ?

For ex: BIN +3.6 MB tensorflow_datasets/testing/test_data/fake_examples/abstract_reasoning/attr.tar.gz

MadhavEsDios:
Hey, thank you for reviewing the output-dir.  Thanks to you pointing out the abstract_reasonin dataset, I realized that using filename.split('.') would not work for filenames with multiple periods like attr.rel.pairs.tar.gz, which is why I have switched back to case-wise extension removal. Please refer to the recent commit which looks pretty OK.
As for the tar.gz file sizes, you are right it is very weird. For eg in abstract reasoning,
1.  Original file size: 3770530 bytes, output file size: 37771251 (attr.rel.pairs.tar.gz) - 2. Original file size: 3769957 bytes, output file size: 3770539 bytes (attrs.shape.color.tar.gz) - and so on
Similarly, in the imdb_reviews dataset,
1. Original file size: 541 bytes, output file size: 642 bytes (acllmdb_v1.tar.gz)
If you notice both these examples you would find that they contain **.npz** and **.txt** files respectively. Both of which are not affected by the process of decompression and compression.
Whereas, if you look at the caltech_birds2010 dataset, 
1. (Images.tar.gz) Original file size: 351KB, Output file size: 1KB
In this case, the tar.gz consists of images because of which there is a drastic reduction in size ( which is expected behavior)

I will investigate why this happens as I have no intuitive explanations!
From this small sample set, it seems that when internal file-sizes are same (*npz, *txt) the  compressed file size increases whereas when images (which are modified by the script) are the internal files, compressed file size reduces drastically (expected behaviour).

MadhavEsDios:
Update: I just tried a simple experiment, I decompressed files from fake_examples/abstract_reasoning  manually on my GUI filesystem and then used the tar -czf command to recompress them back. 
I found that this leads to a significant increase in size (from 3.8 MB to 4.2 MB), unlike 'shutil' which only results in a marginal increase.
Maybe this is a 'tar' compression problem and not a python specific one? (Thoughts on this)

tensorflow_datasets/scripts/replace_fake_images.py
Eshan-Agarwal: Please have a look for these files : `tensorflow_datasets/testing/test_data/fake_examples/curated_breast_imaging_ddsm/1.3.6.1.4.1.9590.100.1.2.0004618681.0223128662/1.3.6.1.4.1.9590.100.1.2.0824689759.0062556928`.  Some image sizes increases from 107 to 116 which is small but still increasing in size.

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -104,20 +101,24 @@ def rewrite_tar(root_dir, tar_filepath):
     # Checking the extension of file to be extract
     if tar_filepath.lower().endswith('gz'):
       extension = ':gz'
+      shutil_extension = 'gztar'
+      extension_removal_index = 7 # Hardcode number of indices to be skipped from the end to obtain filename. This is hard-coded as filenames containing "." might exist and thus filepath.split(".") would give an incorrect result.
```
Conchylicultor: Rather than hardcoding a number, I think it would be more explicit to use `len('.tar.gz')`, `len('.zip')`,...

Conchylicultor: This won't works with `.tgz` files. Even if we may not have now, we may in the future. Is there a way to specify the extension in `make_archive` ?

tensorflow_datasets/scripts/replace_fake_images.py
Conchylicultor: I think increase smaller than 1KB are fine.

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -36,6 +36,9 @@
 import absl.flags
 import numpy as np
 import PIL.Image
+import shutil # New import
```
Conchylicultor: Remove `# New import`

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -80,14 +83,8 @@ def rewrite_zip(root_dir, zip_filepath):
...
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
-
-    # Compressed the .zip file again
-    with zipfile.ZipFile(zip_filepath, 'w') as zip_file:
-      for file_dir, _, files in os.walk(temp_dir):
-        for file in files:
-          file_path = os.path.join(file_dir, file)
-          zip_file.write(file_path)
-
+    shutil.make_archive(zip_filepath[:-4], 'zip', temp_dir) #Copies the directory structure of the extracted file ans stores it back as zip
```
Conchylicultor: `zip_filepath[:len(.zip)]`, also I would add a check to make sure the file ends by `.zip`

tensorflow_datasets/scripts/replace_fake_images.py
MadhavEsDios: Unfortunately, there is no way to specify the target extension to make_archive.
There are 5 options: 
1. 'zip' - .zip extension
2.'tar' - .tar extension
3.'gztar' - .tar.gz extension
4.'bztar' - .tar.bz2 extension
5.'xztar' - .tar.xz extension

As you already know .tgz is equivalent to .tar.gz. So will renaming a file from .tgz to tar.gz be a major issue in the future?

tensorflow_datasets/scripts/replace_fake_images.py
MadhavEsDios: In rewrite_dir the zipfile.is_zipfile(path) check is done. To the best of my knowledge, I know only .zip extensions which would be allowed by this filter. Do you know of other '.zip' extensions which come to mind that might create issues ?
In any case, in the latest commit, I perform a possibly redundant check just as you mentioned above.

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -83,8 +81,8 @@ def rewrite_zip(root_dir, zip_filepath):
       zip_file.extractall(path=temp_dir)
 
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
-    shutil.make_archive(zip_filepath[:-4], 'zip', temp_dir) #Copies the directory structure of the extracted file ans stores it back as zip
-    # The extraction target in the above line can be changed to gztar for additional compression 
+    if zip_filepath.lower().endswith('zip'):
+        shutil.make_archive(zip_filepath[:-len('.zip')], 'zip', temp_dir) #Copies the directory structure of the extracted file ans stores it back as zip
```
Eshan-Agarwal: @MadhavEsDios It's look great, after all things please make sure to check code style as described [here](https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style). 
Also is it solve problem of .tar files like in `Imagenet2012.tar` file, also with remaining issues?

MadhavEsDios: Thanks for pointing out the code style, I have fixed it.
No, it does not solve the problem of .tar files (at least for Imagenet2012). As I have written in a previous issue before, this does not seem to be a python tarfile problem, but more of a 'tar' compression format issue. This is because the 'tar' format usually introduces some overhead for hashing internal files.
As mentioned above, I am not entirely sure whether this is the cause, so will still investigate it.
This script fixes .zip file issues, image compression, and absolute path issues along with support for .tgz and .xz files.

Eshan-Agarwal: Thanks for your quick response, please run lint.sh in oss_script as descibed in link i given above as I seen you not run it properly like one  line in code should not be more than 80 words. Can you please describe how use of shutil is advantageous than other low level zip and tar as low level also works  ? What is problem associated with absolute path as i not understand it ?

MadhavEsDios:
Hey @Conchylicultor, please review the final version. 
Also, the 'tar' file size issue has started to make more sense. I found out the '.tar' merely creates an archive and does not additionally compress internal files. [https://unix.stackexchange.com/questions/127169/does-tar-actually-compress-files-or-just-group-them-together](url)
Thus, it does not seem too weird that there is a slight discrepancy in file sizes, especially when we take 'tar' archiving overhead into account.

tensorflow_datasets/scripts/replace_fake_images.py
MadhavEsDios: 1. Running lint, hi I ran the lint command on the only file I made changes to i.e. replace_fake_images.
`./oss_scripts/lint.sh tensorflow_datasets/scripts/replace_fake_images.py 

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)`
As you can see it gives a score of 10. I have not run it on the whole directory, as I do not want to affect other scripts.So please verify if the warnings that you see are from this particular script. Thanks !

2. Advantages of shutil: I have also mentioned this as a comment below the issue.
If you ran the previous script, the generated zip files contained an unnecessary directory structure like follows:
Let's say the original file was abc.zip and the temp file was temp1, the previous output instead of being abc.zip was '/Users/Name/...../testing/test_data/fake_examples/temp1/abc.zip' instead of just abc.zip. As this was done recursively, lets say this abc.zip file contained another xyz.zip file, this too would use the absolute path instead of just xyz.zip. This issue increased the compressed file size as it contains unnecessary information.
Shutil.make_archive solves this problem by internally using os.chg_dirs and gives the desired output. So the low-level API in the older script was not working as intended.
However, as I have mentioned in a previous issue, modifications to the low-level API can be made by using os library functions to get the desired output. However, using shutil.make_archive is cleaner and has no major disadvantage.

Eshan-Agarwal: Okay it seems good but adding folders in folders not increasing size, still if get off from this problem using shutil thats good.
Maybe it increases some time complexity ? Than recursive approach i am not sure about that still good work. 
Thanks

MadhavEsDios: Adding folders within folders did increase the size I think but only for 'zip' files (For eg. cats_vs_dogs). Which is why I guess shutil resolved this issue.
Also, you have rightly pointed out (i have written this in my comment on the issue as well) that using shutil and PIL optimize=True takes longer to execute than the previous version. This is expected behavior.

Thanks for your contributions.

Conchylicultor:
Looking at the tests for `tf2`, it seems that the fake data are modified and make some tests crashed. For
* LibriTTSTest: `KeyError: "filename 'LibriTTS/speakers.tsv' not found"`
* `CelebAHQTest`: `ValueError: fake_examples dir /tmpfs/src/github/datasets/tensorflow_datasets/testing/test_data/fake_examples/celeb_a_hq not found.`
* ...

Could you update the script to make sure those tests still pass ?

MadhavEsDios:
@Conchylicultor 
Thanks for providing the stack-trace.

Note: The following issue is only for tar-files and not zip. So we can keep using shutil.make_archive() for zip files without any issues.

1. Unfortunately after debugging I found that this seems to be a shutil.make_archive() issue.
So for tar-files, shutil.make_archive() add an additional './' for no good reason. 
For eg. : "123.JPEG" becomes "./123.JPEG". Also make_archive() adds an additional "." directory for no good reason. It is because of this issue that the files for both LibriTTSTest and Imagenet2012Test were not found.
There is already a simple fix for this issue: [https://github.com/python/cpython/pull/13738#event-3149177972](url), however unfortunately it has not been merged yet.
I verified that this is the reason by modifying both LibriTTSTest,  Imagenet2012Test and managed to pass both tests without any errors.

2. The CelebAHQTest failure was because of an incorrect copy on my system, I will fix it in the next commit. (It is not related to the above issue)

I am currently working on a fix for issue no. 1 by switching back to and modifying the low-level API for tar-files.
Sorry about the waste of time !

tensorflow_datasets/scripts/replace_fake_images.py
Conchylicultor: > As you already know .tgz is equivalent to .tar.gz. So will renaming a file from .tgz to tar.gz be a major issue in the future?

It would be. We don't control the dataset filenames. For example, if imagenet use `train.tgz`, then the fake data should match this name too.

MadhavEsDios: I have tested this with the latest commit, .tgz files will retain their extension without changing to .tar.gz. The same is true for other tar compression variants as well. So marking this issue as resolved.

MadhavEsDios:
@Conchylicultor 
Sorry about the delay, I did not get the time ! 
So I have fixed all the issues faced before.
Have tested for all 3 above datasets and many other random ones. All tests have passed.

For the data-path issues, like I tagged the link in the previous comment, the problem was with the tarfile.add() method as along with files it also keeps adding directories leading to failure in the tests. 
To solve this issue, I have created an additional method called 
`tar_add` which is similar to the source code of tarfile.add() and performs the same function but with a few differences.
The new method is as belows:
```python3
def tar_add(tar_file, name, arcname=None, recursive=True):
  """Perform addition of files to tarfile.
  This method replicates tarfile.add() and solves the issue of
  unnecessary filepaths like ., ./ and ./folder/.. being added to tarinfo.

  Args:
    tarfile: pass open tarfile object to which files need to be added
    name: Add the file `name' to the archive.
    `name' may be any type of file (directory, fifo, symbolic link, etc.)
    arcname: If given, `arcname' specifies an
    alternative name for the file in the archive.
    recursive: Option to specify if directories
    need to be added recursively.
    Directories are added recursively by default.
  """
  if arcname is None:
    arcname = name
  tarinfo = tar_file.gettarinfo(name, arcname)
  if tarinfo.isreg():
    with bltn_open(name, "rb") as f:
      # Filewise addition into tarball
      tar_file.addfile(tarinfo, f)
  elif tarinfo.isdir():
    if tarinfo.isfile(): # Only files should be added to tarinfo and not directories
      tar_file.addfile(tarinfo)
    if recursive:
      for f in sorted(os.listdir(name)):
        if name == os.curdir: # Remove redundant current directory address ',' from name
          name = ''
        arcname = name # Since arcname=None always, any changes to any changes to name should affect arcname
        tar_add(tar_file, os.path.join(name, f), os.path.join(arcname, f),
                recursive) # Recursive call to add all existing files to tarball
```

tensorflow_datasets/scripts/replace_fake_images.py
```diff
@@ -115,9 +117,43 @@ def rewrite_tar(root_dir, tar_filepath):
...
     rewrite_dir(temp_dir)  # Recursivelly compress the archive content
 
-    # Converting into tarfile again to decrease the space taken by the file-
+    os.chdir(temp_dir) # Change to temporary directory so that temp directory name does not affect original filenames
     with tarfile.open(tar_filepath, 'w' + extension) as tar:
-      tar.add(temp_dir, recursive=True)
+      tar_add(tar, '.', recursive=True) # Call custom recursive tar add operation
+
+
+def tar_add(tar_file, name, arcname=None, recursive=True):
```
Conchylicultor: I am confused by this function.
* Why use `bltn_open` instead of `open` ?
* Why have `recursive` arg if this is always True ?
* The `os.chdir(` seems like a hack and should not be used. Instead, why not simply truncate the name ?
* Why checking `isfile()` inside the `isdir()` scope ?

I don't understand why you need this complicated function.
Can't you just use something like ?

```python
with tarfile.open(tar_filepath, 'w' + extension) as tar:
  tar.tar_file(tmp_dir, arcname='', recursive=True)
```
Or if the above function do not work:
```python
with tarfile.open(tar_filepath, 'w' + extension) as tar:
  with filename in os.path.listdir(tmp_dir):
    tar.tar_file(os.path.join(tmp_dir, filename), arcname=filename, recursive=True)
```

MadhavEsDios: Hey you are absolutely right, there is no need for this complication. I was under the wrong impression that tarfile.getnames() should only return filenames and not directories.(Consequence of overthinking). 
The code is complicated because it is from [https://github.com/python/cpython/blob/master/Lib/tarfile.py](url). (The python library tarfile.py)
You are right about both, os.chdir is hacky and recursive=True is not required.
My sincerest apologies for wasting your time.

Your 2nd suggestion (which I had obviously tried and rejected foolishly as the tarfile still included directories)
```python3
    with tarfile.open(tar_filepath, 'w' + extension) as tar:
      for filename in os.listdir(temp_dir):
        tar.add(os.path.join(temp_dir, filename), arcname=filename,
                recursive=True)
```
works perfectly fine. I have run tests on random datasets from images, audio and text. All test pass. I have made this change along with the new fake_data_dir output in the new commit.
Also no extra code needs to be written for '.tgz' files as the tarfile library takes care of the extension. ( i have tested the latest commit on this as well)

Conchylicultor: Thank you for the update. Why haven't you tried the first suggestion ?
```
with tarfile.open(tar_filepath, 'w' + extension) as tar:
  tar.tar_file(tmp_dir, arcname='', recursive=True)
```

This seems simpler, no ? Was there an issue with this one ?

MadhavEsDios: My apologies, I don't know why I am embarrassing myself !
Yes the first solution is way simpler and easier to read,  also solves the problem of "./" being added to each component file.
I have tried your first suggestion and all random tests pass. Also tested '.tgz' extension check which passes as well with the latest commit.

Again I apologise for wasting precious time. I think I should calm down with the commits.

MadhavEsDios: Also a question, do you think using shutil.make_archive for zip files and low-level API for tar files is OK ? 
Something similar with zipfile library can also be done, however unlike tarfile there is no recursive write option. Should I try to eliminate shutil for zipfiles ?

MadhavEsDios: So for example this piece of code (modified slightly from the previous version):
```python3
    if zip_filepath.lower().endswith('zip'): # new line
      with zipfile.ZipFile(zip_filepath, 'w', compression=zipfile.ZIP_DEFLATED) as zip_file:
        for file_dir, _, files in os.walk(temp_dir):
          for file in files:
            file_path = os.path.join(file_dir, file)
            arcname=os.path.relpath(file_path, temp_dir) # new line
            zip_file.write(file_path, arcname=arcname)
```
does the same job as shutil.make_archive(). I have tested for this version as well and all tests for datasets with zip file PASS. Should I switch this back as well ?

MadhavEsDios:
Latest commit has all requested changes and also removes usage of shutil.make_archive().

PR #1693, opened 2020-03-23
Print error message in registered.py for datasets not updated in pip package.

Eshan-Agarwal:
@Conchylicultor  please have a look on following changes for better printing of error messages

**Is your feature request related to a problem? Please describe.**
Whenever someone load data using `tfds.load("dataset_name)`, using pip package of tfds instead from cloning repo, it show error for some datasets which was not been updated into pip package of tfds. Datasets like `robonet.py` & `waymo_open_dataset`.

**Describe the solution you'd like**
Best solution use `tfds-nightly`

**Describe alternatives you've considered**
We can print extra check [message](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/registered.py#L81) that display, data is not been updated in current pip package so use `tfds-nightly` or clone tfds repo

**Additional context**
Printing message helps user to understand why dataset he wants to load is not loaded.

### DIFF
--- a/tensorflow_datasets/core/registered.py
+++ b/tensorflow_datasets/core/registered.py
@@ -84,6 +84,7 @@ Check that:
     - dataset class is not in development, i.e. if IN_DEVELOPMENT=True
     - the module defining the dataset class is imported
     - dataset you want to use is updated in current pip package that you are using.
+    - Try tfds-nightly using pip install tfds-nightly.
 """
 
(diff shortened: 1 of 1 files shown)


Conchylicultor:
Good idea. Sending a fix now.

Eshan-Agarwal:
Thanks for you support

Issue #1223, opened 2019-11-23
Better doc to define datasets outside TFDS


olk:
Documentation is incomplete  see https://github.com/tensorflow/datasets/issues/704#issuecomment-557826887.

Eshan-Agarwal:
Hey! Can you please provide further information regarding this ?

Conchylicultor:
Don't hesitate to send a PR to update the missing info. From the comment:

> You have to document the naming format of the file that holds the checksums (== name of the dataset class in lower cases with under scores) and pass DownloadConfig(register_checksums=True) to DownloadManager.download_and_prepare()

Eshan-Agarwal:
@Conchylicultor In this issue only doc is needed to be updated, like  before defining `tfds.download.add_checksums_dir('/path/to/checksums_dir') `  we can add like name of checksum is in small letters with  under scores. and in script `tensorflow_datasets.scripts.download_and_prepare` add register_checksums=True.

Is I am interpret it correct ?

vijayphoenix:
Outdated issue.
See https://github.com/tensorflow/datasets/issues/2418

PR #1702, opened 2020-03-23
Improve define the dataset outside TFDS doc

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/1223


vijayphoenix:
See #2531

PR #1715, opened 2020-03-24
TFDS fails on older versions

### LINKED ISSUE #1708: [GSoC] TFDS should fail when user try to generate an old version
TFDS only support generation of dataset for the last versions.

We should make sure to raise an error here if the user try to generate an older version. The `download_and_prepare()` function should be updated here to raise an error: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/dataset_builder.py#L290

A version can only be installed if the version is the `self.canonical_version` or the `max(self.versions)`

For example:
```
VERSION = tfds.core.Version('1.0.0')  # Can be generated
SUPPORTED_VERSION = [
    tfds.core.Version('2.1.0'),  # Can be generated
    tfds.core.Version('2.0.0'),  # Cannot be generated
    tfds.core.Version('0.1.0'),  # Cannot be generated
]
```

Eshan-Agarwal:
Fixes #1708 
@Conchylicultor  please have a look. when I run dataset_builder_test.py it shows me error 

> Traceback (most recent call last):
  File "dataset_builder_test.py", line 323, in test_non_preparable_version
  File "C:\Users\eshan\Anaconda3\envs\keras-gpu\lib\contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "C:\Users\eshan\Anaconda3\envs\keras-gpu\lib\site-packages\tensorflow_core\python\framework\test_util.py", line 2755, in assertRaisesWithPredicateMatch
    (str(type(e)), str(e)))
AssertionError: Exception of type <class 'AssertionError'>: The version of the dataset you are trying to use (dummy_dataset_shared_generator:0.0.7) can only be generated using TFDS code synced @ v1.0.0 or earlier. Either sync to that version of TFDS to first prepare the data or use another version of the dataset (available for `download_and_prepare`: 1.0.0, 2.0.0).

This is because version 0.0.7 is not canonical nor maximum version for dummy_dataset_shared_generator

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -287,7 +287,7 @@ class DatasetBuilder(object):
       return
 
     if self.version.tfds_version_to_prepare:
-      available_to_prepare = ", ".join(str(v) for v in [self.canonical_version] + [max(self.versions)]
+      available_to_prepare = ", ".join(str(v) for v in self.versions
                                        if not v.tfds_version_to_prepare)
       raise AssertionError(
           "The version of the dataset you are trying to use ({}:{}) can only "
@@ -298,6 +298,27 @@ class DatasetBuilder(object):
               self.name, self.version, self.version.tfds_version_to_prepare,
               available_to_prepare))
 
+    # Check if requested version is canonical version or max(self.versions)
+    # if not raises error as tfds not supports older versions of datasets.
+    new_versions = [str(v) for v in [self.canonical_version] + [max(self.versions)]]
+    if str(self._version) not in new_versions:
+      # if canonical and maximum versions are same then sugeest only one version
+      if max(self.versions) != self.canonical_version:
+        raise AssertionError(
+            "The version of the dataset you are trying to use ({}:{}) cannot"
+            " be generated as this version is too older version to use. Please "
+            " use another new availble version of the dataset"
+            " (available for `download_and_prepare`: "
+            "{}).".format(
+                self.name, self.version, new_versions))
+      else:
+        raise AssertionError(
+            "The version of the dataset you are trying to use ({}:{}) cannot"
+            " be generated as this version is too older version to use. Please "
+            " use another new availble version of the dataset"
+            " (available for `download_and_prepare`: "
+            "{}).".format(
+                self.name, self.version, new_versions[0]))
     # Currently it's not possible to overwrite the data because it would
     # conflict with versioning: If the last version has already been generated,
     # it will always be reloaded and data_dir will be set at construction.
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -287,7 +287,7 @@ class DatasetBuilder(object):
       return
 
     if self.version.tfds_version_to_prepare:
-      available_to_prepare = ", ".join(str(v) for v in self.versions
+      available_to_prepare = ", ".join(str(v) for v in [self.canonical_version] + [max(self.versions)]
                                        if not v.tfds_version_to_prepare)
       raise AssertionError(
           "The version of the dataset you are trying to use ({}:{}) can only "
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -286,39 +286,19 @@ class DatasetBuilder(object):
       logging.info("Reusing dataset %s (%s)", self.name, self._data_dir)
       return
 
-    if self.version.tfds_version_to_prepare:
-      available_to_prepare = ", ".join(str(v) for v in self.versions
-                                       if not v.tfds_version_to_prepare)
-      raise AssertionError(
-          "The version of the dataset you are trying to use ({}:{}) can only "
-          "be generated using TFDS code synced @ {} or earlier. Either sync to "
-          "that version of TFDS to first prepare the data or use another "
-          "version of the dataset (available for `download_and_prepare`: "
-          "{}).".format(
-              self.name, self.version, self.version.tfds_version_to_prepare,
-              available_to_prepare))
-
     # Check if requested version is canonical version or max(self.versions)
+    # Also check for tfds_version_to_prepare.
     # if not raises error as tfds not supports older versions of datasets.
-    new_versions = [str(v) for v in [self.canonical_version] + [max(self.versions)]]
-    if str(self._version) not in new_versions:
-      # if canonical and maximum versions are same then suggest only one version
-      if max(self.versions) != self.canonical_version:
-        raise AssertionError(
-            "The version of the dataset you are trying to use ({}:{}) cannot"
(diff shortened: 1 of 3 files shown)


tensorflow_datasets/core/dataset_builder.py
```diff
@@ -287,7 +287,7 @@ def download_and_prepare(self, download_dir=None, download_config=None):
       return
 
     if self.version.tfds_version_to_prepare:
-      available_to_prepare = ", ".join(str(v) for v in self.versions
+      available_to_prepare = ", ".join(str(v) for v in [self.canonical_version] + [max(self.versions)]
```
Conchylicultor: You should add a test. This won't raise an error in most cases as  most dataset will ever go in the `if self.version.tfds_version_to_prepare:` scope

Conchylicultor:
> when I run dataset_builder_test.py it shows me error

If tests fails, then you should either update your code to make the tests pass, or update the tests if you think the current tests are wrong.

tensorflow_datasets/core/dataset_builder.py
Eshan-Agarwal: Ok so test should be in [this](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/dataset_builder_test.py#L107) class or I have to create class for `dataset_builder.download_and_prepare` ?

Eshan-Agarwal:
> > when I run dataset_builder_test.py it shows me error
> 
> If tests fails, then you should either update your code to make the tests pass, or update the tests if you think the current tests are wrong.

I think I have to update test because when change dataset_builder to run only for canonical and maximum version this test which I show above always fails

Eshan-Agarwal:
@Conchylicultor  please have a look and review the changes

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -298,6 +298,27 @@ def download_and_prepare(self, download_dir=None, download_config=None):
               self.name, self.version, self.version.tfds_version_to_prepare,
               available_to_prepare))
 
+    # Check if requested version is canonical version or max(self.versions)
+    # if not raises error as tfds not supports older versions of datasets.
+    new_versions = [str(v) for v in [self.canonical_version] + [max(self.versions)]]
+    if str(self._version) not in new_versions:
+      # if canonical and maximum versions are same then suggest only one version
+      if max(self.versions) != self.canonical_version:
+        raise AssertionError(
```
Conchylicultor: Rather than duplicating three times the error message, can't you refactor to only raise the exception once ?

tensorflow_datasets/core/dataset_builder_test.py
```diff
@@ -324,6 +324,18 @@ def test_non_preparable_version(self, *unused_mocks):
...
       builder.download_and_prepare()
+      
+  def test_with_old_versions(self):
+    expected = (
+        "The version of the dataset you are trying to use"
+        " (dummy_dataset_shared_generator:0.0.8) cannot be"
+        " generated as this version is too older to use."
+        " Please  use another new availble version of the dataset"
+        " (new availble version: ['1.0.0', '2.0.0'])")
+    builder = DummyDatasetSharedGenerator(version="0.0.8")
```
Conchylicultor: I think it's best to create a new specific dataset in this file for this.
This allow the test to be self-contained and reader don't have to look into multiple files to understand the test..

tensorflow_datasets/core/dataset_builder.py
Eshan-Agarwal: Yes its good to have only once, I will make changes

Eshan-Agarwal: @Conchylicultor  i am not understanding what you are asking when you look on changes there is a check for `if self.version.tfds_version_to_prepare:` which raises exception and then there is a check which check if version is canonical or maximum and it raises error and there is two assertion errors  because when canonical and max version is same it should recommend only one of them version .

tensorflow_datasets/core/dataset_builder_test.py
Eshan-Agarwal: Thanks for review this, I will make changes soon

tensorflow_datasets/core/dataset_builder.py
Conchylicultor: * To merge the two raise here, you should use `set()` instead of `list` so that duplicate versions are merged.
* To merge this one with the `tfds_version_to_prepare `, the `tfds_version_to_prepare` check should be added to this error message if present, as `tfds_version_to_prepare ` is only present for older versions.

Eshan-Agarwal: @Conchylicultor  thanks for suggestions it is good to use set as they take only unique elements in it just forgot to use them here, I will make changes soon

Eshan-Agarwal: @Conchylicultor  if we merge `tfds_version_to_prepare` check with new version checks so they both show same error message , it is good to include unittest for  `tfds_version_to_prepare` as both of them shows same errors.

Conchylicultor: Yes, we should have unitests for the different way to raise error.

Eshan-Agarwal: Thanks for reply I will make changes soon.

Eshan-Agarwal: @Conchylicultor  If we are including both unittests so I think I have to change in `test_utils.py file` in `utils.Version("0.0.7", tfds_version_to_prepare="v1.0.0"),]`, 0.0.7 to 2.0.0 for test passes. Also for test purpose you suggest to create new specific dataset in `test_with_old_version` so should I include mnist or re create new one in `test_util`

Eshan-Agarwal:
@Conchylicultor  please have a look on the changes for unittest of newly added check I use DummyMnist which I updated slightly in versions to successfully run tests

tensorflow_datasets/core/dataset_builder_test.py
Conchylicultor: By the way, why haven't you done this ? As explain it's best for datasets to be self-contained.

Eshan-Agarwal:
> Looks better thank you. There are still some issues I'll fix internally.
> 
> For instance, `SUPPORTED_VERSIONS = [` should contains `Version("1.0.0")` rather than strings.
> 
> Also if `tfds_version_to_prepare` is present, it's best to add it to the error which is raised as debug info.

Sorry for trouble I will fix this. I am confused I thought you said to combine the error messages for `tfds_version_to_prepare ` with new checks

Conchylicultor:
I've fixed those changes internally, so no need to do anything for this.

vijayphoenix:
@Eshan-Agarwal you can close this PR as it got merged internally https://github.com/tensorflow/datasets/commit/0e1e4f854d2b7a43fe776837f4b6f7f2834e3ac7

Issue #1716, opened 2020-03-24
[GSoC] Update checksums for missing urls


Conchylicultor:
Thanks to https://github.com/tensorflow/datasets/commit/5a8ee376c39a885c9132cdd22a8b2669febe6366 (https://github.com/tensorflow/datasets/issues/1397), we now check that all urls from `dl_manager.download()` are correctly registered.

However, some of the urls are not registered for some of our datasets. We should generate the checksums for the missing urls by running `download_and_prepare` script with `--register_checksums` as per https://www.tensorflow.org/datasets/add_dataset#2_run_download_and_prepare_locally.

Note: The dataset don't need to be fully generated, so you can modify the dataset implementation to only download the files.

The datasets to updates are the ones with `SKIP_CHECKSUMS = True`.
To know which urls are missing, just run the tests and the error message should show the missing urls.

Eshan-Agarwal:
wmt_test.py also have skip_checksum = True because its checksum is defined individually like in wmt_19 etc, so no need to cahnge ?

Eshan-Agarwal:
Also it shows `skip_checksum = True` for `wikipedia_test.py ` but wikipedia have correct checksum file
[see](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/url_checksums/wikipedia.txt)

Conchylicultor:
@Eshan-Agarwal 

* wmt shouldn't be updated.
* If you run the wikipedia tests, there is one missing url.

Eshan-Agarwal:
thanks for your response, we have to update checksums for existing wikipedia.txt or create new one ?

Conchylicultor:
> thanks for your response, we have to update checksums for existing wikipedia.txt or create new one ?

Update the existing one. This is the only way to make tests pass.

Eshan-Agarwal:
@Conchylicultor  Thanks for reply, please have a look on PR #1718 I think checksum should be update in sun397 ?

Conchylicultor:
@Eshan-Agarwal Nice catch, I'll update this internally when merging the dataset.

Eshan-Agarwal:
Thanks for looking into this

vijayphoenix:
The tests fail in sun_test.txt fail because the name of url_checksums doesn't match. Updating url_checksums in sun397.txt doesn't work for me.

vijayphoenix:
[...]

PR #1718, opened 2020-03-24
Fix sun_test.py checksums

vijayphoenix:
Update sun397 checksums. (#1716)

### DIFF
--- a/tensorflow_datasets/image_classification/sun.py
+++ b/tensorflow_datasets/image_classification/sun.py
@@ -216,6 +216,7 @@ class Sun397(tfds.core.GeneratorBasedBuilder):
         citation=_SUN397_CITATION)
 
   def _split_generators(self, dl_manager):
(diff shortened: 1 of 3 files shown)


vijayphoenix:
@Conchylicultor The checksums error was caused because we were setting `DATASET_CLASS` as `TestableSun397`.
https://github.com/tensorflow/datasets/blob/147fc7a58856c03596f8e8f399df1084b075091d/tensorflow_datasets/image_classification/sun_test.py#L46

So, I updated it to
https://github.com/tensorflow/datasets/blob/a993b4ea27a0949ce4a7aac317e7f7c334a287f8/tensorflow_datasets/image_classification/sun_test.py#L38

Conchylicultor:
This was merged. You contribution was recorded in https://github.com/tensorflow/datasets/commit/b9199e11da8ad1723217a2c210bf6bdf1bafd16d but our internal system seems to have some bugs..

PR #1730, opened 2020-03-26
metadata_cleanup script for delete useless metadata files

### LINKED ISSUE #1534: [GSoC] Script to cleanup the testing/metadata dir

Eshan-Agarwal:
Fixes #1534 

@Conchylicultor  please review on `metadata_cleanup.py` script which cleans all metadata versions which are not registered.

This PR not includes dry run option if everything seems fine to you I will include it in my next commit.


tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,35 @@
...
+for i in range(0,len(paths)):
+    if paths[i] not in registered_versions:
+        number_of_meta_removed+=1
+        shutil.rmtree(os.path.join(os.pardir, "testing/metadata",paths[i]))
+
+print("*"*80)
+print("Total number of registered datasets : ", len(registered_versions))  
+print("Total number of subdirectories in metadata direcotry : ", len(paths)) 
+print("Total number of meta directories removed : ", len(number_of_meta_removed)) 
+print("*"*80)
```
Conchylicultor: Similarly to other `scripts/`, could you use absl and `app.run` ?

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,35 @@
...
+
+meta_dirs = []
+for root, dirs, files in os.walk(os.path.join(os.pardir, "testing/metadata")):
+    meta_dirs.append(root)
+
+# Get all paths in proper format.
+
+paths = []
+for meta_dir in meta_dirs:
+    match = re.search(r'\d.\d.\d', meta_dir)
```
Conchylicultor: Could you format your file ?

tensorflow_datasets/core/registered.py
```diff
@@ -23,6 +23,7 @@
 import abc
```
Conchylicultor: Please add `# Lint as...` and a docstring with instructions to run the script.

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: yes I tried and it successfully removes all files which are not in registered_versions from metadatafolder . I will shows you results of these print statements again

Eshan-Agarwal: > Similarly to other `scripts/`, could you use absl and `app.run` ?

Ok I will its just confirmation example if it seem okay I will re create it with proper format in next commits

tensorflow_datasets/core/registered.py
Eshan-Agarwal: It already have lint as, and is you want me to include docstring for metadata_cleanup.py here ?

Eshan-Agarwal:
@Conchylicultor  `registered_version` list not includes files like open_images_v4 so is we remove it ?

tensorflow_datasets/core/registered.py
Conchylicultor: My bad. I was looking at the wrong file.

Eshan-Agarwal:
@Conchylicultor  please have a look on the changes also I add results to it. This PR not include --dry-run option I will include it after your approval for these new changes. I got this output when run `metadata_cleanup.py` script.

> (keras-gpu) C:\Users\eshan\Desktop\TF_TFDS\datasets\tensorflow_datasets\scripts>python metadata_cleanup.py
********************************************************************************
Total number of registered datasets :  1990
Total number of subdirectories in metadata direcotry :  2368
Total number of meta directories removed :  1310
Total number of directories in metadata, empty after process :  7
********************************************************************************

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,141 @@
...
+    registered_versions = get_registered_versions()
+
+    number_of_meta_removed = 0
+
+    for i in range(0,len(subpaths)):
+        if subpaths[i] not in registered_versions:
+            number_of_meta_removed+=1
+
+            # Remove directories from metadata which not in registered version list
+            shutil.rmtree(os.path.join(metadata_path,subpaths[i]),ignore_errors =True)
```
Conchylicultor: Please use gfile for all file modifications.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,141 @@
...
+
+            empty_folder_paths = os.path.join(metadata_path,subpaths[i][:-6])
```
Conchylicultor: Avoid magic numbers. Where does `-6` come from ?

Eshan-Agarwal: @Conchylicultor  thanks for your response, When you look on metadata file paths its look like `.\\testing/mmd\\abstract_reasoning\\attr.rel.pairs\\0.0.2'`, So for remove sub directory completely from metadata directory I need  `.\\testing/mmd\\abstract_reasoning\\attr.rel.pairs\'` so this `-6` always woks as version are of form `x.y.z`. As some files like `open_images_v4` not have any version in registered versions but it is in `metadata `directory

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Okay I will make changes soon

Eshan-Agarwal:
> Thank you for the update, but this seems too complicated.
> 
> How about something like:
> 
> ```python
> def delete_metadata_dirs(metadata_dir):
>   all_full_names = set(_extract_full_names_from_dirs(metadata_dir))
>   valid_full_names = set(iter_dataset_full_names())
>   # Iterate over all extra full names
>   for extra_full_name in version in all_full_names - valid_full_names:
>     tf.io.gfile.rmtree(os.path.join(metadata_dir, extra_full_name))
> ```

yes its look complicated I will look this

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,141 @@
...
+"""
+from absl import app
+import os
+import re
+import shutil
+from tensorflow_datasets.core import registered
+
+metadata_path = os.path.join(os.pardir, "testing/metadata")
+
+def get_registered_versions():
```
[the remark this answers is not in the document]
Eshan-Agarwal: So should I remove `get_registered_versions()` function and use `registered.iter_dataset_full_names()` directly ? I sort them to look files alphabetically, don't need this here I can remove it .

Eshan-Agarwal:
@Conchylicultor  please review this again I change the script slightly.

Eshan-Agarwal:
@Conchylicultor please review this so that I complete my work on it.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,92 @@
...
+    for i in range(0, len(meta_paths)):
+        
+        if meta_paths[i] not in registered_paths:
+            number_of_meta_removed+=1
+            tf.io.gfile.rmtree(os.path.join(metadata_dir, meta_paths[i][1:]))
+
+    return number_of_meta_removed, len(registered_paths), len(meta_paths)
+
+
+def removeEmptyFolders(path, removeRoot=True):
```
Conchylicultor: Why using camel case ?  You know you have to format your code correctly, so you shouldn't wait for me to point this out.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,92 @@
...
+    for root, dirs, files in os.walk(metadata_dir):
+        for fileName in files:
+            meta_paths.add(os.path.join( root[len(metadata_dir):]))
+    meta_paths.remove('')
+    return list(meta_paths)
+
+def get_registered_version():
+    ''' Get total registered paths'''
+    registered_path = []
+    for i in registered.iter_dataset_full_names():
```
Conchylicultor: `i` usually is integer Please use better name

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,92 @@
...
+def _extract_full_names_from_dirs(metadata_dir):
+    '''Get all metadata direcotry versions paths '''
+    meta_paths = set()
+    for root, dirs, files in os.walk(metadata_dir):
+        for fileName in files:
+            meta_paths.add(os.path.join( root[len(metadata_dir):]))
+    meta_paths.remove('')
+    return list(meta_paths)
+
+def get_registered_version():
```
Conchylicultor: What are you trying to do with this function ? You should not need this and use `iter_dataset_full_names `

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,92 @@
...
+    ''' Get total registered paths'''
+    registered_path = []
+    for i in registered.iter_dataset_full_names():
+        registered_path.append(os.path.join("\\",i))
+    return registered_path
+
+
+def delete_metadata_dirs(metadata_dir):
+    '''Remove metadata directories not present in registered versions '''
+    number_of_meta_removed = 0
```
Conchylicultor: I don't think we need to count and print stats. This complexify the code and isn't required.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,92 @@
...
+    registered_paths = get_registered_version()
+    meta_paths= _extract_full_names_from_dirs(metadata_dir)
```
Conchylicultor: You should be consistent in the naming (alwyas use `_forler` prefix)

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: Why is it required ? Also the implementation seems overly complicated.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,92 @@
...
+    for i in range(0, len(meta_paths)):
```
Conchylicultor: Why haven't you use `set()` as suggested ? This would be simpler than this double loop.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,92 @@
...
+import shutil
+import tensorflow as tf
+from tensorflow_datasets.core import registered
+
+metadata_path = os.path.join(os.pardir, "testing/metadata")
+
+def _extract_full_names_from_dirs(metadata_dir):
+    '''Get all metadata direcotry versions paths '''
+    meta_paths = set()
+    for root, dirs, files in os.walk(metadata_dir):
```
Conchylicultor: You should not add all sub-directories. Only extract the `dataset_name/version` or `dataset_name/config/versions`.

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Its just to show the result to you I will change this

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Okay Thanks for suggestions I will

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: After removing metadata versions which are not present in register versions, there are some folders which are completely empty because `delete_metadata_dirs` removes versions and left empty folders so to cleanup every empty folder in Last I created it

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Yes this function is exactly doing this it takes only all `dataset_name/version` or `dataset_name/config/versions` from metadata folder, thats why I haven't include `fileName` in `meta_paths ` so it give only versions paths not files paths

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: This function create list of all register versions paths which is exactly matches with format of paths extracted from metadata. I will remove this function but then I have to define it in `delete_metadata_dirs` function

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Sorry for trouble, I will not do this mistake in future.

Eshan-Agarwal:
`_remove_empty_folders` function was used to delete all directories or sub directories which is completely empty after removing all metadata versions not present in registered versions.

Eshan-Agarwal:
@Conchylicultor  please review this. Include `--dry_run` for print all dataset version to be removed and `-r` to remove them.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+flags.DEFINE_boolean('dry_run', False, "Shows all versions\
+                      which are not registered")
+flags.DEFINE_boolean('r', False, "Removes unmatched\
+                      dataset versions from metadata")
+
+__metadata_path = os.path.join(os.pardir, "testing/metadata")
+
+def _extract_metadata_versions(metadata_dir):
+    '''
+       Get all metadata direcotry versions paths,
```
Conchylicultor: Could you format the doc string the similarly to the rest of the codebase ? And use `"""` instead of `'''`

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+    _meta_paths = set()
+    for root, dirs, files in tf.io.gfile.walk(metadata_dir):
+        for fileName in files:
+            _meta_paths.add(os.path.join(root[len(metadata_dir)+1:]))
+    _meta_paths.remove('')
+    return _meta_paths
+
+def _delete_metadata_dirs(metadata_dir):
+    '''Remove metadata versions not present in registered versions '''
+    _registered_path = set(i for i in registered.iter_dataset_full_names())
```
Conchylicultor: As already explained previously, `i`, `j` means integer by convention. You shouldn't use it to name string. Please use `n` or `x`

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+    _meta_paths= _extract_metadata_versions(metadata_dir)
+    for extra_full_name in sorted(_meta_paths-_registered_path):
+        if FLAGS.dry_run:
+            print(extra_full_name)
+        if FLAGS.r:
```
Conchylicultor: If `r=True`, then it will delete directories even when `dry_run=True`. You shouldn't have `r` flag. Only `dry_run`

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+
+def _extract_metadata_versions(metadata_dir):
+    '''
+       Get all metadata direcotry versions paths,
+       It only extract the paths like 'dataset_name/version'
+       or 'dataset_name/config/versions' in metadata dir
+    '''
+    _meta_paths = set()
+    for root, dirs, files in tf.io.gfile.walk(metadata_dir):
+        for fileName in files:
```
Conchylicultor: You should format your code correctly (only 2 space for tab). Did you run pylint ?

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+    '''Remove metadata versions not present in registered versions '''
+    _registered_path = set(i for i in registered.iter_dataset_full_names())
+    _meta_paths= _extract_metadata_versions(metadata_dir)
+    for extra_full_name in sorted(_meta_paths-_registered_path):
+        if FLAGS.dry_run:
+            print(extra_full_name)
+        if FLAGS.r:
+            tf.io.gfile.rmtree(os.path.join(metadata_dir, extra_full_name))
+    # delete directories which is completely empty after process.
+    _remove_empty_folders(__metadata_path)
```
Conchylicultor: Rather than globally removing the empty dir, I think it would have been simpler to check directly after the `tf.io.gfile.rmtree` with something like:

```
curr_dir = os.path.join(metadata_dir, extra_full_name)
tf.io.gfile.rmtree(curr_dir)
_remove_empty_folders(curr_dir)
```
```python
def _remove_empty_folders(root_dir):
  root_dir = os.path.basedir(root_dir)
  if not tf.io.gfile.listdir(root_dir):  # Empty dir
    tf.io.gfile.remove(root_dir)
    _remove_empty_folders(root_dir)
```

But I don't think this is required at all.
If a folder is empty, I'm don't think it will appear in Github at all, so this could be removed entirely.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+    '''
+       Get all metadata direcotry versions paths,
+       It only extract the paths like 'dataset_name/version'
+       or 'dataset_name/config/versions' in metadata dir
+    '''
+    _meta_paths = set()
+    for root, dirs, files in tf.io.gfile.walk(metadata_dir):
+        for fileName in files:
+            _meta_paths.add(os.path.join(root[len(metadata_dir)+1:]))
+    _meta_paths.remove('')
```
Conchylicultor: Why is it required ?

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: Also why not `set(registered.iter_dataset_full_names())`

Eshan-Agarwal:
@Conchylicultor  Please have a look. I made changes

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+def _delete_metadata_dirs(metadata_dir):
+    '''Remove metadata versions not present in registered versions '''
+    _registered_path = set(i for i in registered.iter_dataset_full_names())
+    _meta_paths= _extract_metadata_versions(metadata_dir)
+    for extra_full_name in sorted(_meta_paths-_registered_path):
+        if FLAGS.dry_run:
+            print(extra_full_name)
+        if FLAGS.r:
+            tf.io.gfile.rmtree(os.path.join(metadata_dir, extra_full_name))
+    # delete directories which is completely empty after process.
```
Conchylicultor: Comments should start by capital

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: Do not use CamelCase

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+def _extract_metadata_versions(metadata_dir):
+    '''
+       Get all metadata direcotry versions paths,
+       It only extract the paths like 'dataset_name/version'
+       or 'dataset_name/config/versions' in metadata dir
+    '''
+    _meta_paths = set()
+    for root, dirs, files in tf.io.gfile.walk(metadata_dir):
+        for fileName in files:
+            _meta_paths.add(os.path.join(root[len(metadata_dir)+1:]))
```
Conchylicultor: I see what you've done but this would only work if metadata do not contains sub-directory. Otherwise it might accidentally remove metadata from other folders.

It would be safer to check the path format.

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+    _meta_paths.remove('')
+    return _meta_paths
+
+def _delete_metadata_dirs(metadata_dir):
+    '''Remove metadata versions not present in registered versions '''
+    _registered_path = set(i for i in registered.iter_dataset_full_names())
+    _meta_paths= _extract_metadata_versions(metadata_dir)
```
Conchylicultor: Variables names should not starts with `_`, only function to indicates they are private. Please read https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: Why iterating over `for fileName in files:` when you're not using it ?

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,97 @@
...
+from tensorflow_datasets.core import registered
+
+
+FLAGS = flags.FLAGS
+flags.DEFINE_boolean('dry_run', False, "Shows all versions\
+                      which are not registered")
+flags.DEFINE_boolean('r', False, "Removes unmatched\
+                      dataset versions from metadata")
+
+__metadata_path = os.path.join(os.pardir, "testing/metadata")
```
Conchylicultor: Double underscore has specific meaning in Python. Please don't use this.

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: Why `os.path.join` if you're not joining anything ?

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: they are used for private something ? so i think to use them here but its okay i will remove them

Conchylicultor: In Python single and double underscore have very different meaning `_` and `__`. You should learn on the internet the difference.

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: this is because to take only paths in format dataset/version because of `for fileName in files`  it traverse on each files and if we are not using them we are also getting directory path with `dataset/version` paths

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-a-single-and-a-double-underscore-before-an-object-name

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Yes I provide `r` flag so that we can check or delete directories together, `r` is only true when we pass this argument otherwise we simply check by using `dry_run` not including `r`

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: So should I remove this completely ?

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: No its work fine I tested it like in Wikipedia directory `testing\metadata\wikipedia\20190301.hz\0.0.3`  it removes all files as this is not present in register versions

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: yes

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: I meant that ` testing\metadata\wikipedia\20190301.hz\0.0.3\som_dir` would be deleted too even if they should not. Ideally we should add some `_validate_name(root[len(metadata_dir)+1:])` which would check the name is of format `x/y/a.b.c`.

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Thanks for giving resource

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: You should not have to do this if you're using `_validate_name` suggested bellow.
Also this is inefficient. `if len(files)` would be much better

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: `_meta_paths.remove('')` is because it returns empty str because one directory path is like testing/metadata.

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Okay thanks for suggestions

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Thanks for explanation. Yes it removes it too but is there is any `testing\metadata\wikipedia\20190301.hz\0.0.3\som_dir` directory is present inside versions like `0.0.3` in metadata ?
If so I have to match patterns as you suggest and as I was done in my previous commit?

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: Matching patterns like I was done in my previous commit [here](https://github.com/tensorflow/datasets/pull/1730/commits/651711349336790414ece3b41c8dc38510abf739#diff-d3640dd24433be763b6ce93db70001c5L73-L80)

tensorflow_datasets/scripts/metadata_cleanup.py
Eshan-Agarwal: So is it good to include it or should I remove it?

tensorflow_datasets/scripts/metadata_cleanup.py
Conchylicultor: We want to make sure we won't create bug in the future.

To match the pattern, I would add a new `is_full_name` function to register, and would make sure to add unittests to check this works properly.

```
register.is_full_name('ds/config/1.0.2/other') -> False
register.is_full_name('ds/config/1.0.2/') -> False
register.is_full_name('ds/config/1.2') -> False
register.is_full_name('ds/config') -> False
register.is_full_name('ds/1.2') -> False

register.is_full_name('ds/config/1.0.2') -> True
register.is_full_name('ds/1.0.2') -> True
register.is_full_name('ds_with_number123/1.0.2') -> True
```

Thank you very much.

Eshan-Agarwal: Thanks you so much it will really help so we can get rid of `for fileName in files:`   we can use `tf.iogfile.glob() `for `is_full_name`?

Conchylicultor: `is_full_name` just take a string, not a path, so there is no path / gfile involved. A simple regex should be enough

Eshan-Agarwal: Thank you , I will complete this soon

Eshan-Agarwal:
@Conchylicultor  Please have a look I added `is_full_name` in `registered.py`

tensorflow_datasets/scripts/metadata_cleanup.py
```diff
@@ -0,0 +1,74 @@
...
+metadata_path = os.path.join(os.pardir, "testing/metadata")
+
+
+def _extract_metadata_versions(metadata_dir):
+  """Get all metadata direcotry versions paths,
+  It only extract the paths like 'dataset_name/version'
+  or 'dataset_name/config/versions' in metadata dir
+
+  Args:
+  metadata_dir : Path to metadat directory (testing/metadata)
```
Conchylicultor, line 40: Could you try to format the docstring like the other files: http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

And don't forget the return sections

tensorflow_datasets/core/registered.py
```diff
@@ -379,3 +380,30 @@ def _cast_to_pod(val):
...
+      for config in builder_cls.BUILDER_CONFIGS:
+        for v in _get_all_versions(
+            [config.version] + config.supported_versions):
+          yield os.path.join(builder_name, config.name, v)
+    else:
+      for v in _get_all_versions(
+          [builder_cls.VERSION] + builder_cls.SUPPORTED_VERSIONS):
+        yield os.path.join(builder_name, v)
+
+def is_full_name(path_string):
```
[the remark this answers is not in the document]
Conchylicultor: By unittests, I mean this: https://docs.python.org/3/library/unittest.html

You should update `registered_test.py` to make sure the function returns the correct values.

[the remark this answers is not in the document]
Eshan-Agarwal: Yes, working on it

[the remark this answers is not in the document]
Eshan-Agarwal: For unittest only one true path is efficient ?

Eshan-Agarwal:
@Conchylicultor  Please review this Updated unitests, all tests are passed see [gist](https://gist.github.com/Eshan-Agarwal/31fa191468e80cd70bb4c35cc863d2d9). Also I slightly modified script please look this is because in previous commit I forgot to consider name `attr.rel.pairs` due to dots its not matching but its corrected in it.

And output is also updated as per latest metadata dir

Issue #1644, opened 2020-03-14
`pylint` Errors. 


ChanchalKumarMaji:
**Short description**
Many `pylint` errors exists in TFDS. Running `pylint` not generating 10.0/10 rating. 

For example running `./oss_scripts/lint.sh tensorflow_datasets/image` gives - 
```
Using config file /tmp/pylintrc
************* Module tensorflow_datasets.image.so2sat
C: 93, 2: Missing method docstring (missing-docstring)
C:150, 6: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)
************* Module tensorflow_datasets.image.clevr
C:100, 2: Missing method docstring (missing-docstring)
************* Module tensorflow_datasets.image.places365_small_test
C:  1, 0: Missing module docstring (missing-docstring)
************* Module tensorflow_datasets.image.patch_camelyon
C: 69, 2: Missing method docstring (missing-docstring)
************* Module tensorflow_datasets.image.eurosat
C:102, 2: Missing method docstring (missing-docstring)
************* Module tensorflow_datasets.image.caltech_birds
C: 84, 2: Missing method docstring (missing-docstring)
************* Module tensorflow_datasets.image.oxford_flowers102_test
C: 44, 0: Trailing newlines (trailing-newlines)
************* Module tensorflow_datasets.image.patch_camelyon_test
C: 27, 0: Missing class docstring (missing-docstring)
************* Module tensorflow_datasets.image.visual_domain_decathlon
C:111, 2: Missing method docstring (missing-docstring)
************* Module tensorflow_datasets.image.sun
C:225, 2: Missing method docstring (missing-docstring)
R:242, 4: Unnecessary "else" after "return" (no-else-return)
C:276, 2: Missing method docstring (missing-docstring)
C:305, 2: Missing method docstring (missing-docstring)
************* Module tensorflow_datasets.image.stanford_online_products_test
C:  1, 0: Missing module docstring (missing-docstring)
************* Module tensorflow_datasets.image.duke_ultrasound
C:116, 2: Missing method docstring (missing-docstring)
C:143, 2: Missing method docstring (missing-docstring)
************* Module tensorflow_datasets.image.imagenet2012_corrupted_test
C: 28, 0: Missing class docstring (missing-docstring)
************* Module tensorflow_datasets.image.cars196_test
C:  1, 0: Missing module docstring (missing-docstring)
************* Module tensorflow_datasets.image.caltech_test
[...]

PR #1731, opened 2020-03-26
Fix pylint for tfds script and proto

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/1644

### DIFF
--- a/tensorflow_datasets/scripts/document_datasets_test.py
+++ b/tensorflow_datasets/scripts/document_datasets_test.py
@@ -24,7 +24,7 @@ from tensorflow_datasets.core import dataset_builder
 from tensorflow_datasets.core import utils
 from tensorflow_datasets.scripts import document_datasets
 
-DummyMnist = testing.DummyMnist
+DummyMnist = testing.DummyMnist # pylint: disable = invalid-name
 
 
 class DummyMnistConfigs(DummyMnist):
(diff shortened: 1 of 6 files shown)


tensorflow_datasets/scripts/document_datasets_test.py
```diff
@@ -40,20 +40,20 @@ class DummyMnistConfigs(DummyMnist):
 
 
 class DocumentDatasetsTest(testing.TestCase):
-
+  """Test for document_datasets module"""
   @classmethod
-  def setUpClass(cls):
+  def setUpClass(cls): # pylint: disable = invalid-name
```
Eshan-Agarwal: It is in camel case. Please change this don't disable it.

tensorflow_datasets/scripts/document_datasets_test.py
```diff
@@ -40,20 +40,20 @@ class DummyMnistConfigs(DummyMnist):
...
     super(DocumentDatasetsTest, cls).setUpClass()
     cls._tfds_tmp_dir = testing.make_tmp_dir()
     builder = DummyMnist(data_dir=cls._tfds_tmp_dir)
     builder.download_and_prepare()
 
   @classmethod
-  def tearDownClass(cls):
+  def tearDownClass(cls): # pylint: disable = invalid-name
```
Eshan-Agarwal: Same as above

tensorflow_datasets/scripts/document_datasets_test.py
```diff
@@ -40,20 +40,20 @@ class DummyMnistConfigs(DummyMnist):
...
     super(DocumentDatasetsTest, cls).tearDownClass()
     testing.rm_tmp_dir(cls._tfds_tmp_dir)
 
-  def setUp(self):
+  def setUp(self): # pylint: disable = invalid-name
```
Eshan-Agarwal: same

tensorflow_datasets/scripts/document_datasets_test.py
vijayphoenix: Thanks for the review but I feel that we should not change the naming format here because it is a common practice to name the class setup methods in such way

tensorflow_datasets/scripts/document_datasets_test.py
vijayphoenix: We can change this.
Thank you for pointing this out.

NikhilBartwal (approved):

PR #1742, opened 2020-03-27
Update groove to use Audio feature 

### LINKED ISSUE #1741: [GSoC] Update Groove and Nsynth to use tfds.features.Audio
It seems that groove dataset is currently using `tfds.features.Tensor` to store the audio. For consistency with other datasets, we should use `tfds.features.Audio` instead.

https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/groove.py#L135

Eshan-Agarwal:
Fixes #1741 @Conchylicultor  please have a look on this [colab](https://colab.research.google.com/drive/1jGcEmcEg17ledth5A0HBeHL-7WUTLJs-) it shows not it henerate audio_keys successfully

### DIFF
--- a/tensorflow_datasets/audio/groove.py
+++ b/tensorflow_datasets/audio/groove.py
@@ -130,8 +130,9 @@ class Groove(tfds.core.GeneratorBasedBuilder):
             "secondary": tf.string,
         },
         "midi": tf.string,
-        'audio': tfds.features.Audio(file_format='wav')
     }
+    if self.builder_config.include_audio:
+      features_dict["audio"] = tfds.features.Audio(file_format='wav')
     return tfds.core.DatasetInfo(
         builder=self,
         description=_DESCRIPTION,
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/audio/groove.py
```diff
@@ -129,11 +129,9 @@ def _info(self):
             "primary": tfds.features.ClassLabel(names=_PRIMARY_STYLES),
             "secondary": tf.string,
         },
-        "midi": tf.string
+        "midi": tf.string,
+        'audio': tfds.features.Audio(file_format='wav')
     }
-    if self.builder_config.include_audio:
-      features_dict["audio"] = tfds.features.Tensor(
```
Conchylicultor: You should not modify the logic. Otherwise some configs will fails

Eshan-Agarwal:
@Conchylicultor  please look on[ `nsynth.py`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/nsynth.py#L149) it also have `tfds.features.Tensor`

Conchylicultor:
The reason why this was this wasn't using audio was that the Tensor dtype was float32, instead of int64.
I ended up fixing this internally and will updload to github soon.

PR #1745, opened 2020-03-27
Use `tfds.feature.Audio` in nsynth

### LINKED ISSUE #1741: [GSoC] Update Groove and Nsynth to use tfds.features.Audio

vijayphoenix:
Fix #1741

### DIFF
--- a/tensorflow_datasets/audio/nsynth.py
+++ b/tensorflow_datasets/audio/nsynth.py
@@ -151,8 +151,8 @@ class Nsynth(tfds.core.BeamBasedBuilder):
         "id":
             tf.string,
         "audio":
-            tfds.features.Tensor(
-                shape=(_AUDIO_RATE * _NUM_SECS,), dtype=tf.float32),
+            tfds.features.Audio(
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/audio/nsynth.py
```diff
@@ -169,14 +170,14 @@ def _info(self):
       f0_and_ld_shape = (_F0_AND_LOUDNESS_RATE * _NUM_SECS,)
       features["f0"] = {
           "hz":
-              tfds.features.Tensor(shape=f0_and_ld_shape, dtype=tf.float32),
```
Conchylicultor: Only the audio feature should be updated.

Conchylicultor:
The reason why this was this wasn't using audio was that the Tensor dtype was float32, instead of int64.
I ended up fixing this internally and will updload to github soon.

PR #1746, opened 2020-03-27
Update readme.md

Nabagata:
Added TensorFlow logo in readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+![tf logo](https://www.tensorflow.org/images/tf_logo_social.png)
 # TensorFlow Datasets
 
 TensorFlow Datasets provides many public datasets as `tf.data.Datasets`.


Nabagata:
@googlebot I signed it!

Conchylicultor:
Thank you for your contribution, but I'm not sure we should add the TF logo, which may be confusing with the tensorflow/tensorflow github. Also the logo seems too big and may hurt readability. If anything, we should add a TFDS logo, but not sure we have this yet.

Nabagata:
> Thank you for your contribution, but I'm not sure we should add the TF logo, which may be confusing with the tensorflow/tensorflow github. Also the logo seems too big and may hurt readability. If anything, we should add a TFDS logo, but not sure we have this yet.

But the main TF repo has the same logo with the same size. Maybe we can add the same logo but smaller size.

Issue #1496, opened 2020-02-23
CelebaHQ manual data preparation no longer possible?


zplizzi:
The docs [here](https://www.tensorflow.org/datasets/catalog/celeb_a_hq) suggest that you need a set of `.tar` files in order to use the CelebaHQ dataset, but the [linked repository](https://github.com/tkarras/progressive_growing_of_gans) for generating them now has what appears to be a different format for the generated dataset, with `.tfrecords` files. I dug around in the history of that repo but couldn't find instructions for the old way of generating `.tar` files. If someone could link to a way to generate these old-style files it would be much appreciated, otherwise I think this dataset loader needs updated.

dhirensr:
@zplizzi : I agree that the mentioned repo doesn't has any links to download the .tar file, however I think this link would help you https://drive.google.com/drive/folders/1jU-hzyvDZNn_M3ucuvs9xxtJNc9bPLGJ to download celeba 1024x1024 images.

Conchylicultor:
Thanks for catching this. It seems the processing is now done with https://github.com/tkarras/progressive_growing_of_gans/blob/master/dataset_tool.py#L454 but not sure where the data is coming from.
If you found where they moved file (e.g. looking at the README history, or even open a bug in their repo.), don't hesitate to send a PR to update the dataset.

dhirensr:
@Conchylicultor : I tried to find the links however was not sucessful. However I found the drive link where one can download the celebahq 1024x1024 images. (https://drive.google.com/drive/folders/1jU-hzyvDZNn_M3ucuvs9xxtJNc9bPLGJ)
so maybe if we are not able to find the link, then we can change the processing code of the dataset to adapt to them.

vijayphoenix:
Hi @Conchylicultor @dhirensr @zplizzi 
I don't think a lot of changes will be required if we use the .zip files from this [drive link](https://drive.google.com/open?id=11Vz0fqHS2rXDb5pprgTjpD7S2BAJhi1P).

Thank you

moflo:
Hi - there is an s3 mirror of CelebA images available and I wrote a TFDS loader for it:
https://github.com/moflo/celebalocal

PR #1747, opened 2020-03-28
Fix CelebaHQ manual data preparation

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/1496


vijayphoenix:
@Conchylicultor Please review the changes.
Thank you

Conchylicultor:
I think there is conflicts. Could you try to sync to head ?

tensorflow_datasets/image/celebahq.py
```diff
@@ -105,8 +107,8 @@ def _info(self):
             "image":
                 tfds.features.Image(
                     shape=(self.builder_config.resolution,
-                           self.builder_config.resolution, 3),
-                    encoding_format="png"),
```
Conchylicultor: Why removing this ?

Conchylicultor:
Tests are failings. Could you fix them?

vijayphoenix:
@Conchylicultor The error was caused because of the `iter_archive()`.
It also returned paths to the directories. 
Made a PR to fix this https://github.com/tensorflow/datasets/pull/1750

tensorflow_datasets/image/celebahq.py
```diff
@@ -68,20 +69,23 @@ def __init__(self, resolution, **kwargs):
...
 class CelebAHq(tfds.core.GeneratorBasedBuilder):
   """Celeba_HQ Dataset."""
 
   MANUAL_DOWNLOAD_INSTRUCTIONS = """\
-  manual_dir should contain multiple tar files with images (data2x2.tar,
-  data4x4.tar .. data1024x1024.tar).
+  manual_dir should contain the following compressed files with images.
+  (data2x2.zip, data4x4.tar .. data1024x1024.bz2).
+  You can find some of the celebaHQ data here:
+  https://drive.google.com/drive/folders/11Vz0fqHS2rXDb5pprgTjpD7S2BAJhi1P
```
Conchylicultor: Where is this url coming from. Is it an official url ?

tensorflow_datasets/image/celebahq.py
```diff
@@ -116,23 +120,25 @@ def _info(self):
...
-      msg = "You must download the dataset files manually and place them in: "
-      msg += dl_manager.manual_dir
-      msg += " as .tar files. See testing/test_data/fake_examples/celeb_a_hq "
+    file_names = ([
+        os.path.join(dl_manager.manual_dir, self.builder_config.file_name + ext)
+        for ext in self.builder_config.extensions])
+    image_file = list(filter(tf.io.gfile.exists, file_names))
+    if not image_file:
+      msg = ("You must manually download the compressed dataset files with any"
+             f" of the following extensions {self.builder_config.extensions}"
```
Conchylicultor: We still support Python2 internally, please remove f-string

tensorflow_datasets/image/celebahq.py
```diff
@@ -68,20 +69,23 @@ def __init__(self, resolution, **kwargs):
         version=v2,
         **kwargs)
     self.resolution = resolution
-    self.file_name = "data%dx%d.tar" % (resolution, resolution)
+    self.file_name = "data%dx%d" % (resolution, resolution)
+    self.extensions = resource._KNOWN_EXTENSIONS
```
Conchylicultor: Why is it required ? You should not import protected members.

tensorflow_datasets/image/celebahq.py
```diff
@@ -116,23 +120,25 @@ def _info(self):
 
   def _split_generators(self, dl_manager):
     """Returns SplitGenerators."""
-    image_tar_file = os.path.join(dl_manager.manual_dir,
```
Conchylicultor: Why modifying the logic here ?

tensorflow_datasets/image/celebahq.py
vijayphoenix: I found it on fast.ai forum.
https://forums.fast.ai/t/download-celeba-hq-dataset/45873/3

tensorflow_datasets/image/celebahq.py
vijayphoenix: Noted. I will replace it with value of `resource._KNOWN_EXTENSIONS`.
I used it for making the code more generic by supporting all types of archives.

tensorflow_datasets/image/celebahq.py
vijayphoenix: I did not modify the base logic.
Just updated it to support all types of archives `zip`, `tar`, etc.

Conchylicultor: I understand the motivation but this add unnecessary complexity to the code. Please revert the change.

vijayphoenix: The drive files contain only `.zip` files.
So, should I change to `tar` to `zip`?

Conchylicultor: Yes

tensorflow_datasets/image/celebahq.py
Conchylicultor: Does that mean this is not the original files ? We should always download from the original source. Mirrors from external users can be corrupted so we should not trust them.

Could you update the dataset to download the official data ?

tensorflow_datasets/image/celebahq.py
Conchylicultor: From what I understand, thos zip files re not the official source, so we should not use them.

tensorflow_datasets/image/celebahq.py
Conchylicultor: If anything we should contact the official team so they upload a better version of their files.

vijayphoenix: Okay, sure

vijayphoenix: @Conchylicultor I could do that but the code says
```
# The current celebahq generation code depends on a concrete version of 
# pillow library and cannot be easily ported into tfds.
```

Conchylicultor: Then the users should process the data themselves externally from the celeb_a official website and then put this data into the MANUAL_DIR

tensorflow_datasets/image/celebahq.py
```diff
@@ -68,20 +68,20 @@ def __init__(self, resolution, **kwargs):
         version=v2,
         **kwargs)
     self.resolution = resolution
-    self.file_name = "data%dx%d.tar" % (resolution, resolution)
-
+    self.file_name = "data%dx%d" % (resolution, resolution)
+    self.extensions = ['.tar', '.zip', '.tar.gz', '.tgz']
```
Conchylicultor, line 72: We should not support many format. We should only support the original datasets files from: https://github.com/tkarras/progressive_growing_of_gans.

From the README, it seems data is coming from https://drive.google.com/corp/drive/folders/0B4qLcYyJmiz0TXY1NG02bzZVRGs

Either this dataset should be updated to automatically download and apply the official pre-processing, or the pre-processed `tf_record` files should be generated using https://github.com/tkarras/progressive_growing_of_gans tools and placed inside the `manual_dir`

tensorflow_datasets/image/celebahq.py
```diff
@@ -116,19 +116,20 @@ def _info(self):
 
   def _split_generators(self, dl_manager):
     """Returns SplitGenerators."""
-    image_tar_file = os.path.join(dl_manager.manual_dir,
-                                  self.builder_config.file_name)
-    if not tf.io.gfile.exists(image_tar_file):
-      # The current celebahq generation code depends on a concrete version of
-      # pillow library and cannot be easily ported into tfds.
```
vijayphoenix, line 123: Closing this because of a lack of solutions for now.
We can integrate it if the upstream dataset gets updated

PR #1750, opened 2020-03-28
`iter_archive()` should not return directory path name

vijayphoenix:
Problem: The function `iter_archive()` returned path to all directories in a `zip` file.


vijayphoenix:
Consider a zip file `file.zip`. It contains a folder `dir` and an image `dir/image.jpg`
The expected output of `iter_archive(file.zip)` is
```
dir/image.png <ExFileObject>
```

but the actual output was 
```
dir <ExFileObject>
dir/image.png <ExFileObject>
```
So, it should not return `(path_in_archive, f_obj)` for a directory in a zip file

Eshan-Agarwal:
@vijayphoenix Yes it should not , thanks for clarifying.

tensorflow_datasets/core/download/download_manager.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, url, tmp_path):
     Exception.__init__(self, msg)
 
 
-class DownloadConfig(object):
```
Conchylicultor: We still want to support Python 2 internally, could you revert those changes ?

Conchylicultor:
Could you explain what is the issue if directories are returned ?

vijayphoenix:
Archive iterators of other file types like`tar`, `bz2` etc do not return directories. 
So to maintain uniformity, I made these changes.

Also, we will have to write more code to exclude directories in  `DatasetBuilder _generate_examples` as yielding `fobj` of directories will give errors.

tensorflow_datasets/core/download/download_manager.py
vijayphoenix: Done

Conchylicultor:
@vijayphoenix  Do you have an example where this is an issue ?

Also in the code, I see that directories are filtered already:
```
extract_file = z.open(member)	      extract_file = z.open(member)
if extract_file:  # File with data (not directory):
```
What is the issue with the current implementation ?

vijayphoenix:
> What is the issue with the current implementation?

@Conchylicultor `z.open(member)` does not indicate if a `member` is a directory or not.
Use `ZipInfo.is_dir()` instead. See https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.is_dir

> @vijayphoenix Do you have an example where this is an issue?

The `celebahq_test` fails when we use `.zip` instead of `.tar`. (The code should work for all type of archive files as it uses `iter_archive()` )
See https://github.com/tensorflow/datasets/pull/1747#issuecomment-605429713.

vijayphoenix:
@Conchylicultor I had to add new test files because `iter_archive()` function worked correctly on them.

PR #1800, opened 2020-04-03
Reduce test fake example size for open_image_v4

### LINKED ISSUE #1792: [GSoC] Update fake data for caltech_birds2011 & other datasets
The fake data for `caltech_birds2011` is way to big (> 100MB). We should investigate where does this huge size comes from and try to reduce it.

Fake data is at https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/testing/test_data/fake_examples/caltech_birds2011
The test is at: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image_classification/caltech_birds_test.py

Among the other datasets which have huge fake data size are:
* abstract_reasoning
* waymo
* duke_ultra_sound
* starcraft

Those datasets take more than 70% of all fake data size. caltech_birds2011 is almost half of it.
Reducing the size of those fake data would have a huge impact on our github repository size.

Eshan-Agarwal:
fixes #1792


Eshan-Agarwal:
Closing this as after running better compression of data size already reduced

Issue #1830, opened 2020-04-07
[GSoC] Avoid registering tests datasets


Conchylicultor:
Currently `tfds.list_builder` list [`dummy_mnist`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/testing/test_utils.py#L393) and other test-only datasets. We should make sure to avoid registering those datasets.

* Add `with tfds.core.registered.skip_registeration():` context manager which when applied skip dataset registration. Changes should go in `registered.py`. Add tests in `registered_test.py`.
* Apply this to the public API: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/public_api.py#L24

```
with core.registered.skip_registeration():
  from tensorflow_datasets import testing
```

* Add a `test_dataset_registered.py` which check that `tfds.list_builder` do not contains `dummy_mnist` or other test datasets.

Eshan-Agarwal:
Is test only include dataset with prefix `dummy `is avoid?

Conchylicultor:
@Eshan-Agarwal No, this should works for any datasets, no matter which name it is.

Registration is done in https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/registered.py#L148

vijayphoenix:
@Conchylicultor Please review https://github.com/tensorflow/datasets/pull/1832
Thank you

Conchylicultor:
This should be fixed thanks to #1833

PR #1832, opened 2020-04-07
Avoid registering tests datasets

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/1830

### DIFF
--- a/tensorflow_datasets/core/dataset_registered_test.py
+++ b/tensorflow_datasets/core/dataset_registered_test.py
@@ -20,19 +20,18 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from tensorflow_datasets import testing
-from tensorflow_datasets import list_builders
+import tensorflow_datasets as tfds
 
 _TEST_DATASETS = set(["dummy_mnist", "dummy_dataset_shared_generator"])
 
-class ListBuilderTest(testing.TestCase):
+class ListBuilderTest(tfds.testing.TestCase):
 
   def test_list_builder(self):
-    registered_datasets = set(list_builders())
+    registered_datasets = set(tfds.list_builders())
     self.assertEmpty(_TEST_DATASETS.intersection(registered_datasets),
                      "List of Registered datasets should not contain {}"
                      .format(_TEST_DATASETS)
                     )
 
 if __name__ == "__main__":
-  testing.test_main()
+  tfds.testing.test_main()
--- a/tensorflow_datasets/core/dataset_registered_test.py
+++ b/tensorflow_datasets/core/dataset_registered_test.py
@@ -23,16 +23,17 @@ from __future__ import print_function
 from tensorflow_datasets import testing
 from tensorflow_datasets import list_builders
 
+
 class ListBuilderTest(testing.TestCase):
 
-	_datasets = set(["dummy_mnist", "dummy_dataset_shared_generator"])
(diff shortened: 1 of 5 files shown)


tensorflow_datasets/core/dataset_registered_test.py
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+
+# Lint as: python3
+"""Tests for tensorflow_datasets.core.registered."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from tensorflow_datasets import testing
```
Conchylicultor, line 23: Here we want to test the global import `import tensorflow_datasets as tfds`.

tensorflow_datasets/core/registered.py
```diff
@@ -106,6 +107,14 @@
     version=r"[0-9]+\.[0-9]+\.[0-9]+",
 ))
 
+_skip_reg = True
+
+@contextmanager
+def skip_registeration():
+  global _skip_reg
+  _skip_reg = False
+  yield
```
Conchylicultor: Note, it's good practice to wrap yield inside `try`/'finaly' block,

This make sure that `_skip_reg = True` is executed even when an exception is raised inside the context manager

tensorflow_datasets/core/registered.py
```diff
@@ -106,6 +107,14 @@
     version=r"[0-9]+\.[0-9]+\.[0-9]+",
 ))
 
+_skip_reg = True
+
+@contextmanager
+def skip_registeration():
```
Conchylicultor: Please also add a unittest inside `registered_test.py` directly testing the feature.

tensorflow_datasets/core/registered.py
```diff
@@ -106,6 +107,14 @@
     version=r"[0-9]+\.[0-9]+\.[0-9]+",
 ))
 
+_skip_reg = False
+
+@contextmanager
+def skip_registration():
+  global _skip_reg
+  _skip_reg = True
+  yield
```
Conchylicultor, line 116: It is good practice to wraps yield inside a try/finally block, to execute `_skip_reg = False` even in case an exception is raised in the context manager.

tensorflow_datasets/core/registered.py
```diff
@@ -106,6 +107,14 @@
     version=r"[0-9]+\.[0-9]+\.[0-9]+",
 ))
 
+_skip_reg = False
+
+@contextmanager
+def skip_registration():
```
Conchylicultor, line 113: Could you also add a test in `registered_test.py` to test this feature directly (creates a new Builder class inside the context manager, without import) ?

tensorflow_datasets/core/registered.py
vijayphoenix: Okay

tensorflow_datasets/core/registered.py
vijayphoenix: Sure

vijayphoenix:
`test_load`, `test_determinism` in `dataset_builder_test.py` load `dummy_dataset_shared_generator` from the registered datasets.
We will have to change this.

vijayphoenix:
Merged by commit https://github.com/tensorflow/datasets/commit/3bf95a3571266d9048a2cbb91bdf12e8c92c0037

PR #1837, opened 2020-04-07
Fix Clevr dataset Test Set Data generation

### LINKED ISSUE #1834: [GSoC] Clevr dataset generation is failing
Clevr was updated recently in https://github.com/tensorflow/datasets/commit/af54aaf4b525fd6624319a0e38d6b3fc50870e23. 

However generation seems to be failing for the `test` set with:

```
tensorflow_datasets/image/clevr.py", line 107, in _generate_examples
    "answer": q["answer"],
KeyError: 'answer'
```

We should investigate and fix Clevr generation.

vijayphoenix:
Fix #1834
dataset_info.json: [link](https://gist.github.com/vijayphoenix/ededc2f2ea060b5bc7d918944d2b216f)


vijayphoenix:
Generated dataset only for Val and test set of clevr 3.1.0 version

Conchylicultor:
Thank you very much for confirming the issue and proposing a solution.
We ended up merging  an alternative implementation which uses `q.get("answer", "")` instead.

PR #1845, opened 2020-04-08
Update tfds/core tests to use tfds.testing

Eshan-Agarwal:


tensorflow_datasets/core/decode/base_test.py
```diff
@@ -22,7 +22,7 @@
 
 import numpy as np
 import tensorflow.compat.v2 as tf
-from tensorflow_datasets import testing
+import tensorflow_datasets as tfds
```
Conchylicultor, line 25: Note: In `tensorflow_datasets/core/`, we do not want to import the full API, but only individual modules. This might create circular dependencies and other issues..

Eshan-Agarwal: Okay so should I close this pr or change some files?

Conchylicultor: I'll apply the modifications internally. No need to do anything

PR #1855, opened 2020-04-10
Update wmt version

### LINKED ISSUE #1853: `stable_versions.txt` should not have `VERSION` equal to `None`
**Short description**
The list of stable versions of the datasets includes `wmt_translate/None`
https://github.com/tensorflow/datasets/blob/d87ac2c0ecc0655400841a3a502c7c6775f1b6bf/tensorflow_datasets/stable_versions.txt#L1525

So, this can be interpreted as a stable dataset `wmt_translate` with `VERSION = None`

**Enviroment-info**
* `tensorflow-datasets`/`tfds-nightly` version: tfds:2.1.0 and tfds-nightly

**Expected behavior**
Include a default version field in `wmt_translate` or exclude `wmt_translate` from the `tfds.core.registered.list_full_names()`.

vijayphoenix:
Fix #1853


Conchylicultor:
The WMT dataset do not have version on purpose because it cannot be instantiated without config. This is not the correct fix. If anything, the `list_full_names` should filter datasets for which no versions exists.

vijayphoenix:
@Conchylicultor so, we should update `list_full_names` such that it doesn't yield any dataset which no versions.

Conchylicultor:
@vijayphoenix Yes. I think this would be the proper fix.

PR #1861, opened 2020-04-11
Add support for custom decoders in mock_data

lgeiger:
`tfds.testing.mock_data` is really useful for integration testing of neural networks in user space. 

Often there is a need to customize decoding using `decoders={"image": tfds.decode.SkipDecoding()}` to speedup data preprocessing which is currently not supported my `tfds.testing.mock_data` since directly generates decoded data.

This PR adds support for custom decoders in `tfds.testing.mock_data`.

### DIFF
--- a/tensorflow_datasets/testing/mocking.py
+++ b/tensorflow_datasets/testing/mocking.py
@@ -22,6 +22,7 @@ from __future__ import division
 from __future__ import print_function
 
 import contextlib
+import functools
 import os
 import random
 
@@ -30,6 +31,7 @@ import numpy as np
 import tensorflow.compat.v2 as tf
 
 from tensorflow_datasets.core import features as features_lib
+from tensorflow_datasets.core import utils
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/testing/mocking.py
```diff
@@ -104,18 +106,34 @@ def mock_download_and_prepare(self, *args, **kwargs):
...
     del kwargs
+
+    features = self.info.features
+    # `from_generator` takes a callable with signature () -> iterable
+    # Recreating a new generator each time ensure that all pipelines are
+    # using the same examples
+    def generator():
+      for ex in RandomFakeGenerator(builder=self, num_examples=num_examples):
+        if decoders:
+          ex = features.encode_example(ex)
```
lgeiger, line 121: Unfortunately we cannot always encoded examples which would simplify the code, since it would fail for datasets using `Text` features, due to the way how the fake data is generated.

tensorflow_datasets/testing/mocking.py
```diff
@@ -160,17 +178,16 @@ def _generate_random_array(self, feature, tensor_info):
...
 
+    dtype = tensor_info.dtype
     # Generate some random values, depending on the dtype
-    if tensor_info.dtype.is_integer:
-      return self._rgn.randint(0, max_value, shape)
-    elif tensor_info.dtype.is_floating:
-      return self._rgn.random_sample(shape)
-    elif tensor_info.dtype == tf.string:
+    if dtype.is_integer:
+      return self._rgn.randint(0, max_value, shape).astype(dtype.as_numpy_dtype)
```
lgeiger, line 184: We need to properly cast the data here since `encode_example` would throw otherwise.

lgeiger:
@Conchylicultor Any chance you could take a look at this?

Conchylicultor:
Hey, sorry for the delay. I agree this is a feature we should add. Will try to merge this soon.

tensorflow_datasets/testing/mocking.py
Conchylicultor: Could you provide more context ? Why text encoding fails when encoding the random data ?

lgeiger: Sorry for the unclear description. Random data for the text features are generated as a list of random integers depending on the vocab size. However for encoding, the text feature expects strings as it often has a custom additional encoder to convert the text into integers encoding the vocabulary:
https://github.com/tensorflow/datasets/blob/2fc35db14c418e3c5415bb3511fdecabb8a1b252/tensorflow_datasets/core/features/text_feature.py#L111-L114

Conchylicultor: Thank for the explanations, that make sense. The encoder API is deprecated so hopefully it shouldn't really matter in this case.
I'm thinking but don't think there are other instance where `encode_example` expect a different format than what is returned by `decode_example`.
I'll try to simplify the code when merging this internally.

lgeiger: Thanks for looking into it.

Conchylicultor:
Thanks again, this should appear in the next tfds-nightly

PR #1885, opened 2020-04-16
Update visualization to return fig

Eshan-Agarwal:
Earlier `visualization/show_examples.py` was not returning figure plot see this [colab](https://colab.research.google.com/drive/1Lgj-HISneS9Vt3AU9NqbjwtVKlD5M9zd).

After changes see [colab](https://colab.research.google.com/drive/1zTI_igkUDnZkkdcRUOUNtLIR1JkR6oUS)


tensorflow_datasets/core/visualization/image_visualizer.py
```diff
@@ -27,6 +27,7 @@
 from tensorflow_datasets.core import lazy_imports_lib
 from tensorflow_datasets.core.visualization import visualizer
 
+plt = lazy_imports_lib.lazy_imports.matplotlib.pyplot
```
vijayphoenix, line 30: `lazy_imports` should not be global.
This will introduce unnecessary dependencies on the user and hence the purpose of `lazy_imports` is defeated.
Please revert back the changes.

tensorflow_datasets/core/visualization/show_examples.py
```diff
@@ -56,8 +56,8 @@ def show_examples(ds_info, ds, **options_kwargs):
   """
   for visualizer in _ALL_VISUALIZERS:
     if visualizer.match(ds_info):
-      visualizer.show(ds_info, ds, **options_kwargs)
-      break
+      fig = visualizer.show(ds_info, ds, **options_kwargs)
+      return fig
   else:
     raise ValueError(
         "Visualisation not supported for dataset `{}`".format(ds_info.name)
```
vijayphoenix, line 63: Consider rewriting as:
```
    if visualizer.match(ds_info):
      return visualizer.show(ds_info, ds, **options_kwargs)
raise ValueError(
        "Visualisation not supported for dataset `{}`".format(ds_info.name)
```

Eshan-Agarwal: @vijayphoenix 👍 Thanks for suggestions you are write it should be like this

tensorflow_datasets/core/visualization/image_visualizer.py
Eshan-Agarwal: Okay I will make changes but don't get like why importing them global make trouble, `lazy imports` are used for external dependencies and here in all methods lazy import for `matplotlib `was used ?

vijayphoenix: Consider a case when the `show_example()` of a different visualizer is called.
Assume, this visualizer does not require `matplotlib`.

But because of the global import, the user will get an error if he doesn't have this dependency installed.

Eshan-Agarwal: If I am correct, I make global for Image Visualizer which I think have no connection with other visualizers in future like Video or audio

Conchylicultor: @Eshan-Agarwal, vijayphoenix is right. This line will be executed during `import tensorflow_datasets as tfds` and will crash for all users which don't have this dependency

Eshan-Agarwal: @Conchylicultor thanks for clarify this, changing

vijayphoenix: Try reading the code of `show_examples.py`.

Eshan-Agarwal:
Thanks @vijayphoenix  @Conchylicultor  changed

Eshan-Agarwal:
@Conchylicultor please review

Conchylicultor:
Great! Will merge this soon!

Issue #1888, opened 2020-04-16
Horses_or_Humans - No Examples Yielded


lmoroney:
**Short description**
Using Windows 10

Run:
dataset_name = "horses_or_humans"
dataset, info = tfds.load(name=dataset_name, split=tfds.Split.TRAIN, with_info=True)

Dataset will download, but when extracting will give error:

Shuffling and writing examples to C:\Users\lmoro\tensorflow_datasets\horses_or_humans\3.0.0.incomplete185N4P\horses_or_humans-train.tfrecord
Traceback (most recent call last):
  File "D:/bookwork/testwindows/venv/test.py", line 5, in <module>
    dataset, info = tfds.load(name=dataset_name, split=tfds.Split.TRAIN, with_info=True)
  File "D:\bookwork\testwindows\venv\lib\site-packages\wrapt\wrappers.py", line 567, in __call__
    args, kwargs)
  File "D:\bookwork\testwindows\venv\lib\site-packages\tensorflow_datasets\core\api_utils.py", line 52, in disallow_positional_args_dec
    return fn(*args, **kwargs)
  File "D:\bookwork\testwindows\venv\lib\site-packages\tensorflow_datasets\core\registered.py", line 305, in load
    dbuilder.download_and_prepare(**download_and_prepare_kwargs)
  File "D:\bookwork\testwindows\venv\lib\site-packages\wrapt\wrappers.py", line 606, in __call__
    args, kwargs)
  File "D:\bookwork\testwindows\venv\lib\site-packages\tensorflow_datasets\core\api_utils.py", line 52, in disallow_positional_args_dec
    return fn(*args, **kwargs)
  File "D:\bookwork\testwindows\venv\lib\site-packages\tensorflow_datasets\core\dataset_builder.py", line 334, in download_and_prepare
    download_config=download_config)
  File "D:\bookwork\testwindows\venv\lib\site-packages\tensorflow_datasets\core\dataset_builder.py", line 1030, in _download_and_prepare
    max_examples_per_split=download_config.max_examples_per_split,
  File "D:\bookwork\testwindows\venv\lib\site-packages\tensorflow_datasets\core\dataset_builder.py", line 883, in _download_and_prepare
    self._prepare_split(split_generator, **prepare_split_kwargs)
  File "D:\bookwork\testwindows\venv\lib\site-packages\tensorflow_datasets\core\dataset_builder.py", line 1058, in _prepare_split
    shard_lengths = writer.finalize()
[...]

PR #1892, opened 2020-04-16
Fix to similar bug with horses or humans #1888

lmoroney:
Fix to allow file to be unzipped correctly on Windows

Similar to : https://github.com/tensorflow/datasets/issues/1888

## Description
Fixes bug where regex didn't give correct Windows path syntax, and files weren't unzipped correctly to \Users\ directory.

### DIFF
--- a/tensorflow_datasets/image_classification/rock_paper_scissors.py
+++ b/tensorflow_datasets/image_classification/rock_paper_scissors.py
@@ -40,7 +40,7 @@ _TEST_URL = "https://storage.googleapis.com/laurencemoroney-blog.appspot.com/rps
 _IMAGE_SIZE = 300
 _IMAGE_SHAPE = (_IMAGE_SIZE, _IMAGE_SIZE, 3)
 
-_NAME_RE = re.compile(r"^(rps|rps-test-set)/(rock|paper|scissors)/[\w-]*\.png$")
+_NAME_RE = re.compile(r"^(rps|rps-test-set)(/|\\)(rock|paper|scissors)(/|\\)[\w-]*\.png$")
 
 
 class RockPaperScissors(tfds.core.GeneratorBasedBuilder):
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/image_classification/rock_paper_scissors.py
```diff
@@ -40,7 +40,7 @@
 _IMAGE_SIZE = 300
 _IMAGE_SHAPE = (_IMAGE_SIZE, _IMAGE_SIZE, 3)
 
-_NAME_RE = re.compile(r"^(rps|rps-test-set)/(rock|paper|scissors)/[\w-]*\.png$")
+_NAME_RE = re.compile(r"^(rps|rps-test-set)(/|\\)(rock|paper|scissors)(/|\\)[\w-]*\.png$")
```
Conchylicultor, line 43: There shouldn't be brace as this indicates a capture group. Will fix this internally.

vijayphoenix:
@lmoroney would you mind checking if there is a similar bug in `caltech_birds`.
If there is a bug, a potential fix would be
```
_NAME_RE = re.compile(r"((\w*)[/|\\])*(\d*).(\w*)[/|\\](\w*.jpg)$")
```

Eshan-Agarwal:
@vijayphoenix  `caltech_birds` also have same problem in windows but suggested solution was not working path string are like this `CUB_200_2011\images\035.Purple_Finch\Purple_Finch_0091_27425.jpg` in windows and this path read as  `CUB_200_2011\images\x035.Purple_Finch\Purple_Finch_0091_27425.jpg` in windows because of `\035` it read this as hex `\x0035` which is problem

vijayphoenix:
@Eshan-Agarwal you can do the following
```
import pathlib
fp = pathlib.Path('CUB_200_2011\images\035.Purple_Finch\Purple_Finch_0091_27425.jpg')
fn = str(fp)
```
Now the backslash will be escaped and it will not be read as hexadecimal

Issue #1530, opened 2020-02-27
[GSoC] Update documentation generation script to display a few examples


Conchylicultor:
**Task:** It would be nice to have our dataset catalog to display a snippet of the dataset content, similarly to https://www.tensorflow.org/datasets/overview#visualization.

**Instructions:**

* Assume that the datasets are already generated and that you could load a few examples with either with `ds.take()` or the subsplit-API (DO NOT generate all datasets yourself, but only test on a few ones)
* Create a separate script to generate all datasets figures/content and save them in some new folder in our repo.
* Use [tfds.show_examples()](https://www.tensorflow.org/datasets/overview#visualization) to generate the image/content to save and display.
* You probably want to wrap the `tfds.show_examples(ds_info, ds)` in a `try/catch` block as not all datasets supports it
* Then update our [documentation template](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/templates/dataset.mako.md) to add a figure section which display the images saved previously.
* Make sure to test with a variety of datasets (with/without config, with/without images)
* You probably want to split the script generation and documentation update in separate step.

To test the documentation generation, you can use the following code snippet:

```python
import os

from tensorflow_datasets.scripts import generate_visualization  # Your new script
from tensorflow_datasets.scripts import document_datasets

DATASET_TO_TESTS = ['mnist', 'cifar10',...]  # Datasets you want to test the script on.
dst_dir = ...  # Destination directory

def main(_):
  for ds_name in DATASET_TO_TESTS:
    # 1) Generate the datasets (as script assume the datasets are already generated)
    tfds.load(ds_name)  

    # 2) Generate the figure for the dataset
    generate_visualization.generate_visualization([ds_name])

    # 3) Generate the documentation page which uses the generated figure in 2
    with open(os.path.join(dst_dir, f'{ds_name}.md')) as f:
      doc_content = document_datasets.dataset_docs_str([ds_name])
      f.write(doc_content)
```

**Difficulty**: Intermediate

vvkio:
I've started working on this, I will make a PR begging of next week

Nikhilnama18:
I would like to contribute on this issue.
Hey @vvkio , share your status, if possible we could work on this parallely.

vvkio:
Hey @Nikhilnama18, sure we can work on this together. What's left is writing the logic for updating the documentation and testing.

ManishAradwad:
[...]

PR #1909, opened 2020-04-18
Add scripts to display examples on datasets catalog

Eshan-Agarwal:
Issue #1530

@Conchylicultor  please review.

TODO : 
1. Update `dataset.mako.md` 
2. Generate one image per config. : means if datasets like `coco` have 3 config so should we add example grid for all 3 configs ?

Edit : results updated see [result](https://github.com/tensorflow/datasets/blob/802cb2023604870de8cb645a487e0cdd76463a61/docs/catalog/cats_vs_dogs.md)
Also check for variety of datasets (with/without config, with/without images) all works fine, here also includes result for `groove/full-midionly` gives ` No Example available for this dataset`


tensorflow_datasets/scripts/templates/dataset.mako.md
```diff
@@ -1,283 +1,310 @@
...
-%endif
-
-<%doc>First case: Single builder.</%doc>\
-% if not builder.builder_config:
-${display_builder(builder, all_sections)}
-<%doc>Second case: Builder configs.</%doc>\
-% else:
-${display_all_builders(config_builders)}
-% endif
+<%!
```
Conchylicultor, line 1: Could you fix the diffs ? I can't see what has changed here.

docs/catalog/cats_vs_dogs.md
```diff
@@ -57,5 +55,14 @@ month = {October},
...
+}
+```
+
+
+
+
+
+
+
+ * Figure:
```
Conchylicultor, line 67: Formatting should be consistent with other `* ` bullet points.

docs/catalog/cats_vs_dogs.md
```diff
@@ -57,5 +55,14 @@ month = {October},
 publisher = {Association for Computing Machinery, Inc.},
 url = {https://www.microsoft.com/en-us/research/publication/asirra-a-captcha-that-exploits-interest-aligned-manual-image-categorization/},
 edition = {Proceedings of 14th ACM Conference on Computer and Communications Security (CCS)},
-}
-```
+}
+```
+
```
Conchylicultor, line 60: Why too many empty lines

tensorflow_datasets/scripts/templates/dataset.mako.md
```diff
@@ -1,283 +1,310 @@
...
+    Section(get_source, display_source),
+    Section(get_versions, display_versions),
+    Section(get_download_size, display_download_size),
+    Section(get_dataset_size, display_dataset_size),
+    Section(get_manual, display_manual),
+    Section(get_autocache, display_autocache),
+    Section(get_splits, display_splits),
+    Section(get_features, display_features),
+    Section(get_supervised, display_supervised),
+    Section(get_citation, display_citation),
```
Conchylicultor, line 228: You should add a new `get_figure` section here.

tensorflow_datasets/scripts/templates/dataset.mako.md
Eshan-Agarwal: I think because of adding figure it shows this difference as I didn't change anything here

Eshan-Agarwal:
@Conchylicultor  please review update scripts to display one image per configs, Also includes results for various datasets of different configs, have a look on this [commit ](https://github.com/tensorflow/datasets/blob/cc78b273b9fc1c8ed5a6143c27f7ab268b26f7a1/docs/catalog/flic.md) of flic dataset.

tensorflow_datasets/core/visualization/show_examples.py
```diff
@@ -56,9 +56,7 @@ def show_examples(ds_info, ds, **options_kwargs):
   """
   for visualizer in _ALL_VISUALIZERS:
     if visualizer.match(ds_info):
-      visualizer.show(ds_info, ds, **options_kwargs)
-      break
```
Conchylicultor, line 60: Why remove the break ?

tensorflow_datasets/scripts/templates/dataset.mako.md
Conchylicultor: Could you search on the internet if there are solutions and why those diffs are showing if those lines hasn't been changed. Does `git diff` works ?

tensorflow_datasets/core/visualization/show_examples.py
Eshan-Agarwal: Because I am returning `visualizer.show(ds_info, ds, **options_kwargs)` this changes I have to do because in `generate_visualization.py` I am saving fig . See PR #1885 have these changes separately

tensorflow_datasets/scripts/templates/dataset.mako.md
Eshan-Agarwal: Sure

Eshan-Agarwal: This happens because I  replace the same file from one place to other although no changes in inside of file to solve this I have to revert back whole commits

Conchylicultor: I think you can also mark the file as branched afterward

Eshan-Agarwal: Should I also increase quality of images as I keep it to 30 now ?

tensorflow_datasets/scripts/generate_visualization.py
```diff
@@ -0,0 +1,62 @@
...
+
+def generate_vizz(data_name):
+    """Save the generated figures for the dataset
+    Args:
+      data_name: name of the dataset
+    """
+    try:
+        print("Generating examples...")
+        builder = tfds.builder(data_name)
+        builder.download_and_prepare()
```
Conchylicultor, line 41: As explained on the issue, the script should assume the datasets already exists (no call to `download_and_prepare`). Could you instead create a script like: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/document_datasets.py

Create a `def generate_visualization(datasets=None):` which use thread pool to generate the builder images in parallel.

tensorflow_datasets/scripts/generate_visualization.py
```diff
@@ -0,0 +1,62 @@
...
+FULL_PATH = tfds.core.get_tfds_path(FIG_DIR)
+
+def generate_vizz(data_name):
+    """Save the generated figures for the dataset
+    Args:
+      data_name: name of the dataset
+    """
+    try:
+        print("Generating examples...")
+        builder = tfds.builder(data_name)
```
Conchylicultor, line 40: You should create one image per-config.

tensorflow_datasets/scripts/generate_visualization.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+import os
+from absl import app
+
+import tensorflow as tf
+import tensorflow_datasets as tfds
+
+FIG_DIR = os.path.join('..', 'docs', 'catalog', 'images')
+FULL_PATH = tfds.core.get_tfds_path(FIG_DIR)
```
Conchylicultor, line 31: For the destination path, could you use add a `--dst_dir` flag instead ?

tensorflow_datasets/scripts/generate_visualization.py
```diff
@@ -0,0 +1,62 @@
...
+    Args:
+      data_name: name of the dataset
+    """
+    try:
+        print("Generating examples...")
+        builder = tfds.builder(data_name)
+        builder.download_and_prepare()
+        split = list(builder.info.splits.keys())[0]
+        data, data_info = tfds.load(data_name, split=split, with_info=True)
+        figure = tfds.show_examples(data_info, data)
```
Conchylicultor, line 44: Only the `tfds.show_examples` should be in try/except block

tensorflow_datasets/scripts/templates/dataset.mako.md
```diff
@@ -165,6 +165,31 @@ ${builder.info.citation}
...
+<%def name="display_figure(builder)">\
+<%
+  import requests
+  import os
+
+  config_name = os.path.split(builder.info.full_name)[-2]
+  ds_name_config = config_name.replace("\\", "/").replace("/", "_")
+
+  def dataset_examples_paths(ds_name):
+    github_path = "https://" + "github.com/Eshan-Agarwal/datasets/blob/patch-58/docs/catalog/images/" + ds_name + ".jpg"
```
Conchylicultor, line 177: You should not store images in your repository, nor use request. Instead files should be accessed with tf.io.gfile.exists

Conchylicultor: Also the path should be a constant passed in document_dataset.py

tensorflow_datasets/scripts/templates/dataset.mako.md
```diff
@@ -165,6 +165,31 @@ ${builder.info.citation}
 % endif
 </%def>
 
+<%def name="display_figure(builder)">\
+<%
+  import requests
+  import os
+
+  config_name = os.path.split(builder.info.full_name)[-2]
+  ds_name_config = config_name.replace("\\", "/").replace("/", "_")
```
Conchylicultor, line 174: Why is replace("\\", "/") required ? full_names returns the posix path.

Conchylicultor, line 174: Replace `.replace("/", "-")` to avoid potential colisions `my_dataset_xyz` `my_dataset/xyz`

tensorflow_datasets/scripts/templates/dataset.mako.md
```diff
@@ -212,6 +237,9 @@ def get_supervised(builder):
 def get_citation(builder):
   return builder.info.citation
 
+def get_figure(builder):
+  return builder.info.full_name
```
Conchylicultor, line 241: This should returns the same constant if none of the images exists. Otherwise, it will create multiple empty sections.

tensorflow_datasets/scripts/templates/dataset.mako.md
Eshan-Agarwal: okay

tensorflow_datasets/scripts/templates/dataset.mako.md
Eshan-Agarwal: images are stored in docs/catlog/images/ and here use this to check if file exists to show some results, I will correct this

tensorflow_datasets/scripts/generate_visualization.py
Eshan-Agarwal: In `document_dataset_examples.py` file I create a list of dataset having config or without config. From their I call this `generate_visualization.py` which will take dataset with configs and save image in dest_sir. Is it okay?

tensorflow_datasets/scripts/templates/dataset.mako.md
Eshan-Agarwal: I think of windows also on adding this but i think its not require make changes

tensorflow_datasets/scripts/document_dataset_examples.py
```diff
@@ -0,0 +1,75 @@
...
+        file.write(doc_builder)
+
+def dataset_configs_list():
+    """Returns list of datasets with configs."""
+    data_name_configs = []
+    for data_name in DATASET_TO_TESTS:
+        builder = tfds.builder(data_name)
+        if builder.BUILDER_CONFIGS:
+            for config in builder.BUILDER_CONFIGS:
+                data_name_configs.append(os.path.join(builder.name, config.name))
```
Conchylicultor, line 57: posixpath, or `'/'.join`, otherwise it will bug on windows

tensorflow_datasets/scripts/generate_visualization.py
Conchylicultor: The document_dataset_examples.py is nice for you to test, but won't be merged to TFDS.

We want a standalone script which generates the dataset config for all the datasets. similarly to 
 https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/document_datasets.py

So you should create a new function:
```
def generate_visualization(datasets=None):
```
which if None loads all datasets and configs from all datasets.

And then you can pass those names to your `generate_vizz` function (that you could rename `generate_single_visualization(ds_name)`)

Eshan-Agarwal: I am applying `ThreadPoolExecutor` to generate builders like this:

```py
with futures.ThreadPoolExecutor(max_workers=WORKER_COUNT_DATASETS) as tpool:
            builder_examples = tpool.map(generate_single_visualization, builders)
```
And then in `generate_single_visualization` I am making list of all datasets with or without config like
```py
def generate_single_visualization(builder):
    """Save the generated figures for the dataset
    Args:
      data_name: name of the dataset
    """
    dataset_config_list = []
    if builder.builder_configs:
        for config in builder.BUILDER_CONFIGS:
            dataset_config_list.append(os.path.join(builder.name, config.name))
    else:
        dataset_config_list.append(builder.name)
```
So just want to ask is this is okay or I have to generate configs also using `ThreadPoolExecutor` ?

tensorflow_datasets/scripts/templates/dataset.mako.md
Eshan-Agarwal: Sorry for trouble but I don't get you properly what that means `This should returns the same constant if none of the images exists.` . Like when dataset is of text ?

tensorflow_datasets/scripts/generate_visualization.py
Conchylicultor: I think the builder config should be instantiated inside the thread pool, as loading the builder may add some overhead that we want to paralellize.
Also I think we want to parallelize the configs from the same builder.

So the best would be to have something like:
```
tpool.map(generate_single_visualization, ['dataset0/config0', 'dataset0/config1', 'dataset1',...])
```

To generate the list of config names, the best would be to add a new function in `registered.py` like:

```python
def list_config_names(predicate_fn: Callable[Type[DatasetBuilder], bool] = None) -> List[str]:
  """Lists all registered `dataset/config` and `dataset` names."""
  # Strip the version from the full names
  config_names = {n.rsplit('/', maxsplit=1)[0] for n in list_full_names(predicate_fn)}
  return sorted(config_names)
```

tensorflow_datasets/scripts/templates/dataset.mako.md
Conchylicultor: I'm talking about: `github_path = "https://" + "github.com/Eshan-Agarwal/datasets/blob/patch-58/docs/catalog/images/`

You are hardcoding your own repo url, then use `request` to query the images. Instead you should use the `tf.io.gfile` API

tensorflow_datasets/scripts/templates/dataset.mako.md
Conchylicultor: Currently, if the dataset do not have visualization, it will create multiple empty sections:

```
# Dataset
## Config0
* Figure: No figure
## Config1
* Figure: No figure
```
With this change:
```py
def get_figure(builder):
  return builder.info.full_name if _visualization_exists(builder) else None
```
Only one section gets created, which avoid redundancy:
```
# Dataset
* Figure: No figure
## Config0
## Config1
```

tensorflow_datasets/scripts/generate_visualization.py
Eshan-Agarwal: Thanks for help, I will make changes, So after this `list_config_names` I have to pass this functions return list instead `['dataset0/config0', 'dataset0/config1', 'dataset1',...]`

tensorflow_datasets/scripts/templates/dataset.mako.md
Eshan-Agarwal: Yes its create multiple FIgure: No figure available I will make changes thanks

tensorflow_datasets/scripts/generate_visualization.py
```diff
@@ -23,40 +23,82 @@
...
 
-def generate_vizz(data_name):
-    """Save the generated figures for the dataset
     Args:
-      data_name: name of the dataset
+      data_name: name of the dataset with or without config for which
+      generate figure.
     """
+    print("Generating examples %s..." % data_name)
+    builder = tfds.builder(data_name)
```
Conchylicultor, line 48: Note: If the image already exists, we should skip the generation entirely (to avoid recomputing the same image twice.

tensorflow_datasets/scripts/generate_visualization.py
```diff
@@ -23,40 +23,82 @@
...
 import os
 from absl import app
+from absl import flags
+from concurrent import futures
 
 import tensorflow as tf
 import tensorflow_datasets as tfds
+from tensorflow_datasets.core import registered
 
+WORKER_COUNT_DATASETS = 200
```
Conchylicultor, line 33: 200 seems to big as I don't think 200 datasets can be loaded in parallel. Maybe set to 10 ?

tensorflow_datasets/scripts/generate_visualization.py
```diff
@@ -23,40 +23,82 @@
...
+    Args:
+        datasets: list of datasets for which to generate figures.
+              If None, then all available datasets will be used.
 
+    Returns:
+        List of dataset names with or without config.
+    """
+    dataset_config_list = []
+    if not datasets:
+        dataset_config_list = [x for x in registered.list_config_names()]
```
Conchylicultor, line 75: tfds.core.registered.list_config_names()

Conchylicultor, line 75: Why not `return tfds.core.registered.list_config_names()` ?

tensorflow_datasets/scripts/generate_visualization.py
Eshan-Agarwal: Actually I include this here only because of next line to it ie, `split = list(builder.info.splits.keys())[0]` as splits for different datasets are different

tensorflow_datasets/scripts/generate_visualization.py
Eshan-Agarwal: Okay I will change this

tensorflow_datasets/scripts/generate_visualization.py
Conchylicultor: I just meant to add somethng like:

```python
if tf.io.gfile.exists(dst_path):
  return  # If the image already exists, skip the image generation
builder = tfds.builder(data_name)
```

tensorflow_datasets/scripts/generate_visualization.py
Eshan-Agarwal: Yes this should be return direct my mistake I will correct this

tensorflow_datasets/scripts/generate_visualization.py
Eshan-Agarwal: okay this is good to include I will make changes

Eshan-Agarwal:
@Conchylicultor 

> Thank you, this looks better. Let me know when it's ready so I can merge this.

Thanks, Actually I need some time to fix an error everything works fine accept one in Last this error occurs which I think because of thread executor fails to stop error is [here ](https://gist.github.com/Eshan-Agarwal/d8096083b2aafb2ac431c9c68fa034a4). But all processes running fine although this error occur in last.

Eshan-Agarwal:
@Conchylicultor  Everything working fine previous error is nothing to do with scripts also I am running on colab which uses Ipython version 5.5.0 but after installing latest version of Ipython `pip install ipython==7.13.0`, no error occurs.

Conchylicultor:
This looks great. I started importing this and I'm currently generating all images in: https://pantheon.corp.google.com/storage/browser/tfds-data/visualization

This should be merged soon! Thanks again.

Issue #1945, opened 2020-04-24
[GSoC] Better lazy_imports usability


Conchylicultor:
**Part1**: Currently, lazy imports feels counter intuitive and confusing. It would be nice to improve the API to allow something like:

```py
with tfds.core.lazy_imports():
  import some_module
```
It would patch the default module import:
* If the module is installed (e.g. tested with `try/except ImportError`): Import the module as standard
* If the import isn't installed: Replace the module by a dummy ModuleType which raise error when access attribute:

```py
with tfds.core.lazy_imports():
  import some_module

some_module.some_function()  # Raise the lazy import error (please install some_module)
```

vijayphoenix:
Working on this!

PR #1947, opened 2020-04-24
Add lazy_importer context manager

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/1945


tensorflow_datasets/core/lazy_imports_lib_test.py
```diff
@@ -58,6 +58,12 @@ def test_bad_import(self):
     with self.assertRaisesWithPredicateMatch(ImportError, "extras_require"):
       _ = tfds.core.lazy_imports.test_foo
 
+  def test_lazy_import(self):
+    with self.assertRaisesWithPredicateMatch(ImportError, "extras_require"):
+      with tfds.core.lazy_importer():
+        import some_module
+
+      some_module.some_function()
```
Conchylicultor: I believe we should only accept whitelisted deps to be registered. Accepting any deps may be more difficult for the user to debug if he make a typo. Instead we should add some `_ALLOWED_LAZY_DEPS = ['matplotlib', ...]`
```py
with lazy_imports():
  import matplotlib  # Works
  import invalid_import  # Raise 'Unknown import {name}. If you believe this is correct, please add it to the whitelisted imports'
```

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -21,23 +21,47 @@
...
+class Fakemodule(object):
+
+  def __init__(self, module_name):
+    self.module_name = module_name
+
+  def __getattr__(self, _):
+    err_msg = _ERR_MSG.format(name=self.module_name)
+    raise ImportError(err_msg)
+
+class LazyImporterContextManager(object):
```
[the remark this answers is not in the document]
vijayphoenix: will do this soon

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -46,6 +82,36 @@ class LazyImporter(object):
...
+    if fullname in _ALLOWED_LAZY_DEPS:
+      return self
+    return None
+
+  def load_module(self,fullname):
+    if fullname not in sys.modules:
+      mod = imp.new_module(fullname) # TODO: Give FakeModule(fullname) instead
+      mod.__loader__ = self
+      sys.modules[fullname] = mod
+    return sys.modules[fullname]
```
vijayphoenix, line 101: @Conchylicultor Most of the part of the PR is completed.
However, I am running into infinite recursion when I try to load `FakeModule(fullname)` instead of a `imp.new_module(fullname)`.
Could you please guide me through this issue.

Thank you

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -21,23 +21,59 @@
...
-               "setup.py extras_require.").format(name=module_name)
+    err_msg = _ERR_MSG.format(name=module_name)
     utils.reraise(suffix=err_msg)
 
 
+class FakeModule(types.ModuleType):
+  """Create a fake module and raise ImportError whenever a attribute is accessed""" 
+  def __getattribute__(self, _):
+    err_msg = _ERR_MSG.format(name=self.module_name)
+    raise ImportError(err_msg)
```
Eshan-Agarwal, line 74: I think `return super(FakeModule, self).__getattribute__(_)` can avoid infinite recursion error

vijayphoenix: That won't be required as I am already raising an `ImportError`.
Program execution stops as soon as an exception is encountered.

Eshan-Agarwal: Okay Try this  replace `self.module_name` by `_name ` where `_name=` object.__getattribute__(self, 'module_name') `

vijayphoenix: Thanks for the help.
I used `_getattribute__` function instead of `_getattr__`.
Issue resolved. :smile:

tensorflow_datasets/core/lazy_imports_lib.py
vijayphoenix: Issue resolved

vijayphoenix:
@Conchylicultor please review the changes.
Thank you

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -46,6 +85,38 @@ class LazyImporter(object):
...
+  def load_module(self, fullname):
+    """Load a `FakeModule` if the requested module is not present in sys.modules"""
+    if fullname not in sys.modules:
+      mod = FakeModule(fullname)
+      mod.__loader__ = self
+      mod.__path__ = []
+      sys.modules[fullname] = mod
+    return sys.modules[fullname]
+
+  def __enter__(self):
```
Conchylicultor, line 108: Use contextlib.contextmanager with try/except.

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -154,3 +225,5 @@ def test_foo(cls):
 
 
 lazy_imports = LazyImporter  # pylint: disable=invalid-name
+sys.path_hooks.append(LazyImporter)
```
Conchylicultor, line 228: I don't think `LazyImporter` should be the same class as the hook. I would move all the hook logic inside `LazyImportHook` class

Conchylicultor, line 228: This adds the imported globally, while we only want to add it within the lazy context manager.

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -154,3 +225,5 @@ def test_foo(cls):
...
+sys.path.append("FAKE_PATH_TRIGGER")
```
Conchylicultor, line 229: Why are you adding this ? Please add comments when you write something non trivial which might be confusing for readers.

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -46,6 +85,38 @@ class LazyImporter(object):
...
+    return self
+
+  def __exit__(self, type_, value, _):
+    # Raise ImportError for modules not present in `_ALLOWED_LAZY_DEPS`
+    if isinstance(value, ImportError):
+      err_msg = ("Unknown import {name}. Currently lazy_imports does not "
+                 "support {name} module. If you believe this is correct, "
+                 "please add it to the list of `_ALLOWED_LAZY_DEPS` in "
+                 "`tfds/core/lazy_imports_lib.py`".format(name=value.name))
+      raise ImportError(err_msg)
```
Conchylicultor, line 118: use `utils.reraise(prefix=)` as we want to keep the original stacktrace.

tensorflow_datasets/core/lazy_imports_lib_test.py
```diff
@@ -58,6 +60,27 @@ def test_bad_import(self):
...
       _ = tfds.core.lazy_imports.test_foo
 
+  # pylint: disable=import-outside-toplevel
+  def test_lazy_import_context_manager(self):
+    with tfds.core.lazy_imports():
+      import pandas # pylint: disable=unused-import
+      import matplotlib # pylint: disable=unused-import
+
+    self.assertIn("pandas", sys.modules)
+    self.assertIn("matplotlib", sys.modules)
```
Conchylicultor, line 70: If you want to test the module was loaded properly, you should test something like `hasattr(matplotlib, 'function_name')`

tensorflow_datasets/core/lazy_imports_lib_test.py
```diff
@@ -58,6 +60,27 @@ def test_bad_import(self):
     with self.assertRaisesWithPredicateMatch(ImportError, "extras_require"):
       _ = tfds.core.lazy_imports.test_foo
 
+  # pylint: disable=import-outside-toplevel
+  def test_lazy_import_context_manager(self):
```
Conchylicultor, line 64: Thank you. To make sure the lazy importer isn't globally added, could you also add a test to make sure importing outside the lazy importer still works ?
```py
import nltk
self.assertTrue(hasattr(nltk, 'function'))
```
And that unknown import still fails, even when added to `_ALLOWED_LAZY_DEPS`.
```py
_ALLOWED_LAZY_DEPS.append('valid_module')
self.assertRaisesRegex(ImportError, '^Import error 'valid_module'$'):
  import valid_module
```

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,24 +20,98 @@
...
+  def __init__(self, name):
+    self.__path__ = None
+    super(FakeModule, self).__init__(name)
+
+  def __getattr__(self, attr):
+    err_msg = _ERR_MSG.format(name=self.__name__)
+    raise ImportError(err_msg)
+
+
+class LazyImporterHook(object):
```
Conchylicultor, line 88: If I understand the implementation correctly, if executing:

```py
with lazy_imports():
  import x
```
And that `x` internally imports `y`:
```py
# x/__init__.py
import y
```
I'm a little concerned that `y` gets lazy imported even if this is internal code that TFDS shouldn't modify.
Do you think it would be possible to make the importer such as only the top-most level gets lazy imported ?

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,24 +20,98 @@
...
+    "matplotlib",
+    "matplotlib.pyplot",
+    "mwparserfromhell",
+    "nltk",
+    "pandas",
+    "PIL_Image",
+    "PIL.TiffImagePlugin",
+    "pretty_midi",
+    "pydub",
+    "scipy",
```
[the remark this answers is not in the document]
Conchylicultor: Actually, whitelisting individual members seems safer to avoid typo.

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -46,6 +120,28 @@ class LazyImporter(object):
...
+  @staticmethod
+  @contextlib.contextmanager
+  def lazy_importer():
+    """Context Manager for lazy_imports.
+
+    Fake Module is created if the lazy import is not present in `sys.modules`.
+    A fake module is created only if module_name is present in `_ALLOWED_LAZY_DEPS`.
+    """
+    try:
+      sys.path_hooks.append(LazyImporterHook)
```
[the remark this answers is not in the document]
vijayphoenix: Although `sys.meta_path` is a more powerful, our needs can be fulfilled by `sys.path_hooks`

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,24 +20,98 @@
...
+    raise ImportError(err_msg)
+
+
+class LazyImporterHook(object):
+  """Finder and Loader for modules in `_ALLOWED_LAZY_DEPS`"""
+  # https://www.python.org/dev/peps/pep-0302/#specification-part-2-registering-hooks
+  # Each path_hook is called with one argument, the path item.
+  # It must raise ImportError if it is unable to handle the path item,
+  # and return an importer object if it can handle the path item
+  PATH = "FAKE_PATH_TRIGGER"
```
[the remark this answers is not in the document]
vijayphoenix: We can remove the constant, however, a path is passed to `__init__` of the class.
This path variable is given to `path_hook` just to raise an import error if that particular path item should not be handle by that particular loader.

I think we can simplify this in our case

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -46,6 +120,28 @@ class LazyImporter(object):
   lazily imported here.
   """
 
+  @staticmethod
+  @contextlib.contextmanager
+  def lazy_importer():
```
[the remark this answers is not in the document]
Conchylicultor: Actually, after more though, I think your first solution of creating a new symbol was the right one, but the name should be changed to `maybe_import()` or `try_import`. I think that lazy_import isn't a great name for this feature as the imports are not lazy.

```
with tfds.core.try_import():
  import abc
```

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,24 +20,98 @@
...
-               "manually installed (usually with `pip install {name}`). See "
-               "setup.py extras_require.").format(name=module_name)
+    err_msg = _ERR_MSG.format(name=module_name)
     utils.reraise(suffix=err_msg)
 
 
+class FakeModule(types.ModuleType):
+  """A fake module which raise ImportError whenever an unknown attribute is accessed"""
+  def __init__(self, name):
+    self.__path__ = None
```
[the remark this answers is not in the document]
vijayphoenix: Yeah, we don't need to change the value of `__path__`. Sorry, about that.

`__path__` variable was accessed by some `importlib` module functions. So I had to create it to prevent `__getattr__` from throwing any errors.

tensorflow_datasets/core/lazy_imports_lib.py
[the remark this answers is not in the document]
vijayphoenix: Sure, no problem. 
Reverting it will require only copy/pasting :smile:

tensorflow_datasets/core/lazy_imports_lib.py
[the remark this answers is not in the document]
vijayphoenix: So, no changes needed, right?

tensorflow_datasets/core/lazy_imports_lib.py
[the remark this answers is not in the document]
vijayphoenix: I was surprised to discover that ModuleType does not have `__path__`.
Is there a better parent `FakeModule` can inherit from?

tensorflow_datasets/core/lazy_imports_lib.py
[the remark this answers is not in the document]
vijayphoenix: Also, `import matplotlib` does not import `matplotlib.pylot`.
So, I thought it is better to whitelist modules like `matplotlib`

tensorflow_datasets/core/lazy_imports_lib.py
vijayphoenix: I will look into this.

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,22 +20,116 @@
...
+    "cv2",
+    "h5py",
+    "langdetect",
+    "librosa",
+    "matplotlib",
+    "matplotlib.pyplot",
+    "mwparserfromhell",
+    "nltk",
+    "pandas",
+    "PIL_Image",
```
Conchylicultor, line 49: `PIL_Image` Isn't a valid module name

tensorflow_datasets/core/lazy_imports_lib.py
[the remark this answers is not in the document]
Conchylicultor: Yes, no change needed.

tensorflow_datasets/core/lazy_imports_lib.py
Conchylicultor: After thinking about this, I think it may be better to patch the built-in `__import__()` directly.
* The importer hook implementation seems too complicated to correctly handle edge cases.
* We do not want to add the lazy imports to `sys.modules`, so it seems we're breaking some assumptions of the hook protocol

You can use `with py_utils.temporary_assignment(built_in, '__import__', new_import):` to patch temporary the built-in.

vijayphoenix: I can definitely do this.
However, if please review the new implementation using `sys.meta_path`.
If you think this is complicated, I will switch to the suggested method

Conchylicultor: The issue with `sys.meta_path` is that it won't handle correctly the nested case (lazy-imports inside the import). Also, the lazy imports shouldn't be registered in `sys.modules` as it might break others libs.

We should add a test to make sure that nested work correctly:

```py
with self.assertRaisesRegex(ImportError, 'y not found'):
  with maybe_imports():
    import x
```
With:
* ALLOWED_DEPS is x, y
* x is installed, but not y
* `x.py`:
```py
# x.py
import y
# y should raise ImportError, instead of being a dummy module
```

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,21 +20,93 @@
...
+    super(FakeModule, self).__init__(name)
+
+  def __getattr__(self, attr):
+    err_msg = _ERR_MSG.format(name=self.__name__)
+    raise ImportError(err_msg)
+
+
+def _custom_import(name, *args, **kwargs):
+  try:
+    return original_import_fun(name, *args, **kwargs)
```
Conchylicultor, line 88: original_import_fun should be the `__import__` function at the time the `with try_import` is called. Not as import time.
Otherwise it might creates issue if other libraries also try to patch `__import__`:

```
with user_custom_import():
  with tfds.core.try_import():
    import x
  import y  # Oups, use the built-in import instead of the user_custom_import
```

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,21 +20,93 @@
...
+  except ImportError:
+    if name in _ALLOWED_LAZY_DEPS:
+      return FakeModule(name)
+    raise ImportError
```
Conchylicultor, line 92: `raise ImportError` -> `raise`. We want to forward the original error message and stacktrace

tensorflow_datasets/core/lazy_imports_lib.py
```diff
@@ -20,21 +20,93 @@
...
+
+
+@contextlib.contextmanager
+def try_import():
+  """Context Manager for lazy_imports.
+
+  A fake module is created if the lazy import is not present in `sys.modules`.
+  It is created only if module_name is present in `_ALLOWED_LAZY_DEPS`.
+  """
+  try:
```
Conchylicultor, line 102: The `try/except` statement should go inside the `with` statement.

tensorflow_datasets/core/lazy_imports_lib_test.py
```diff
@@ -58,6 +58,48 @@ def test_bad_import(self):
...
+
+    get_deps = "tensorflow_datasets.core.lazy_imports_lib._ALLOWED_LAZY_DEPS"
+    with mock.patch(get_deps, ["new_module"]):
+      with tfds.core.try_import():
+        import new_module
+
+      with self.assertRaisesWithPredicateMatch(ImportError, "extras_require"):
+        new_module.some_function()
+
+  def test_nested_imports(self):
```
Conchylicultor, line 90: This test is good, so we should keep it. However this doesn't test nested import. Importing `matplotlib` do not try to import `matplotlib.pyplot`.

We should add a test to make sure that nested work correctly:
```
with self.assertRaisesRegex(ImportError, 'y not found'):
  with tfds.core.maybe_imports():
    import x
```
With:
* ALLOWED_DEPS is x, y
* x is installed, but not y
* x.py:
```
# x.py
import y
# y should raise ImportError, instead of being a dummy module
```
To test this, you'll have to create a dummy `x.py` file.

Basically this test that if `x` is installed but `import x` fails, then adding `with tfds.core.maybe_imports()` should fail too.

vijayphoenix: Done.
I was not sure where to create the `dummy_module_x.py`, so created it in the same directory.

PR #1950, opened 2020-04-25
update generate_visualization to use multiprocessing

### LINKED ISSUE #1949: [GSoC] Fix documentation images
https://github.com/tensorflow/datasets/pull/1948 is adding visualisation to the documentation, however it seems that images have visual artefacts (missing images, bad labels,...).
It is likely due to the fact that matplotlib do not support multi-threading. We should investigate if the `scripts/generate_visualization.py`  can use multi-processing instead of multi-processing.

Eshan-Agarwal:
fixes #1949


Eshan-Agarwal:
@Conchylicultor Please look on the changes :)

Conchylicultor:
Thanks for the fix. I've already worked on a similar fix internally which should appear externally in a few days, so closing this.

Issue #1946, opened 2020-04-24
[GSoC] Add a dummy beam wrapper


Conchylicultor:
Currently, implementing a beam dataset is quire painful due to lazy imports, which makes it impossible to use `beam.DoFn`,... in the global scope.
It would be nice to be able to implement beam datasets without worrying about declaring beam methods only in the main scope.

By replacing the `beam` module by a dummy no-op module, it would not crash when importing tensorflow datasets. It would greatly improve the usability:

```py
try:
  import appache_beam as beam
except ImportError:
  beam = MyDummyBeam

class SomeFn(beam.DoFn):
  pass

@beam.ptransform_fn
@beam.typehints.with_input_types(beam.Pipeline)
@beam.typehints.with_output_types(Union[tf.train.Example, bytes])
def some_ptransform_fn():
  pass
```
* If beam is installed, everything works as expected.
* If beam isn't found, then the imports works, but calling `SomeFn()`, `some_ptransform_fn()`,... raise the lazy import error ('Please install appache_beam')

Once https://github.com/tensorflow/datasets/issues/1945 is done, the `try/except` could be replaced by:
```py
with tfds.core.lazy_imports():
  import appache_beam as beam


class SomeFn(beam.DoFn):
  pass

@beam.ptransform_fn
def some_ptransform_fn():
  pass
```

Note: Not all the beam API has to be implemented. New methods can be added on the go as they are needed.

Eshan-Agarwal:
Working on it, can you please help me with few things which I don't get :

We have to implement `DummyBeam` class in `dummy_beam.py` which can be used as described in code above? or is above code is hint for `DummyBeam` ? or can be used to test after implementing `DummyBeam?`.

Sorry for trouble.

Conchylicultor:
> We have to implement DummyBeam class in dummy_beam.py

Yes. The code snippet above is an example of `DummyBeam` dataset usage.

PR #1951, opened 2020-04-25
Add dummy beam wrapper

Eshan-Agarwal:
Issue : #1946

ToDO: Add tests

### DIFF
--- a/tensorflow_datasets/core/utils/dummy_beam.py
+++ b/tensorflow_datasets/core/utils/dummy_beam.py
@@ -1,0 +1,39 @@
+# coding=utf-8
+# Copyright 2020 The TensorFlow Datasets Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Lint as: python3
+
+
+_MSG = ("Failed importing {name}. Please install {name}."
+                  " Using pip install {name}")
+
+
+class DoFn(object):
+   def __call__(self, *args, **kwargs):
+       return None
+  
+class Pipeline(object):
+  def __call__(self, *args, **kwargs):
+       return None
+
+
+class DummyBeam(object):
+
+    def __init__(self):
+      self.module_name = "apache_beam"
+      
+    def __getattr__(self, _):
+      err_msg = _MSG.format(name=self.module_name)
+      raise ImportError(err_msg)
--- a/tensorflow_datasets/core/utils/__init__.py
+++ b/tensorflow_datasets/core/utils/__init__.py
@@ -23,4 +23,5 @@ from tensorflow_datasets.core.utils.tf_utils import *
 from tensorflow_datasets.core.utils.tqdm_utils import *
 from tensorflow_datasets.core.utils.version import Experiment
 from tensorflow_datasets.core.utils.version import Version
+from tensorflow_datasets.core.utils.dummy_beam import DummyBeam
 # pylint: enable=wildcard-import


tensorflow_datasets/core/utils/dummy_beam.py
```diff
@@ -0,0 +1,39 @@
...
+# limitations under the License.
+
+# Lint as: python3
+
+
+_MSG = ("Failed importing {name}. Please install {name}."
+                  " Using pip install {name}")
+
+
+class DoFn(object):
```
Conchylicultor, line 23: Rather than duplicating this code, how about creating a single `Empty` class and use `DoFn = Empty`,...

tensorflow_datasets/core/utils/dummy_beam.py
```diff
@@ -0,0 +1,39 @@
...
+   def __call__(self, *args, **kwargs):
+       return None
+  
+class Pipeline(object):
+  def __call__(self, *args, **kwargs):
+       return None
+
+
+class DummyBeam(object):
```
Conchylicultor, line 32: `DoFn`,... should be accessed from the `DummyBeam`.

I think this should be inherited from `types.ModuleType`.

```
class DummyBeam(types.ModuleType):
  DoFn = Empty
  Pipeline = Empty
```

Eshan-Agarwal:
@Conchylicultor Please have a look on changes.

tensorflow_datasets/core/utils/dummy_beam.py
```diff
@@ -0,0 +1,44 @@
...
+class DummyBeam(types.ModuleType): # pylint: disable=too-few-public-methods
+    """Dummy class.
+    Raise:
+      ImportError, when calling beam API and apache_beam was not installed.
+    """
+    DoFn = Empty
+    Pipeline = Empty
+
+    def __init__(self):
+        self.module_name = "apache_beam"
```
vijayphoenix, line 38: `ModuleType` already has a `__name__` variable, consider using that instead of `self.module_name`

tensorflow_datasets/core/utils/dummy_beam.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def __getattribute__(self, _):
+        if getattr(DummyBeam, _, None) is Empty:
+            _name = super().__getattribute__('module_name')
+            err_msg = _MSG.format(name=_name)
+            raise ImportError(err_msg)
```
vijayphoenix, line 44: Define `__getattr__` instead of `__getattribute__`.

tensorflow_datasets/core/utils/dummy_beam.py
Eshan-Agarwal: Yeah sure thanks for catching this :-)

tensorflow_datasets/core/utils/dummy_beam.py
Eshan-Agarwal: `__getattr__` make some trouble here because we want `DoFn ` to call with DummyBeam so if use `getattr `here this will executing without any errors here, the reason why I am using `__getattribute__` is `__getattribute__` calls no matter what unlike `getattr `

tensorflow_datasets/core/utils/dummy_beam.py
Eshan-Agarwal: Thanks but I think `self.module_name` is require otherwise in error message it gives `None ` instead

tensorflow_datasets/core/utils/dummy_beam.py
vijayphoenix: I don't think we should throw an error when `DoFn` is accessed.

> If beam isn't found, then the imports works, but calling SomeFn(), some_ptransform_fn(),... raise the lazy import error ('Please install appache_beam')

We don't want this to happen. So, we should not throw any errors when members of `DummyBeam` are accessed.

However, proceed only after consulting @Conchylicultor. 
(Just in case I am wrong :sweat_smile:)

tensorflow_datasets/core/utils/dummy_beam.py
vijayphoenix: So, there is no point in inheriting from `ModuleType` ;-)

tensorflow_datasets/core/utils/dummy_beam.py
Eshan-Agarwal: Beam API have DoFn and pipiline but it dont have SomeFn its just name i think thats why it inherit beam.DoFn so i think it is good to get error when invoking beam.DoFn 🙂

tensorflow_datasets/core/utils/dummy_beam_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_dummy_beam(self):
+        """Test for DummyBeam."""
+        _err_msg = ("Failed importing apache_beam."
+                    " Please install apache_beam. Using pip install apache_beam")
+        try:
+            import apache_beam as beam
+        except ImportError:
+            beam = dummy_beam.DummyBeam()
+        with self.assertRaisesWithPredicateMatch(ImportError, _err_msg):
```
Conchylicultor, line 34: Only add the `with self.assertRaisesWithPredicateMatch(ImportError, _err_msg):` around the line which actually raise the exception.
Otherwise I don't know where the exception is coming from.

tensorflow_datasets/core/utils/dummy_beam_test.py
```diff
@@ -0,0 +1,42 @@
...
+            beam = dummy_beam.DummyBeam()
+
+            class SomeFn(beam.DoFn): # pylint: disable=unused-variable, too-few-public-methods
```
Conchylicultor, line 37: Creating the class should works, which is the entire point.
 but instantiated the class should fail.:
```py
class SomeDoFn(beam.DoFn):  # Works
  pass

with self.assertRaisesRegex(...):
  SomeDoFn()  # Raise error
```

tensorflow_datasets/core/utils/dummy_beam.py
Conchylicultor: Set the name calling `__init__` in super

tensorflow_datasets/core/utils/dummy_beam.py
Conchylicultor: @vijayphoenix is right. The all point of this module is to allow using beam without raising error when declaring the beam functions. However error should be raised when the functions are called.

Eshan-Agarwal: I think Using getattr instead getattribute is also not worked as getattr calls only when it don't fine DoFn or Pipeline which is called insider class DummyBeam, can you suggest some other way ?

vijayphoenix: Add some like 
```
class Empty():
    def __getattr__(self, _)
        raise Error
```

Eshan-Agarwal: Thanks but this won't works 🙂

Eshan-Agarwal: However I found a way to do this and its working when  I call SomeFn it raises error, is it necessary to use `types.ModuleType` ?

Conchylicultor: You shouldn't have to use `__new__`. `__init__` should be enough.

What is the issue with `types.ModuleType` ?

Conchylicultor: Modules should inherit from `ModuleType`, but not other Class or functions.


```
class EmptyClass(object):
  def __init__(self):
    raise ...
```

Eshan-Agarwal: okay then doing in this way ?
```py
class Empty(object):
   
    def __init__(self, *args, **kwargs):
        err_msg = _MSG.format(name="apache_beam")
        raise ImportError(err_msg)
        
    
class DummyBeam(types.ModuleType):

    DoFn = Empty
    Pipeline = Empty
    ptransform = Empty

    def __init__(self):
        return None
```
what is use of `types.ModuleType` here ?

PR #1954, opened 2020-04-25
Improve GitHub workflow

vijayphoenix:
* Update `pull_request_template.md`
* Add new `documentation` issue template
* Add `.vscode` folder to `.gitignore`. [Just for my convenience :sweat_smile: ]

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation-issue.md
+++ b/.github/ISSUE_TEMPLATE/documentation-issue.md
@@ -1,0 +1,32 @@
+---
+name: Documentation Issue
+about: Use this template for documentation related issues
+title: ''
+labels: documentation
+assignees: ''
+
+---
+
+**URL(s) with the issue**
+
+Please provide a link to the documentation entry, for example:
+
+**Description of issue**
+
+A clear and concise description of what the problem is.
+
+Ex. Dataset download link not working or update dataset citation.
+
+**Describe the solution you'd like**
+
+A clear and concise description of what you want to happen.
+
+**Additional context**
+
+Add any other context about the issue.
+
+**Submit a pull request?**
+
+Are you planning to also submit a pull request to fix the issue?
+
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/documentation-issue.md
```diff
@@ -0,0 +1,32 @@
+---
```
Conchylicultor, line 1: What are you expecting this to be used for ?
I think documentations issues could go into other category (bug report or feature request).
Should we improve the other template documentation instead ?

.github/pull_request_template.md
```diff
@@ -16,5 +16,7 @@
 * [ ] Properly cite in `BibTeX` format
 * [ ] Add passing test(s)
 * [ ] Add test data
+* [ ] Use `lazy_imports` to access external dependencies (if applicable)
+* [ ] Update `setup.py`, `LazyImporter`, and `LazyImportsTest` with dependencies (if applicable)
```
Conchylicultor, line 20: Rather than naming some symbols which may be obscure for new contributors, I would instead only add a single item with a link to the doc.
```
* [ ] If using additional dependencies (e.g. `nltk`), use [lazy_imports](https://www.tensorflow.org/datasets/add_dataset#extra_dependencies)
```

.github/ISSUE_TEMPLATE/documentation-issue.md
vijayphoenix: It could be used for the following documentation problems:
* Vague; not clear; Incomplete; Misleading; Outdated API documentation
* Any spelling mistake
* Any update in dataset description, `home_url`, `citation`, etc.

In my opinion, the above problems could be labeled as `documentation` issue. This will also help us to filter the real bugs.
Or as you suggested we could add a new `other` issue template.

Conchylicultor: Ok sounds good.

PR #1956, opened 2020-04-26
Update Chexpert to load only lateral, frontal or both kinds of radiographs

tzapzoor:
# Update dataset

* Dataset Name: Chexpert
  
## Description

I've updated the Chexpert dataset to be able to load only frontal, lateral or both kinds of radiographs as the models used in the paper [1] are trained on either frontal or lateral images. With the current implementation the only solution to obtain frontal / lateral only data is to filter the dataset by using the image path which is kind of a hack.

[1] http://arxiv.org/abs/1901.07031

## Checklist
* [x] Address all TODO's
* [x] Add passing test(s)
* [x] Add test data
* [x] Lint code


tzapzoor:
@googlebot I signed it!

tzapzoor:
> Thank you for your contribution.
> Rather than creating three new configs, how about adding a new `ClassLabel(names=['frontal', 'lateral'])` to feature dict ? This would make it easier to switch between the images, without having to generate three times the same dataset.

Sure. It makes more sense. I'll do what you suggested instead.

tzapzoor:
The `.csv` files needed to contain two entries such that both frontal and lateral view images are tested and I also read somewhere in the docs that the image path is used for partitioning the dataset and, thus, need to be unique <= this is why :smile:

PR #2052, opened 2020-05-24
Fix error message of tfds mocking.

mohitreddy1996:
Edit error message. Previously: 

ValueError: TFDS has been mocked, but metadata files were not found in '...'. You should copy the real metadata files, so that the dataset can be loaded properly, or `set the data_dir kwarg oftfds.testing.mock_tfds(data_dir=...)`.


Conchylicultor:
Closing this as this was fixed internally.

Issue #2048, opened 2020-05-21
[data request] Deepmind's PG-19


afrozenator:
* Name of dataset: PG-19
* URL of dataset: https://github.com/deepmind/pg19
* License of dataset: Apache License 2.0
* Short description of dataset and use case(s): This repository contains the PG-19 language modeling benchmark. It includes a set of books extracted from the Project Gutenberg books library, that were published before 1919. It also contains metadata of book titles and publication dates. PG-19 is over double the size of the Billion Word benchmark and contains documents that are 20X longer, on average, than the WikiText long-range language modelling benchmark.

https://github.com/deepmind/pg19

Folks who would also like to see this dataset in `tensorflow/datasets`, please thumbs-up so the developers can know which requests to prioritize.

And if you'd like to contribute the dataset (thank you!), see our [guide to adding a dataset](https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md).

afrozenator:
@lukaszkaiser - please upvote?

lukaszkaiser:
It's a very nice dataset, would be great to have it on TFDS for more impact!

acharles7:
@afrozenator @lukaszkaiser Can I work on this?

lukaszkaiser:
Sure!! Let us know if we can help in any way!

acharles7:
Thanks @lukaszkaiser 
Is it fine if I put this data on the text section of the dataset?
Do we only need metadata for this dataset? because if not then this dataset is of > 10GB.
Please let me know what you really want in this dataset
Thanks.

lukaszkaiser:
I think we want the whole dataset - with all the data and all splits, right? So anyone using TFDS can run experiments with this dataset same as with any other language modelling one. I guess so at least - I think that'd be the most beneficial for TFDS and all users, or am I missing something?

acharles7:
Thank you, I'm working on it. I'll let you know if I have any queries. I will make PR soon.

lukaszkaiser:
Great news, looking forward to it!! :)

afrozenator:
Hi @acharles7 -- In TFDS you write to code to process the dataset and store as tf-records, while you are iterating, maybe you can just process one or two files from the train split and one or two from validation split etc just to make things fast on your end.

In the eventual PR you can remove these constraints and then anyone who needs it can build the dataset on their own.

Let me know if you have questions!

And thanks for taking it up!

afrozenator:
[...]

PR #2066, opened 2020-05-30
Added PG-19 dataset

acharles7:
# Add Dataset

* Dataset Name: PG-19 Language Modelling Dataset
* Issue Reference: #2048
* `dataset_info.json` Gist: [Link](https://gist.github.com/acharles7/feb2c7253fec7275340580a222d7fa1f)
## Description
This dataset contains the PG-19 language modeling benchmark. It includes a set of books extracted from the Project Gutenberg books project (https://www.gutenberg.org), that was published before 1919. It also contains metadata of book titles and publication dates. PG-19 is over double the size of the Billion Word benchmark and contains documents that are 20X longer, on average, than the WikiText long-range language modeling benchmark.

Dataset link: [Deepminds's PG-19 Dataset](https://github.com/deepmind/pg19)
## Checklist
* [X] Address all TODO's
* [X] Add alphabetized import to subdirectory's `__init__.py`
* [X] Run `download_and_prepare` successfully
* [X] Properly cite in `BibTeX` format
* [X] Add passing test(s)
* [X] Add test data
* [X] [Lint](https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style) code


tensorflow_datasets/text/pg19_test.py
```diff
@@ -0,0 +1,35 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow_datasets.public_api as tfds
+from tensorflow_datasets.text import pg19
+
+class Pg19Test(tfds.testing.DatasetBuilderTestCase):
+  DATASET_CLASS = pg19.Pg19
+  SPLITS = {
+      "train": 28602,   # Number of fake train example
```
Conchylicultor, line 29: You are generating the full dataset every time the test run, rather than generating only the fake data. The reason is that you're reading from `_DATA_DIR`

tensorflow_datasets/text/pg19.py
```diff
@@ -0,0 +1,124 @@
...
+            name=tfds.Split.TEST,
+            gen_kwargs={
+                "metadata": metadata_dict,
+                "filepath": os.path.join(_DATA_DIR, "test")},
+        ),
+    ]
+
+  def _generate_examples(self, filepath, metadata):
+    """Yields examples."""
+    for index, file in enumerate(tf.io.gfile.listdir(filepath)):
```
Conchylicultor, line 112: `listdir` is not deterministic, so generating the dataset twice on different system may lead different keys for the same examples. Instead, you could use `book_id` or similar

tensorflow_datasets/text/pg19_test.py
acharles7: Yes, you're right. 
But If I'm not reading data from `_DATA_DIR` then how am I supposed to read data from to pass tests?
Any suggestion? Thank you

Conchylicultor: You can patched the `_DATA_DIR` to use the fake path. See `wiki40b_test` as an example.

I've applied those changes internally, so you don't have to do it.

Conchylicultor:
Thanks for your contribution. Merging this now.

acharles7:
So Is there any changes you require from my side? @Conchylicultor

tensorflow_datasets/text/pg19_test.py
acharles7: Ok

tensorflow_datasets/text/pg19.py
afrozenator: Or just enumerate(sorted(...)) ?

acharles7: Why do you need it in sorted order?
we can do something like  this 
` enumerate(sorted(tf.io.gfile.listdir(path), key=lambda l: int(l.rstrip('.txt'))))`
any thoughts?

Issue #2067, opened 2020-06-01
New Image Folder dataset


vijayphoenix:
**Is your feature request related to a problem? Please describe.**
Currently, TFDS has a [tfds.image.ImageLabelFolder](https://www.tensorflow.org/datasets/catalog/image_label_folder) class which creates a dataset from a folder of images. However, the current API feels confusing and boilerplate.

**Describe the solution you'd like**

We should create a new `ImageFolder` class that should not pre-process the data to TF-Record.
It would be small util which dynamically loads their images from a folder (No preprocessing, no tf-record, no splits, no versioning, and all other stuff).

A probable new API could be
```
builder = tfds.ImageFolder('/path/to/images')  # Extract the labels,...
ds = builder.as_dataset()
```

ImageFolder would not inherit from `tfds.core.GeneratorBasedBuilder`, but directly from `tfds.core.DatasetBuilder.` `FeatureConnectors`, `ClassLabel` could still dynamically be loaded at start time, but not used for decoding.

**Additional Context**

Some point to consider:
- Support for split API? 
We should at least allow the user to define Train, Val, Test splits (which will be automatically deduced by the directory structure of ImageFolder).
- What output ? `{'image', 'label', 'image/filename'}` ? 
Currently, ImageLabelFolder is a part of image_classification datasets. So, `label` must be one of the outputs. We can also add a more generic ImageFolder class. Here the output could be `{'image', 'image/filename'}`
- Support for `as_supervised`?
- Support for decode API?

vijayphoenix:
PR https://github.com/tensorflow/datasets/pull/2068 got merged internally

PR #2068, opened 2020-06-01
Support for Custom Image Datasets

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/2067


vijayphoenix:
CC: @Conchylicultor

Conchylicultor:
Thank you. For the future, it might be better to write either a google doc, or to submit the proposal as a PR markdown in `tensorflow/datasets/docs/proposal/my_proposal.md`, or directly write the proposal in the github issue.

For this PR, a few though:

- For the API, it might be easier if the dataset didn't inherit from DatasetBuilder at all.
- I don't think that the API should rely on manual dir.
- I don't think there should be `_download_and_prepare`.
- Keeping the split seems good for now as it might simplify the DatasetInfo and would allow checking between splits.

vijayphoenix:
A few doubts:

- I think it would be better if we provide `ImageLabelFolder` class as a public API rather than adding it with other Image Datasets.
- `builder` argument is required for creating `DatasetInfo`. So, should we provide `DatasetInfo` as we are not inheriting from `DatasetBuilder`.

tensorflow_datasets/image_classification/image_folder.py
```diff
@@ -29,147 +29,72 @@
...
-
-    # The splits correspond to the high level folders
-    split_names = list_folders(dl_manager.manual_dir)
-
-    # Extract all label names and associated images
-    split_label_images = {}  # dict[split_name][label_name] = list(img_paths)
+class ImageLabelFolder():
+
+  def __init__(self, dataset_name, data_dir=None):
+    self._name = dataset_name
```
Conchylicultor, line 39: I don't think the ImageFolder need a name

tensorflow_datasets/image_classification/image_folder.py
```diff
@@ -29,147 +29,72 @@
...
+      raise ValueError("Split name {} not present in {}".format(split_name, self._split_label_images.keys()))
 
+    imgs = [img for l in self._split_label_images[split_name].values() for img in l]
+    list_ds = tf.data.Dataset.list_files(imgs)
+    labeled_ds = list_ds.map(process_path, num_parallel_calls=AUTOTUNE)
+    return labeled_ds
+
+def decode_img(img):
+  img = tf.image.decode_jpeg(img, channels=3)
+  return tf.image.convert_image_dtype(img, tf.float32)
```
Conchylicultor, line 84: For consistency with other datasets, I don't think we should convert image to float, but keep uint8

tensorflow_datasets/image_classification/image_folder.py
```diff
@@ -29,147 +29,72 @@
...
-    """Returns SplitGenerators from the folder names."""
-    # At data creation time, parse the folder to deduce number of splits,
-    # labels, image size,
-
-    # The splits correspond to the high level folders
-    split_names = list_folders(dl_manager.manual_dir)
-
-    # Extract all label names and associated images
-    split_label_images = {}  # dict[split_name][label_name] = list(img_paths)
+class ImageLabelFolder():
```
Conchylicultor, line 36: After thinking more about it, it might be best to have `ImageFolder` inherit from `tfds.core.GeneratorBasedBuilder` but we should list all caveat/pain point to make this works properly and found the best implementation:

* It shouldn't be added to the register (maybe useing the `registered.skip_registration` ?).
* The __init__ signature should be different (only contains `data_dir=`)
* Data dir shouldn't be `DATA_DIR/dataset_name/config/version` but the given version.
* `download_and_prepare` should raise an error if called.
* `DatasetInfo`  fields & cie are dynamically computed in `__init__`
* Other ?

For the implementation, it seems there are three options:

1. Do not inherit at all from `DatasetBuilder`
2. Inherit from `DatasetBuilder` and hack around the API.
3. Split DatasetBuilder in two class `DatasetBase` and `DatasetBuilder`

Without trying, it's difficult to know which one of those solutions is the best. My feeling would be to try 2 first and see what are the hack required (overwrite `__init__`,...). If it is really too hard, eventually prototype what would need to split `DatasetBase` and `DatasetBuilder` in 2 class and have `DatasetBuilder(DatasetBase)` and `ImageFolder(DatasetBase)`.

tensorflow_datasets/image_classification/image_folder.py
```diff
@@ -29,147 +29,72 @@
...
-        key = "%s/%s" % (label, os.path.basename(image_path))
-        yield key, {
-            "image": image_path,
-            "label": label,
-        }
+  def _as_dataset(self, split_name):
+    if split_name not in self._split_label_images.keys():
+      raise ValueError("Split name {} not present in {}".format(split_name, self._split_label_images.keys()))
 
+    imgs = [img for l in self._split_label_images[split_name].values() for img in l]
```
Conchylicultor, line 77: The list of images should be shuffled deterministically. (So that two users have examples in the same order).

When `shuffle_files=True`, there should be an additional `ds = ds.shuffle(len(image_paths))` after the `tf.data.Dataset.list_files`.

tensorflow_datasets/image_classification/image_folder.py
vijayphoenix: Okay, sure.

tensorflow_datasets/image_classification/image_folder.py
vijayphoenix: Just for clarification, should I try this first 
* `ImageFolder` inherit from `tfds.core.GeneratorBasedBuilder`
or this
* Inherit from `DatasetBuilder` and hack around the API?

Conchylicultor: I meant inherit from `DatasetBuilder`, not `GeneratorBasedBuilder`. Sorry for the confusion. I updated the text

vijayphoenix: > Data dir shouldn't be DATA_DIR/dataset_name/config/version but the given version.

Data dir should be `DATA_DIR/version` or `DATA_DIR` or something else?

Conchylicultor: Data dir should be the path where the images are located. There is no version.
If images are located in: `path/to/images/train/img0001.jpeg`, then data_dir should be `path/to/images/`

Conchylicultor: Yes, there shouldn't be any copy. Just read the plain images.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,131 @@
...
+        builder=self,
+        description="Generic image classification dataset.",
+        # Generic features before the data is generated
+        features=core.features.FeaturesDict({
+            "image": core.features.Image(), # TODO: Image shape
+            "label": core.features.ClassLabel(names=list(self._labels)),
+        }),
+        supervised_keys=("image", "label"),
+    )
+    ds_info.update_splits_if_different(self._split_dict)
```
Conchylicultor, line 90: Can you remove `self._split_dict` and add `self.info. update_splits_if_different()` in `__init__` ?

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,131 @@
...
+    self._split_label_images, self._labels = self._get_split_label_images()
+    self._splits_info = core.SplitDict(self.name)
+
+    # Calculate splits for DatasetInfo
+    self._splits = {
+        split_name: self._get_split_size(split_name)
+        for split_name in self._split_label_images
+    }
+
+    self._split_dict = core.SplitDict(self.name)
```
Conchylicultor, line 52: We should try avoid adding too many new `self._` attributes. Some seems to duplicate information (e.g. `self._splits_dict` VS `self.info.splits` VS `self._splits`).
I think they could be merged to only use `self.info.splits`.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,131 @@
...
+    return ds_info
+
+  def _download_and_prepare(self, **kwargs):
+    raise Exception("No need to call download and prepare")
+
+  def download_and_prepare(self, **kwargs):
+    raise Exception("No need to call download and prepare")
+
+  def _as_dataset(self, split, shuffle_files=False, ** kwargs):
+    if split not in self._split_label_images.keys():
```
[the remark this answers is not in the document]
vijayphoenix:
```
builder = tfds.ImageLabelFolder(path)
ds = builder.as_dataset(split="train")
```
We won't have this split argument in `__init__`

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,131 @@
...
+import tensorflow.compat.v2 as tf
+from tensorflow_datasets import core
+
+_SUPPORTED_IMAGE_FORMAT = ('.jpg', '.jpeg', '.png')
+_AUTOTUNE = tf.data.experimental.AUTOTUNE
+
+
+class ImageLabelFolder(core.DatasetBuilder):
+
+  VERSION = core.Version(
```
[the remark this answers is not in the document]
vijayphoenix: We get the following error
```
AssertionError: DatasetBuilder image_label_folder does not have a defined version. Please add a `VERSION = tfds.core.Version('x.y.z')` to the class.
```

tensorflow_datasets/core/custom_dataset/image_folder.py
vijayphoenix: Removed the `self._` and made them local variables.
I think they could be merged to only use `self.info.splits` because they are class properties.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,131 @@
...
+        core.SplitInfo(name=split_name, shard_lengths=[size])) # is this correct?
+
+    super(ImageLabelFolder, self).__init__(
+        data_dir=data_dir, version=str(self.VERSION), config=None)
+
+    # _data_dir should not change to DATA_DIR/Version
+    self._data_dir = os.path.expanduser(data_dir)
+
+  def _get_split_label_images(self):
+    split_names = _list_folders(self.data_dir)
```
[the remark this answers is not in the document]
vijayphoenix: Now `_list_folder` and `_list_imgs` return sorted lists

tensorflow_datasets/core/custom_dataset/image_folder.py
vijayphoenix: Done

vijayphoenix:
TODO: Add typing annotations

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -20,30 +20,36 @@
...
+
+def new_init(self, root_dir=None, **kwargs):
+  assert root_dir is None
+  del kwargs
+  root_dir = os.path.join(
+      tfds.core.utils.tfds_dir(), 'testing',
+      'test_data', 'fake_examples', 'image_label_folder')
+  original_init(self, root_dir=root_dir)
+
+tfds.ImageLabelFolder.__init__ = new_init
```
Conchylicultor, line 37: Logic like this should be moved in `setUpClass(cls)` inside `ImageLabelFolderTest ` (and make sure to call `super().setUpClass()`)

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,168 @@
...
+                        for l in self._split_label_images[split_name].values())
+        for split_name in self._split_label_images
+    }
+
+    split_dict = core.SplitDict(self.name)
+    for split_name, size in splits.items():
+      split_dict.add(
+          core.SplitInfo(name=split_name, shard_lengths=[size]))
+
+    self._ds_info.update_splits_if_different(split_dict)
```
Conchylicultor, line 95: Can't `self.info` be called instead ?

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,168 @@
...
+  VERSION = core.Version('2.0.0')
+
+  def __init__(self, root_dir):
+    self._version = str(self.VERSION)
+    self._root_dir = os.path.expanduser(root_dir)
+
+    # dict[split_name][label_name] = list(img_paths)
+    self._split_label_images, labels = _get_split_label_images(self._root_dir)
+
+    self._ds_info = core.DatasetInfo(
```
Conchylicultor, line 70: I would move this back in `_info` to avoid duplication between `self.info` and `self._ds_info`

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,168 @@
...
+        output_shapes=self.info.features.shape,
+        output_types=self.info.features.dtype,
+        args=[imgs])
+
+    if shuffle_files:
+      ds = ds.shuffle(len(imgs))
+    return ds
+
+
+def _get_split_label_images(
```
Conchylicultor, line 139: Could you add unittest for this function ? Note I think that you can use `tfds.testing.MockFs()` to avoid having to create real files/directories.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,168 @@
...
+    self._ds_info.update_splits_if_different(split_dict)
+
+    # Reset `_data_dir` as it should not change to DATA_DIR/Version
+    self._data_dir = self._root_dir
+
+  def _info(self):
+    return self._ds_info
+
+  def _download_and_prepare(self, **kwargs):
+    pass
```
Conchylicultor, line 104: We should raise an error here.
Also could you add unittest to make sure the error is correctly raised ?

If this was changed for the test, then the test should be modified to make this a no-op

tensorflow_datasets/core/custom_dataset/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Custom Datasets APIs"""
+
+from tensorflow_datasets.core.custom_dataset.image_folder import ImageLabelFolder
```
Conchylicultor, line 18: I would wrap this inside `with core.registered.skip_registration()` so that `image_folder` cannot be created from `tffds.load` (has to be explicitly defined)

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -20,30 +20,36 @@
...
       "test": 6,
   }
 
-  def test_info(self):
-    pass
-
   def test_registered(self):
-    self.assertIn("image_label_folder", registered.list_builders(),
-                  "Dataset was not registered.")
+    self.assertNotIn("image_label_folder", tfds.list_builders(),
```
Conchylicultor, line 50: I would move this test in a separate `tfds.testing.TestCase` class to separate `DatasetBuilderTestCase ` from `TestCase`

vijayphoenix: I was overriding a test in `DatasetBuilderTestCase`.
So, should I disable this test and add it in a separate class?

tensorflow_datasets/core/custom_dataset/image_folder.py
vijayphoenix: I had to fix MockFs listdir function a bit.
Making a new PR for the same.

tensorflow_datasets/testing/test_utils_test.py
```diff
@@ -109,8 +109,9 @@ def test_mock_fs(self):
       self.assertEqual(
           tf.io.gfile.listdir('/path/to'), tf.io.gfile.listdir('/path/to/'))
       self.assertEqual(
-          tf.io.gfile.listdir('/path/to'), ['file2', 'file1_moved'])
-
+          tf.io.gfile.listdir('/path/to'), ['file1_moved', 'file2'])
+      self.assertEqual(
+          tf.io.gfile.listdir('/path'), ['to'])
```
vijayphoenix, line 114: Same

tensorflow_datasets/testing/test_utils.py
```diff
@@ -124,11 +124,12 @@ def add_file(self, path, content=None) -> None:
...
-    return [
-        os.path.basename(p)
+    path = path.rstrip(os.path.sep) + os.path.sep
+    return sorted(list({
+        os.path.relpath(p, path).split(os.path.sep)[0]
         for p in self.files
-        if p.startswith(path.rstrip('/') + '/')  # Make sure path is a `dir/`
-    ]
+        if p.startswith(path.rstrip(os.path.sep) + os.path.sep)  # Make sure path is a `dir/`
+    }))
```
vijayphoenix, line 132: I will remove this file later
Just here for the tests to pass.

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -0,0 +1,115 @@
...
+                       'root_dir/train/label2/img2.png'],
+            'label3': ['root_dir/train/label3/img1.png',
+                       'root_dir/train/label3/img2.png',
+                       'root_dir/train/label3/img3.png'],
+        },
+        'val': {
+            'label1': ['root_dir/val/label1/img1.png'],
+            'label2': ['root_dir/val/label2/img2.png'],
+        },
+        'test': {
```
Conchylicultor, line 98: What happen if test split contains a `label4` ?

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+    fs = tfds.testing.MockFs()
+    with fs.mock():
+      for file in images:
+        fs.add_file(file)
+
+      out1, out2 = func('root_dir')
+      self.assertEqual(split_label_img, out1)
+      self.assertEqual(labels, out2)
+
```
Conchylicultor, line 113: Could you also tests that 

```
img_folder = ImageFolder(data_dir)
img_folder.split[].num_examples
img_folder.features['label'].names
```
Are correct ?

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -0,0 +1,115 @@
...
+  SPLITS = {
+      'train': 2,
+      'test': 6,
+  }
+  EXAMPLE_DIR = _EXAMPLE_DIR
+  
+  @classmethod
+  def setUpClass(cls): # pylint:disable = invalid-name
+    super(ImageFolderTest, cls).setUpClass()
+    cls.DATASET_CLASS.__init__ = new_init
```
Conchylicultor, line 52: It would be cleaner to use mocking here.

Also add `tearDownClass` to stop the mock.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,166 @@
...
+    self.info.update_splits_if_different(split_dict)
+
+  def _info(self):
+    return core.DatasetInfo(
+        builder=self,
+        description='Generic image classification dataset.',
+        features=core.features.FeaturesDict({
+            'image': core.features.Image(),
+            'label': core.features.ClassLabel(),
+        }),
```
Conchylicultor, line 97: I would also add the `filename` in the feature dict

tensorflow_datasets/core/custom_dataset/image_folder_test.py
Conchylicultor: See `DatasetBuilderTestCase` implementation for examples on how to use mock

tensorflow_datasets/core/custom_dataset/image_folder.py
Conchylicultor: Thank you for fixing the bug.

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -0,0 +1,115 @@
...
+import os
+
+import tensorflow_datasets.public_api as tfds
+
+_EXAMPLE_DIR = os.path.join(
+      tfds.testing.test_utils.fake_examples_dir(), 'image_folder_data')
+
+original_init = tfds.ImageFolder.__init__
+
+def new_init(self, root_dir=None, **kwargs):
```
Conchylicultor, line 28: `functools.partial(ImageFolder.__init__, root_dir= _EXAMPLE_DIR)`

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,181 @@
...
+      raise ValueError('Split name {} not present in {}'
+                       .format(split, self._split_label_images.keys()))
+
+    imgs = sorted(
+        img for l in self._split_label_images[split].values() for img in l)
+
+    ds = tf.data.Dataset.list_files(imgs)
+    ds = ds.map(self._process_ds,
+                num_parallel_calls=tf.data.experimental.AUTOTUNE)
+    if shuffle_files:
```
Conchylicultor, line 147: The shuffling should happen before the images are loaded in memory, otherwise this will load the full dataset in memory.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,181 @@
...
+
+  def _map_fn(self, path):
+    image_path = tf.strings.split(path, os.path.sep)[-3:]
+    img = tf.io.read_file(path)
+    img = tf.image.decode_jpeg(img, channels=3)
+    img = tf.image.convert_image_dtype(img, tf.uint8)
+    label = image_path[1].numpy()
+    return [
+        img,
+        self.info.features['label'].str2int(label),
```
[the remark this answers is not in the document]
vijayphoenix: So, is there a way to know the integer corresponding to the label?
I used this approach mainly because I used  `.str2int`

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,181 @@
...
+        tf.strings.reduce_join(image_path)
+    ]
+
+  def _process_ds(self, path):
+    output = tf.py_function(self._map_fn,
```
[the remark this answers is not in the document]
vijayphoenix: okay, will try this

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,181 @@
...
+                            inp=[path],
+                            Tout=list(self.info.features.dtype.values()))
+    output = {
+        'image': output[0],
+        'label': output[1],
+        'image/filename': output[2],
+    }
+    for key, value in output.items():
+      value.set_shape(self.info.features.shape[key])
```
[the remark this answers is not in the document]
vijayphoenix: I had to use this because of the current approach.
`py_function` is not able to determine the output shape.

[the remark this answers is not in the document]
Conchylicultor: Yes, which is another reason why we shouldn't use py_function

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -0,0 +1,122 @@
...
+        'root_dir/test/label4/img1.png',
+    ]
+
+    splits_info = {
+        'train': 6,
+        'val': 2,
+        'test': 3,
+    }
+
+    labels = {
```
Conchylicultor, line 76: Order is important. Labels should be a list

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,181 @@
...
+  ds = builder.as_dataset(split='split_name', shuffle_files=True)
+  tfds.show_examples(ds, builder.info)
+  ```
+
+  """
+
+  VERSION = core.Version('2.0.0')
+
+  def __init__(self, root_dir, **kwargs):
+    del kwargs
```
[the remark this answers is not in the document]
vijayphoenix: We get extra args like `data_dir`, `config` in tests.
I added `kwargs` here because I could not find a way to handle these extra args by `functools`.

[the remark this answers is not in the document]
Conchylicultor: If this is only for tests, then the test init function should be patched, not the original file.
If functools do not works, then use a function.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,176 @@
...
+        supervised_keys=('image', 'label'),
+    )
+
+  def _download_and_prepare(self, **kwargs):
+    raise NotImplementedError('No need to call download_and_prepare function.')
+
+  def download_and_prepare(self, **kwargs):
+    raise NotImplementedError('No need to call download_and_prepare function.')
+
+  def _process_ds(self, path, label):
```
Conchylicultor, line 108: If the function do not use `self`, it's best to make it a simple function rather than a method. I help the user understand the function do not have side effects & cie.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,176 @@
...
+
+  def _as_dataset(
+      self,
+      split,
+      shuffle_files=False,
+      decoders=None,
+      read_config=None):
+    """Generate dataset for given split"""
+    del decoders
+    del read_config
```
Conchylicultor, line 127: We should also raise an `NotImplementedError('Xyz not supported with {}'.format(type(self).__name__))` error when decoder or read_configs are passed

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,176 @@
...
+
+  def _process_ds(self, path, label):
+    image_path = tf.strings.split(path, os.path.sep)[-3:]
+    img = tf.io.read_file(path)
+    img = tf.image.decode_jpeg(img, channels=3)
+    img = tf.image.convert_image_dtype(img, tf.uint8)
+    return {
+        'image': img,
+        'label': tf.cast(label, tf.int64),
+        'image/filename': tf.strings.reduce_join(image_path)
```
Conchylicultor, line 116: I think we could pass the full original path here, without the split/join.

tensorflow_datasets/core/custom_dataset/image_folder.py
```diff
@@ -0,0 +1,176 @@
...
+      read_config=None):
+    """Generate dataset for given split"""
+    del decoders
+    del read_config
+    if split not in self.info.splits.keys():
+      raise ValueError('Split name {} not present in {}'
+                       .format(split, self._split_label_images.keys()))
+
+    img_paths = sorted(
+        img for l in self._split_label_images[split].values() for img in l)
```
Conchylicultor, line 133: You could use `_split_label_images[split].items()` to extract both labels & images at once.

Then use `images, labels = zip(*inputs)` to unpack them

tensorflow_datasets/core/custom_dataset/image_folder_test.py
```diff
@@ -0,0 +1,129 @@
...
+      'test': 6,
+  }
+  EXAMPLE_DIR = _EXAMPLE_DIR
+  MOCK_OUT_FORBIDDEN_OS_FUNCTIONS = False
+
+  @classmethod
+  def setUpClass(cls): # pylint:disable = invalid-name
+    super(ImageFolderTest, cls).setUpClass()
+    cls.DATASET_CLASS.__init__ = new_init
+    cls.DATASET_CLASS.download_and_prepare = mock.Mock(return_value=None)
```
Conchylicultor, line 50: What about `tearDownClass` to restore the original functions ?
Or I think there might be some `self.addCleanup()`

Conchylicultor:
It seems the tests are failing

tensorflow_datasets/core/custom_dataset/image_folder.py
vijayphoenix: I tried this... but could not fix the tests.
Tests keep on failing on this line in `dataset_builder.py`

```
ds = ds.prefetch(tf.data.experimental.AUTOTUNE)
```
```
Error: NoneType object has no prefetch method
```

vijayphoenix:
Merged by c47aa6ecd0dd62e114102286fec478af1330107f

PR #2088, opened 2020-06-15
Script to Benchmark dataset

vijayphoenix:
## Description
A util script for benchmarking dataset

### DIFF
--- a/tensorflow_datasets/scripts/benchmark_dataset.py
+++ b/tensorflow_datasets/scripts/benchmark_dataset.py
@@ -8,14 +8,14 @@ Instructions:
 ```
 python -m tensorflow_datasets.scripts.benchmark_dataset \
   --dataset dataset_name \
-  --epochs 3
+  --num_iterations 3
 ```
 
 Use the `--help` flag for more options and details.
 
 """
 import time
-from typing import Dict, Tuple
+from typing import Dict, Tuple, Any
 
 from absl import app
 from absl import flags
@@ -25,7 +25,7 @@ from packaging import version
 import tensorflow.compat.v2 as tf
 import tensorflow_datasets as tfds
 
-NUM_EXECUTIONS = 1
+NUM_ITERATIONS = 1
 FLAGS = flags.FLAGS
 
 flags.DEFINE_string('dataset', None, 'Dataset name')
@@ -33,7 +33,7 @@ flags.DEFINE_string(
     'data_dir', None,
     'Path to dataset directory (auto-computed). Dataset should already present')
 flags.DEFINE_integer(
-    'epochs', NUM_EXECUTIONS,
+    'num_iterations', NUM_ITERATIONS,
     'No. of iterations over the dataset to average out performance')
 flags.DEFINE_integer(
     'num_examples', None,
@@ -70,12 +70,13 @@ def _make_ds(
 
   return ds, num_examples
 
+
 @tf.function
 def benchmark(
     ds: tf.data.Dataset,
-    epochs: int,
+    num_iterations: int,
     ds_size: int,
-) -> None:
+) -> Dict[str, Any]:
   """Benchmarks the dataset.
 
   Reports:
@@ -86,26 +87,46 @@ def benchmark(
   if version.parse(tf.__version__) >= version.parse('2.2.0'):
     ds = ds.apply(tf.data.experimental.assert_cardinality(ds_size))
 
+  # Compute time taken by first iteration separately.
   start_time = time.perf_counter()
-  epoch_start_time = [start_time] # Can be used to plot a graph
+  time_per_iteration_list = []
+  first_batch_time = 0
   with tfds.core.utils.async_tqdm(
-      total=epochs, desc='Iterations completed...', unit=' epochs') as pbar:
-    for _ in range(epochs):
-      for _ in ds:
+      total=num_iterations,
+      desc='Iterations completed...',
+      unit=' iterations') as pbar:
+    for _ in range(num_iterations):
+      t1 = time.perf_counter()
+      iter_ds = iter(ds)
+      next(iter_ds)  # First warmup batch
+      t2 = time.perf_counter()
+      first_batch_time += t2 - t1
+      for _ in iter_ds:  # Other batch
         pass
       pbar.update()
-      epoch_start_time.append(time.perf_counter())
+      time_per_iteration_list.append(time.perf_counter() - t1)
+      # time_per_iteration_list.append(time.perf_counter() - t2) # exclude first batch
 
-  print('\nSummary\n')
   total_time = time.perf_counter() - start_time
-  first_iter_time = epoch_start_time[1] - epoch_start_time[0]
-  remaining_time = total_time - first_iter_time
+  time_per_iter = total_time / num_iterations
+  time_per_example = total_time / ds_size
+  avg_first_batch_time = first_batch_time / num_iterations
+  print('\nSummary\n')
   print('Dataset size:', ds_size)
-  print('Time taken for first iteration:', first_iter_time)
-  print('Avg time per Epoch (excluding first iteration):',
-        remaining_time / (epochs-1))
-  print('Element per second:', remaining_time / (epochs-1) / ds_size)
+  print('Time taken for first batch of dataset:', avg_first_batch_time)
+  print('Avg time per iteration:', time_per_iter)
+  print('Element per second:', time_per_example)
   print('Total Execution Time:', total_time)
+
+  return {
+      'dataset_size': ds_size,
+      'num_iterations': num_iterations,
+      'total_time': total_time,
+      'avg_first_batch_time': first_batch_time,
+      'time_per_iteration': time_per_iter,
+      'time_per_example': time_per_example,
+      'iteration_duration': time_per_iteration_list
+  }
 
 
 def main(_):
@@ -119,7 +140,7 @@ def main(_):
 
   ds_name = FLAGS.dataset
   data_dir = FLAGS.data_dir
-  epochs = FLAGS.epochs
+  num_iterations = FLAGS.num_iterations
 
   if FLAGS.disable_tqdm:
     tfds.disable_progress_bar()
@@ -127,7 +148,7 @@ def main(_):
   ds, num_examples = _make_ds(ds_name, data_dir)
 
   print('Benchmarking\n')
-  benchmark(ds, epochs, num_examples)
+  print(benchmark(ds, num_iterations, num_examples))
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+  start_time = time.perf_counter()
+  ds = tfds.load(dataset, data_dir=data_dir) # TODO: Exclude dataset download time
+  for epoch_num in range(epochs):
+    for split in ds:
+      for _ in ds[split]:
+        pass
+    print('Epoch {} completed at Time: {}'
+          .format(epoch_num, time.perf_counter() - start_time))
+
+  print('Total Execution Time: ', time.perf_counter() - start_time)
```
Conchylicultor, line 37: In addition to the total time, what's more important than the is the number of images/sec as it allow to compare datasets together.

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+flags.DEFINE_string('dataset', None, 'Dataset name')
+flags.DEFINE_string(
+    'data_dir', None,
+    'Path to dataset directory (auto-computed). Dataset should already present')
+flags.DEFINE_integer('epochs', EPOCHS, 'No. of iterations over the dataset')
+
+
+def benchmark(dataset, data_dir, epochs):
+  start_time = time.perf_counter()
+  ds = tfds.load(dataset, data_dir=data_dir) # TODO: Exclude dataset download time
```
Conchylicultor, line 29: You can use `tfds.builder` with `builder.download_and_prepare()` + `builder.as_dataset()`

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+import time
+
+from absl import app
+from absl import flags
+
+import tensorflow_datasets as tfds
+
+# DEFAULT_DATA_DIR = os.path.expanduser(os.path.join("~", "tensorflow_datasets"))
+
+EPOCHS = 1
```
Conchylicultor, line 16: Rather than epoch, it's best to have `num_examples`

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string('dataset', None, 'Dataset name')
+flags.DEFINE_string(
+    'data_dir', None,
+    'Path to dataset directory (auto-computed). Dataset should already present')
+flags.DEFINE_integer('epochs', EPOCHS, 'No. of iterations over the dataset')
+
+
+def benchmark(dataset, data_dir, epochs):
```
Conchylicultor, line 27: We want to compare different pipeline together, with different `batch_size`, `split='train[10%:]'`, `ds = ds.map`,... so the current signature seems too restrictive.

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+  start_time = time.perf_counter()
+  ds = tfds.load(dataset, data_dir=data_dir) # TODO: Exclude dataset download time
+  for epoch_num in range(epochs):
+    for split in ds:
```
Conchylicultor, line 31: We'll want to compare TF1 / TF2 to see if there is performance drop with eager.

Conchylicultor, line 31: The first iteration may have additional overhead so we might want to count it separately (We should display first iteration time, img/sec for the next x iteration, total time).

tensorflow_datasets/scripts/benchmark_dataset.py
Conchylicultor: We want to allow user to import the script to benchmark their own pipeline.

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+) -> None:
+  """Benchmarks the dataset.
+
+  Reports:
+  - Total Execution time
+  - Time taken per Epoch
+  - Dataset elements per second
+  """
+
+  num_examples = tf.data.experimental.cardinality(dataset)
```
vijayphoenix, line 58: Other methods
- Pass `num_examples` as a function parameter. (For TFDS datasets, we will easily get it from `info.splits[split_name].num_examples`)
- Calculate `num_examples` in the epoch loop below.

vijayphoenix:
@Conchylicultor unable to reproduce the tf-nightly test failure. [Colab Link](https://colab.research.google.com/drive/1JWbcBh_1nM7Z81KzbZkWFFR44GKEbQ-Z?usp=sharing)

tensorflow_datasets/scripts/benchmark_dataset.py
vijayphoenix: So should we benchmark the dataset twice?

For example
```
benchmark(ds, epochs)
tf.compat.v1.disable_eager_execution()
benchmark(ds, epochs)
```

Conchylicultor: Eager/non eager mode should be an option of the benchmark util. If the user want to test the two option, they should call the util twice.

tensorflow_datasets/scripts/benchmark_dataset.py
Conchylicultor: We shouldn't auto-compute the number of images for the dataset size. Instead, we want to compute the benchmark of only 10.000 images for a variety of datasets (e.g. first 10.000 images on imagenet, cifar,...)
Num images should be an option of the benchark (which can default to cardinality if not set)

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+  if num_examples == tf.data.experimental.UNKNOWN_CARDINALITY:
+    num_examples = sum(1 for _ in dataset)
```
Conchylicultor, line 60: Let's not iterate twice over the full dataset as it may takes multiple hours so we don't want to double the time.

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+  elif num_examples == tf.data.experimental.INFINITE_CARDINALITY:
+    raise RuntimeError('Dataset is of INFINITE_CARDINALITY')
+  else:
+    num_examples = num_examples.numpy()
+
+  start_time = time.perf_counter()
+  epoch_start_time = [start_time]
+
+  for epoch_num in range(1, epochs+1):
+    for _ in dataset:
```
Conchylicultor, line 70: We should compute the first batch time separately.

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def main(_):
+  ds_name = FLAGS.dataset
+  data_dir = FLAGS.data_dir
+  epochs = FLAGS.epochs
+
+  ds = _preprocess(ds_name, data_dir)
+
+  for split in ds:
```
Conchylicultor, line 93: No need to benchmark all splits. Lets only benchmark train by default.

tensorflow_datasets/scripts/benchmark_dataset.py
Conchylicultor, line 70: We could try to add tqdm so the user can see the progression bar with the live statistics. I think tqdm should add a negligeable overhead (and we can add an option to disable it)

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string('dataset', None, 'Dataset name')
+flags.DEFINE_string(
+    'data_dir', None,
+    'Path to dataset directory (auto-computed). Dataset should already present')
+flags.DEFINE_integer('epochs', EPOCHS, 'No. of iterations over the dataset')
+
+
+def _preprocess(
```
Conchylicultor, line 29: I would rename this `_make_dataset`

tensorflow_datasets/scripts/benchmark_dataset.py
Conchylicultor: Or if for you this correspond to the number of trial to average the performance across multiple trials, maybe rename this `NUM_EXECUTIONS` or `NUM_REPETITIONS` or `NUM_TRIALS`.

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+  """
+  if version.parse(tf.__version__) >= version.parse('2.2.0'):
+    ds = ds.apply(tf.data.experimental.assert_cardinality(ds_size))
+
+  start_time = time.perf_counter()
+  epoch_start_time = [start_time] # Can be used to plot a graph
+  with tfds.core.utils.async_tqdm(
+      total=epochs, desc='Iterations completed...', unit=' epochs') as pbar:
+    for _ in range(epochs):
+      for _ in ds:
```
Conchylicultor, line 94: As explained in https://github.com/tensorflow/datasets/pull/2088#discussion_r440376627 and https://github.com/tensorflow/datasets/pull/2088#discussion_r441950643
we should compute the first iteration separately. 

By first iteration/first batch, I meant first iteration of the dataset, not the first trial.

tensorflow_datasets/scripts/benchmark_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+  ds = builder.as_dataset(split=f'train[:{percentage}%]')
+  print('\nTime taken to preprocess {} dataset: {}\n'
+        .format(name, time.perf_counter() - start_time))
+
+  return ds, num_examples
+
+@tf.function
+def benchmark(
+    ds: tf.data.Dataset,
+    epochs: int,
```
Conchylicultor, line 76: When renaming `EPOCH` -> `NUM_EXECUTION`, all instance of `epoch` should be renamed. Even if I only post a single review comment. I assume that the fix is applied globally.

tensorflow_datasets/scripts/benchmark_dataset.py
vijayphoenix: Sorry, I misunderstood it.
Fixed it now.

Conchylicultor: Sorry for the confusion. By first iteration, I meant:

```python
iter_ds = iter(ds)
next(iter_ds)  # First warmup batch/iteration
for n in iter_ds:  # Other iterations
  print(n)
```

vijayphoenix: ohk..got confused by iteration / batch.
Fixed

vijayphoenix:
Merged by 43bbad11

Issue #2092, opened 2020-06-16
Data set problem


ahmadnazeeh:
**What I need help with / What I was wondering**
I have my own data set in ".csv" format with 2-columns, I need to use my own data set instead of "ted_hrlr_translate/pt_to_en", that is loaded as an example for data set from TED Talks. The code is:

examples, metadata = tfds.load('ted_hrlr_translate/pt_to_en', with_info=True,
                               as_supervised=True)
train_examples, val_examples = examples['train'], examples['validation']

my own data set is " ATB3-all.csv "

Conchylicultor:
`ted_hrlr_translate` will load TED data. If you want your own data, you can create your own dataset: https://www.tensorflow.org/datasets/add_dataset.

We could make it easier to create custom datasets. We've started with `tfds.ImageFolder` in tfds-nightly.

vijayphoenix:
Similar issue https://github.com/tensorflow/datasets/issues/1402
We can simplify this use case by providing some API like `tfds.TranslationDataset`.

Basically, create a dataset from a folder with the given structure.
Each example of a language `lang1` will be a line of text from the file `lang1.split_name`
  ```
  path/to/manual_dir/<dataset_name>/
      lang1.train
      lang2.test
      lang1.test
      lang2.test
      ...
  ```
Some References APIs:
- [tf.data.TextLineDataset](https://www.tensorflow.org/api_docs/python/tf/data/TextLineDataset)
- https://pytorch.org/text/datasets.html#machine-translation

omerfbaspinar:
is there any step by step example for this ? 
i read this and can not apply for my csv dataset. 
https://www.tensorflow.org/datasets/add_dataset#tldr

Issue #1402, opened 2020-01-19
Creating custom dataset in tfds format


ghost:
**What I need help with / What I was wondering**
I am trying to create a sequence to sequence dataset in the exact format as the `ted_hrlr_translate/pt_to_en` data used in the [transformer example](https://www.tensorflow.org/tutorials/text/transformer). How can I do this?

**What I've tried so far**
I tried to create my own dataset but I am unable to feed my data into the transformer model as the dataset format does not match. I basically want to use the existing transformer model code on my custom data which is in text format. Each line is a sample. Each sample has the input sequence and output sequence separated by a tab.

**It would be nice if...**
there was a standalone script to convert text data to tfds format.

Conchylicultor:
To create your dataset, have a look at our guide: https://www.tensorflow.org/datasets/add_dataset

For text data, you just have to load your file and yield the examples that you want:
```
class MyDataset(tfds.core.GeneratorBasedBuilder):

  def _generate_examples(self, path):
    with tf.io.gfile.GFile(path) as f:  # path to custom data
      for i, line in enumerate(f):
        source, target = line.split('\t')
        yield i, {
            'source': source,
            'target': target,
        }

```

Conchylicultor:
Closing this as I think the question was answered. We might add a new util class in the future to simplify this use-case.

8904591612:
> Closing this as I think the question was answered. We might add a new util class in the future to simplify this use-case.

can you please help in solving ,  as per the documentation in tensorflow I followed  https://www.tensorflow.org/datasets/api_docs/python/tfds/folder_dataset/TranslateFolder    , followed the instructions but does not allow  to set  "supervised=True",      can we use  [tensorflow_datasets/core/custom_dataset/translate_dataset.py]
I have the same problem tto create custom data set from parallel corpus for training language model. Please help

PR #2094, opened 2020-06-17
Add Custom Translate Datasets template

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/2092 and also https://github.com/tensorflow/datasets/issues/1402


tensorflow_datasets/core/custom_dataset/translate_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _get_language_examples(
+    root_dir: str
+) -> Tuple[Dict[str, Dict[str, List[str]]], List[str]]:
+  split_lang_examples = dict()
+  languages = set()
+  for split_name in _list_folders(root_dir):
+    split_dir = os.path.join(root_dir, split_name)
+    split_lang_examples[split_name] = {
+        f[:-4]: _list_examples(os.path.join(split_dir, f))
```
Conchylicultor, line 112: Avoid magic numbers. It's best to be explicit (`[:len('.ext')]`)

tensorflow_datasets/core/custom_dataset/translate_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+    lang_example_dict = self._languages_examples_dict[split]
+    ds = tf.data.Dataset.from_tensor_slices(lang_example_dict)
+    if shuffle_files:
+      ds = ds.shuffle(len(lang_example_dict))
+    return ds
+
+
+def _get_language_examples(
+    root_dir: str
+) -> Tuple[Dict[str, Dict[str, List[str]]], List[str]]:
```
Conchylicultor, line 106: Docstring with `Returns:`

tensorflow_datasets/core/custom_dataset/translate_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+    self._languages_examples_dict, languages = _get_language_examples(root_dir)
+    self._languages = sorted(list(languages))
+    super(TranslateDataset, self).__init__()
+    
+    # Reset `_data_dir` as it should not change to DATA_DIR/Version
+    self._data_dir = root_dir
+
+    # Update DatasetInfo splits
+    split_dict = split_lib.SplitDict(self.name)
+    for split_name, examples in self._languages_examples_dict.items():
```
Conchylicultor, line 56: As per image_folder.py, I think `_split_examples` would be a better name than `_languages_examples` as it would indicates that dict is mapping `split_name`->`examples`

tensorflow_datasets/core/custom_dataset/translate_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+    size = len(next(iter(split_lang_examples[split_name].values())))
+    # One-to-One translation
+    for lang, example in split_lang_examples[split_name].items():
+      languages.add(lang)
+      assert len(example) == size
+
+  return split_lang_examples, languages
+
+
+def _list_folders(root_dir: str) -> List[str]:
```
Conchylicultor, line 125: Why not reuse the utils from `image_folder` ? Those utils could be moved in py_utils or in some `io_utils`

tensorflow_datasets/core/custom_dataset/translate_dataset.py
```diff
@@ -0,0 +1,135 @@
...
+  ```
+
+  """
+
+  VERSION = version.Version('1.0.0')
+
+  def __init__(self, root_dir: str):
+    root_dir = os.path.expanduser(root_dir)
+    self._languages_examples_dict, languages = _get_language_examples(root_dir)
+    self._languages = sorted(list(languages))
```
Conchylicultor, line 48: You only need the languages in `__init__`. No need to load all files (e.g. loading the train set if only the eval set is used).

You could only extract dict `split -> filepaths` and load the required files in `_as_datasets`

vijayphoenix: We will have to read all the file content to update the split info, right?

Conchylicultor: I see. I guess it's fine to load the full data then.

PR #2098, opened 2020-06-18
Export Translate API

vijayphoenix:
Currently, documentation is not generated for the following 
- `tfds.features.Translate` 
- `tfds.feature.TranslationVariableLanguages`


Conchylicultor:
I think this was a conscious decision as we should not encourage to use those API.

vijayphoenix:
Both the APIs?
Is there be a alternative for Translation API?

vijayphoenix:
> I think this was a conscious decision as we should not encourage to use those API.

Okay.
Closing

Issue #2105, opened 2020-06-21
GCS utils returns only 1000 files


acharles7:
**Is your feature request related to a problem? Please describe.**

Currently, [tensorflow_datasets/core/utils/gcs_utils.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/utils/gcs_utils.py#L77) uses `requests` to return files from GCP Bucket 

```python
def gcs_files(prefix_filter=None):
  """List all files in GCS bucket."""
  top_level_xml_str = download_gcs_file('', prefix_filter=prefix_filter)
  xml_root = ElementTree.fromstring(top_level_xml_str)
  filenames = [el[0].text for el in xml_root if el.tag.endswith('Contents')]
  return filenames
```
The above function `gcs_files` returns only 1000 objects per request as describe in google [XML_API](https://cloud.google.com/storage/docs/xml-api/get-bucket-list). But there are more than 1000 files in the bucket that we need to complete the whole dataset.

**Describe the solution you'd like**
We can iterate (requests) until we get all files using the value of `NextMarker` provided in the `XML Response` for continue listing the Bucket after the set of results

**Describe alternatives you've considered**
We can also use `tf.io.gfile`

**Additional context**
#1938

acharles7:
@Conchylicultor Please close this issue.

Issue #1938, opened 2020-04-23
tfds.load downloads nsynth with missing parts


casab:
**Short description**
I am using tfds.load to download nsynth dataset. But no matter how many times I try, It's only downloaded up until nsynth-train.tfrecord-00975-of-01024. It seems like 976 to 1023 are missing.

**Environment information**
* Operating System: Windows 10
* Python version: 3.7.3
* `tensorflow-datasets`/`tfds-nightly` version: tensorflow-datasets@2.1.0
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: tensorflow@2.1.0

**Reproduction instructions**

```
`datasets = tfds.load("nsynth", data_dir="data")`
```

**Link to logs**
```
Dl Completed...: 100%|██████████████████████████████████████████████████████████| 989/989 [2:28:10<00:00,  1.89s/ file]

Dl Completed...: 100%|██████████████████████████████████████████████████████████| 989/989 [2:28:10<00:00,  8.99s/ file]

Dataset nsynth downloaded and prepared to data\nsynth\full\2.3.0. Subsequent calls will reuse this data.
```

**Expected behavior**
I expect the whole of the dataset to be downloaded.

Conchylicultor:
This seems to be a bug.
This dataset is pregenerated at: https://pantheon.corp.google.com/storage/browser/tfds-data/datasets/nsynth/full/

It may indicates a bug in our `download_gcs_dataset` which do not fetch the right number of files. https://github.com/tensorflow/datasets/blob/300eb8c751abc25b2780407714a231f0a18567c9/tensorflow_datasets/core/utils/gcs_utils.py#L81

We should check that `gcs_files` returns the right files and investigate why otherwise (Does GCS has a limit toward the number of file which gets returned).
Don't hesitate if someone want to investigate.

Conchylicultor:
In the meantime, you could try to copy the datasets manually, using: 

`gsutil cp tfds-data/datasets/nsynth/full/2.3.0 ~tensorflow_datasets/nsynth/full/2.3.0`

Or you could stream the dataset without downloading with:

`ds = tfds.load("nsynth", data_dir="data", try_gcs=True)`

Eshan-Agarwal:
[...]

PR #2106, opened 2020-06-21
Updated GCS utils 

acharles7:
## Updated `gcs_utils.py`

* Issue Reference: #2105 

## Description
Earlier [`gcs_utils.py`](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/utils/gcs_utils.py#L77) returns only 1000 files, With this PR `gcs_utils.py` return all the files from the bucket (>1000)
This issue arises in  #1938

### DIFF
--- a/tensorflow_datasets/core/utils/gcs_utils.py
+++ b/tensorflow_datasets/core/utils/gcs_utils.py
@@ -22,6 +22,7 @@ from __future__ import print_function
 
 import concurrent.futures
 import posixpath
+import re
 from typing import Optional
 from xml.etree import ElementTree
 
@@ -56,11 +57,13 @@ def gcs_path(suffix: Optional[str] = None) -> str:
   return path
 
 
-def download_gcs_file(path, out_fname=None, prefix_filter=None):
+def download_gcs_file(path, out_fname=None, prefix_filter=None, marker=None):
   """Download a file from GCS, optionally to a file."""
   url = posixpath.join(GCS_BUCKET, path)
   if prefix_filter:
     url += '?prefix=%s' % prefix_filter
+  if marker:
+    url += '&marker={}'.format(marker)
   stream = bool(out_fname)
   resp = requests.get(url, stream=stream)
   if not resp.ok:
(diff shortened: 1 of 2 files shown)


acharles7:
Sure, I'll try it and get back to you.
But in the meantime, we have to fix this while I am working on `tf.io.gfile`

tensorflow_datasets/core/utils/gcs_utils.py
```diff
@@ -58,28 +55,23 @@ def gcs_path(suffix: Optional[str] = None) -> str:
...
-  if not resp.ok:
-    raise ValueError('GCS bucket inaccessible')
+    url = posixpath.join(url, prefix_filter)
   if out_fname:
-    with tf.io.gfile.GFile(out_fname, 'wb') as f:
-      for chunk in resp.iter_content(1024):
-        f.write(chunk)
+    dirpath, file = posixpath.split(out_fname)
+    if not tf.io.gfile.exists(dirpath):
+      tf.io.gfile.makedirs(dirpath)
```
Conchylicultor, line 64: Why was `makedirs` required ? It wasn't required before ?

acharles7: Yes, It was not there before. I added to solve the error on my local machine and thought it will also be required here. I'm not sure it is necessary or not.

Conchylicultor: In which case did you encounter the error. Do you remember ?

acharles7: When any file came for downloading(`out_fname=somepath`), If our local directory is not there then I got 
`NotFoundError: mnist/1.2.0/dataset_info.json; No such file or directory` where `mnist/1.2.0/` is my `local directory` path
I think we don't need this
```python
    dirpath, file = posixpath.split(out_fname)
    if not tf.io.gfile.exists(dirpath):
      tf.io.gfile.makedirs(dirpath)
```
Because, As per my understanding the directories are prebuilt before coming into `download_gcs_file` function
So we don't need to make directories again. Correct me if I'm wrong.
Also, I don't know where to test `gcs_utils.py`, So I test each function by passing some random datasets and local dirs from GCP

Conchylicultor: Thanks I've imported this internally and will merge this soon.

PR #2120, opened 2020-06-24
Remove redundant shape and string checks to improve performance

TonyWZ:
1. Added unit tests to check that in `_item_to_tf_feature` from `tensorflow_datasets.core.example_serializer`, a `ValueError` is raised when
  - the shape of `item` is not compatible with `tensor_info.shape`.
  - `tensor_info.dtype == tf.string` but `item` is not string.
2. Removed redundant checks in `_item_to_tf_feature`. These checks are already performed in `_item_to_np_array`, which is called in the first line of `_item_to_tf_feature`. Performing this check again in `_item_to_tf_feature` would be redundant and slows down data preparation time.

  - The added unit tests confirm that a `ValueError` would still be raised in corresponding scenarios.

  - Performance benchmark: tested with the following command with the data already downloaded.
`time python3 -m tensorflow_datasets.scripts.download_and_prepare --datasets=amazon_us_reviews/Video_v1_00`
Before removing redundant checks: 9m19.551s
After removing redundant checks: 8m9.382s 

3. Fixed typos in `tensorflow_datasets.core.example_serializer`.

@vancezuo Could you please review this PR?

### DIFF
--- a/tensorflow_datasets/core/example_serializer_test.py
+++ b/tensorflow_datasets/core/example_serializer_test.py
@@ -20,6 +20,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import textwrap
 import numpy as np
 import tensorflow.compat.v2 as tf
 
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/example_serializer_test.py
```diff
@@ -201,6 +201,23 @@ def test_add_ragged_fields_single_level_sequence(self):
...
+    with self.assertRaises(ValueError):
+      example_serializer._item_to_tf_feature(example_item, tensor_info)
+
+  def test_item_to_tf_feature_string_check(self):
+    # Test string check in _item_to_tf_feature raises ValueError.
+    example_item = [1, 2, 3, 4, 5]
+    tensor_info = feature_lib.TensorInfo(shape=(5,), dtype=tf.string)
+    with self.assertRaisesRegex(
+        ValueError,
+        'Unsupported value: (.*)\nCould not convert to bytes list.',
```
vancezuo, line 217: Small note: This searches for a partial match. If trying to find a full match, include the start/end special chars (`^``$`). In this particular case it doesn't matter too much, but I just wanted to make sure you're aware.

TonyWZ: Got it. I will keep that in mind.

PR #2122, opened 2020-06-25
Optimizing run_with_reraise

TonyWZ:
1. Added unit tests to check that in `_dict_to_tf_example` from `tensorflow_datasets.core.example_serializer`
  - when an error is raised during the processing of a feature, the exception is reraised with the corresponding prefix.
  - when an error is raised but not during the processing of a feature (for instance during `zip_dict`), the exception is raised directly without reraise.
2. Changed `run_with_reraise` in `_dict_to_tf_example` from using a context manager to using a try except block. Since Python context manager is expensive in terms of performance, such change reduced the time required for preparing data.

  - The added unit tests confirm that an error raised during the processing of a feature would be reraised with the correct prefix message.

  - Performance benchmark: tested with the following command with the data already downloaded.
`time python3 -m tensorflow_datasets.scripts.download_and_prepare --datasets=amazon_us_reviews/Video_v1_00`
Before removing the context manager: 7m21.970s
After removing the context manager: 6m28.479s

@vancezuo Could you please review this PR?

### DIFF
--- a/tensorflow_datasets/core/example_serializer.py
+++ b/tensorflow_datasets/core/example_serializer.py
@@ -69,10 +69,12 @@ def _dict_to_tf_example(example_dict, tensor_info_dict):
     example_proto: `tf.train.Example`, the encoded example proto.
   """
   def run_with_reraise(fn, k, example_data, tensor_info):
-    try:
-      return fn(example_data, tensor_info)
-    except Exception:
-      utils.reraise("Error while serializing feature `{}`: `{}`: ".format(k, tensor_info))
+     try:
+       return fn(example_data, tensor_info)
+     except Exception:
+       utils.reraise(
+           'Error while serializing feature `{}`: `{}`: '.format(k, tensor_info)
+       )
 
   if tensor_info_dict:
     # Add the RaggedTensor fields for the nested sequences
--- a/tensorflow_datasets/core/example_serializer.py
+++ b/tensorflow_datasets/core/example_serializer.py
@@ -69,9 +69,10 @@ def _dict_to_tf_example(example_dict, tensor_info_dict):
     example_proto: `tf.train.Example`, the encoded example proto.
   """
   def run_with_reraise(fn, k, example_data, tensor_info):
-    with utils.try_reraise(
-        "Error while serializing feature `{}`: `{}`: ".format(k, tensor_info)):
+    try:
       return fn(example_data, tensor_info)
+    except Exception:
+      utils.reraise("Error while serializing feature `{}`: `{}`: ".format(k, tensor_info))
 
   if tensor_info_dict:
     # Add the RaggedTensor fields for the nested sequences
--- a/tensorflow_datasets/core/example_serializer.py
+++ b/tensorflow_datasets/core/example_serializer.py
@@ -71,7 +71,7 @@ def _dict_to_tf_example(example_dict, tensor_info_dict):
   def run_with_reraise(fn, k, example_data, tensor_info):
      try:
        return fn(example_data, tensor_info)
-     except Exception:
+     except Exception:  # pylint: disable=broad-except
        utils.reraise(
            'Error while serializing feature `{}`: `{}`: '.format(k, tensor_info)
        )
--- a/tensorflow_datasets/core/example_serializer.py
+++ b/tensorflow_datasets/core/example_serializer.py
@@ -69,12 +69,12 @@ def _dict_to_tf_example(example_dict, tensor_info_dict):
     example_proto: `tf.train.Example`, the encoded example proto.
   """
   def run_with_reraise(fn, k, example_data, tensor_info):
-     try:
-       return fn(example_data, tensor_info)
-     except Exception:  # pylint: disable=broad-except
-       utils.reraise(
-           'Error while serializing feature `{}`: `{}`: '.format(k, tensor_info)
-       )
+    try:
+      return fn(example_data, tensor_info)
+    except Exception:  # pylint: disable=broad-except
+      utils.reraise(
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/example_serializer_test.py
```diff
@@ -218,6 +218,39 @@ def test_item_to_tf_feature_string_check(self):
...
+        ValueError,
+        (
+            '^Error while serializing feature `input`:'
+            ' `TensorInfo\(shape=\(2,\), dtype=tf.int64\)`:(.*)'
+        ),
+    ):
+      example_serializer._dict_to_tf_example(example_data, tensor_info)
+
+  def test_dict_to_tf_example_flatten_nest_dict(self):
+    # Test that a KeyError from zip_dict in _dict_to_tf_example is not reraised
```
vancezuo, line 240: Add period.

tensorflow_datasets/core/example_serializer_test.py
```diff
@@ -218,6 +218,39 @@ def test_item_to_tf_feature_string_check(self):
...
+        'input': feature_lib.TensorInfo(
+            shape=(2,),
+            dtype=tf.int64,
+        ),
+    }
+    with self.assertRaisesRegex(
+        ValueError,
+        (
+            '^Error while serializing feature `input`:'
+            ' `TensorInfo\(shape=\(2,\), dtype=tf.int64\)`:(.*)'
```
vancezuo, line 234: I like that this checks the correct error is raised, but I worry that it's a bit fragile because it depends on the output format of `TensorInfo`. Would just "Error while serializing feature" be sufficient to confirm the error?

vijayphoenix:
@TonyWZ Please report the difference again after using this function
https://www.tensorflow.org/datasets/api_docs/python/tfds/core/benchmark

tensorflow_datasets/core/example_serializer_test.py
TonyWZ: Fixed in 012a267cc5cadcc9ee8678dd9709cc09b818884f.

tensorflow_datasets/core/example_serializer_test.py
TonyWZ: I think it would be. An alternative, if we want to check that the correct `tensor_info` is included in the error message, is to get the string representation of `tensor_info`, but since parentheses are special characters in regex, we would have to replace the parentheses with `\(` and `\)`. Do you think it would be worth it?

TonyWZ:
> @TonyWZ Please report the difference again after using this function
> https://www.tensorflow.org/datasets/api_docs/python/tfds/core/benchmark

@vijayphoenix Thanks for letting me know about this function. Could you clarify a bit about the meaning of each entry in the output? Specifically what does "first", "last", and "raw" mean? I could not find an explanation in the documentation.

tensorflow_datasets/core/example_serializer_test.py
vancezuo: In that case you can call [`re.escape`](https://docs.python.org/3/library/re.html#re.escape) on the `tensor_info` string output to do those substitutions for you.

tensorflow_datasets/core/example_serializer.py
```diff
@@ -69,9 +69,12 @@ def _dict_to_tf_example(example_dict, tensor_info_dict):
     example_proto: `tf.train.Example`, the encoded example proto.
   """
   def run_with_reraise(fn, k, example_data, tensor_info):
-    with utils.try_reraise(
```
Conchylicultor, line 72: Thank you for fixing this. 
Is the contextmanager really the bottleneck ? I would assume it to add a negligeable overhead (~5 s / millions examples)

The f-string however might slow down things. How about updating `try_reraise` instead ?

```
utils.try_reraise('{} {}', k, tensor_info)
```

And do the `.format` inside the `except` block ?

What do you think ?

tensorflow_datasets/core/example_serializer_test.py
TonyWZ: I see. Added the `tensor_info` string in 36dfde0e178c788504a33ab7913cc0711549738c.

tensorflow_datasets/core/example_serializer_test.py
```diff
@@ -218,6 +218,36 @@ def test_item_to_tf_feature_string_check(self):
...
+    example_data = {'x': [1, 2, 3]}
+    tensor_info = {
+        'y': feature_lib.TensorInfo(
+            shape=(2,),
+            dtype=tf.int64,
+        ),
+    }
+    with self.assertRaisesRegex(
+        KeyError,
+        '^(?!Error while serializing feature)(.*)',
```
Conchylicultor, line 251: Note: No need to have a full regex which are difficult to parse.

`self.assertRaisesRegex(KeyError, 'Error while serializing feature')` works well enough and the likelihood that another exception match is very low.

tensorflow_datasets/core/example_serializer.py
TonyWZ: I agree that moving `.format` inside the except block would improve performance, but I think the context manager is also slowing down the performance significantly. I did a timed run after moving `.format` into the except block and kept the context manager. The resulting runtime is 7m25.011s, which is slower than a timed run with the version without the context manager, which took 6m30.327s.

tensorflow_datasets/core/example_serializer_test.py
TonyWZ: Got it. Fixed in 72c8d5adffe5b58d291da3bca50c4251e1b8f2bf.

PR #2123, opened 2020-06-26
Remove py_utils.abstractclassmethod

lgeiger:
Since Python 3.3 it has been possible to use `classmethod` with `abstractmethod()`, making this decorator redundant.


Conchylicultor:
Thank you! Please note the text encoder API is deprecated and will be removed in a future version.

PR #2125, opened 2020-06-26
Simplify access counter in FeatureDict._flatten

lgeiger:
This PR adds a small helper class to track successful accesses of `dict.get` in `FeatureDict._flatten`


lgeiger:
> Is it a performance optimisation or is it just aesthetic ?

I didn't measure so this is mostly aesthetics.

PR #2128, opened 2020-06-26
Set __slots__ for frequently used classes

lgeiger:
Frequently used objects can benefit from defining [`__slots__`](https://docs.python.org/3/reference/datamodel.html#slots).

From the Python docs:
> `__slots__` allow us to explicitly declare data members (like properties) and deny the creation of `__dict__` and `__weakref__` (unless explicitly declared in `__slots__` or available in a parent.)
The space saved over using `__dict__` can be significant. Attribute lookup speed can be significantly improved as well.

This PR explicitely defines `__slots__` for some frequently used classes that have very few attributes, tend not to be inherited from and don't use dynamic assignment of new variables.

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -72,6 +72,8 @@ class BuilderConfig(object):
   `BuilderConfig` and add their own properties.
   """
 
+  __slots__ = ["_name", "_version", "_supported_versions", "_description"]
+
   @api_utils.disallow_positional_args
   def __init__(self, name, version=None, supported_versions=None,
                description=None):
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -72,8 +72,6 @@ class BuilderConfig(object):
   `BuilderConfig` and add their own properties.
   """
 
-  __slots__ = ["_name", "_version", "_supported_versions", "_description"]
-
   @api_utils.disallow_positional_args
   def __init__(self, name, version=None, supported_versions=None,
                description=None):
--- a/tensorflow_datasets/core/utils/tf_utils.py
+++ b/tensorflow_datasets/core/utils/tf_utils.py
@@ -62,6 +62,8 @@ class TFGraphRunner(object):
 
   """
 
+  __slots__ = ["_graph_run_cache"]
+
   def __init__(self):
     """Constructor."""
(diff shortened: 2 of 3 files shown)


tensorflow_datasets/core/dataset_builder.py
```diff
@@ -72,6 +72,8 @@ class BuilderConfig(object):
   `BuilderConfig` and add their own properties.
   """
 
+  __slots__ = ["_name", "_version", "_supported_versions", "_description"]
```
Conchylicultor, line 75: For `BuilderConfig`, this seems over-optimisation. How much does it actually save ? Unless you have a dataset with millions of configs, I don't really understand where it would change something.

Additionally, this class is frequently subclassed.

TensorInfo & GraphRunner are good though.

tensorflow_datasets/core/utils/tf_utils.py
```diff
@@ -62,6 +62,8 @@ class TFGraphRunner(object):
 
   """
 
+  __slots__ = ["_graph_run_cache"]
```
Conchylicultor, line 65: Not in this PR, but a better optimization is probably to remove the `TFGraphRunner` entirely. I believe it is only used for encoding images, and currently it recreate a new graph each time an input shape is different.
I believe images could be encoded using a unique graph with `placeholder(shape=(None, None, channel))` (or tf.function(inputs=[TensorSpec(shape=(None,None,channel)]) in eager)).

Currently eager functions are not even wrapped in tf.function

tensorflow_datasets/core/dataset_builder.py
lgeiger: Good point, I removed it in e3f7366

tensorflow_datasets/core/utils/tf_utils.py
lgeiger: > Currently eager functions are not even wrapped in tf.function

Good point. Is there a timeline for dropping TF 1.x support in TFDS which would make such optimizations a lot easier?

Conchylicultor: > Good point. Is there a timeline for dropping TF 1.x support in TFDS which would make such optimizations a lot easier?

Even after dropping TF 1.5 support, we're not planning to drop support for graph mode any time soon as it is still widely used (People can still use `tf.compat.v1.Session` in TF 2.x).
The easiest here would be to only implement graph mode wrapped in `tf.Graph` as it should work both for eager/graph.

Issue #2139, opened 2020-06-30
One folder per dataset


vijayphoenix:
Currently, we're grouping files per file type: `(checksums/, fake_data/, datasets/).`
We should try to move toward a more modular approach where datasets are self-contained directories:
```
dataset_type0/   # image, text, etc
    dataset0/
        dataset0.py
        dataset0_test.py
        checksums.txt
        fake_data/
    dataset1/
        dataset1.py
        ...
```

**TODOs:**

1) We should make it easy for external users to follow this model:
Checksums path, fake data,... should be automatically detected without the user having to do the additional setup (tfds.core.add_checksums_dir, FAKE_DATA_DIR,...)

2) Update scripts like `create_new_dataset`, `document_dataset`, etc.

3) Update dataset testing methods

4) Update all files depending on checksums, fake_dataset locations etc

PR #2141, opened 2020-07-01
Support datasets in folders with checksums & fake_data

vijayphoenix:
Related to https://github.com/tensorflow/datasets/issues/2139

Required directory structure for datasets outside TFDS:
```
path/to/dataset/
    my_dataset.py
    my_dataset_test.py
    my_dataset.txt
    fake_examples/
 ```

TODO:
- [ ] Update [Define the dataset outside TFDS](https://www.tensorflow.org/datasets/add_dataset#define_the_dataset_outside_tfds) doc.
- [ ] Remove unnecessary flags in `download_and_prepare` script

### CODE REVIEW

File: tensorflow_datasets/core/dataset_builder.py [hunk 1]
```diff
@@ -166,6 +166,8 @@ class DatasetBuilder(object):
   # displayed in the dataset documentation.
   MANUAL_DOWNLOAD_INSTRUCTIONS = None
 
+  # This flag should be set to `True` if the dataset is defined outside tfds.
+  OUTSIDE_TFDS = False
```
Conchylicultor, line 170: I don't think we should add a new flag.
TFDS should automatically detect whether the checksums,... are defined, then fallback to default values otherwise.

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -205,6 +207,10 @@ def __init__(self, data_dir=None, config=None, version=None):
     else:  # Use the code version (do not restore data)
       self.info.initialize_from_bucket()
 
+    if self.OUTSIDE_TFDS:
+      path = os.path.dirname(inspect.getfile(self.__class__))
+      download.add_checksums_dir(path)
```
Conchylicultor, line 212: Adding a checksums dir may have a significant overhead if many datasets are loaded at the same time. In addition, checksums are memoized, so new directories will be ignored.

Instead, we should update `download_and_prepare` or `DownloadManager` to check whether some `checksums.tsv` is present locally, and load checksums from this file if so. Otherwise, fallback to legacy checksums.

tensorflow_datasets/core/download/download_manager.py
```diff
@@ -336,6 +339,10 @@ def _handle_download_result(
           url=resource.url, url_path=url_path, url_info=url_info)
     elif resource.url not in self._url_infos:
       if self._force_checksums_validation:
+        if tf.io.gfile.exists(self._checksums_path):
```
Conchylicultor, line 342: This doesn't seems to be the right place to load the checksums. We should move this when `self._url_infos` is initialised (if `checksums_path ` is set, then load from path, otherwise, load global checksums).

Currently, if the dataset is downloading 1000 images, it will check 1000 times for the path existance.

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -205,6 +205,8 @@ def __init__(self, data_dir=None, config=None, version=None):
     else:  # Use the code version (do not restore data)
       self.info.initialize_from_bucket()
 
+    self._code_dir = os.path.dirname(inspect.getfile(self.__class__))
```
Conchylicultor, line 208: Thank you. Would it be possible to add a test (create a dummy dataset with dummy checksums and dummy fake data) to make sure this works properly ?

Conchylicultor, line 208: It seems that we want to access the file location without having to instantiate the class, so this should be a class member (e.g. set through https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__, or through a memoized class property)

tensorflow_datasets/core/download/download_manager.py
```diff
@@ -336,6 +339,10 @@ def _handle_download_result(
...
+          url_infos = checksums._get_url_infos(self._checksums_path)  # pylint: disable=protected-access
```
Conchylicultor, line 343: `_get_url_infos` should be renamed `get_url_infos` if it is now part of the public API.

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -218,6 +218,12 @@ def canonical_version(self):
     else:
       return self.VERSION
 
+  @utils.classproperty
+  @classmethod
+  @utils.memoize()
+  def code_dir(cls):
```
Conchylicultor, line 224: One line docstring and typing annotations ?

tensorflow_datasets/core/download/download_manager.py
```diff
@@ -213,6 +215,11 @@ def __init__(
     self._register_checksums = register_checksums
     # All known URLs: {url: (size, checksum)}
     self._url_infos = checksums.get_all_url_infos()
+
+    if tf.io.gfile.exists(checksums_path):
```
Conchylicultor, line 219: This will raise an error if this if checksums_path is None

tensorflow_datasets/testing/dataset_builder_testing.py
```diff
@@ -308,8 +309,11 @@ def _test_checksums(self):
                "please add `SKIP_CHECKSUMS = True` to the "
                "`DatasetBuilderTestCase`")
 
-    with utils.try_reraise(suffix=err_msg):
+    filepath = os.path.join(os.path.dirname(
+        inspect.getfile(self.DATASET_CLASS)), 'checksums.txt')
```
Conchylicultor, line 313: `DATASET_CLASS.code_dir()`

tensorflow_datasets/testing/dummy_dataset/dummy_dataset_checksums.py
```diff
@@ -0,0 +1,4 @@
+from tensorflow_datasets.image_classification import cats_vs_dogs
+
+class DummyDatasetChecksums(cats_vs_dogs.CatsVsDogs):
```
Conchylicultor, line 3: Rather than inheriting from an existing dataset, I think it's best to write a small dataset from scratch. 
Otherwise, this introduce a dependency `CatsVsDogs` -> `core tests` which seems to break tests boundary.
For instance, it's not be clear here if the tests would be passing because the original `CatsVsDogs` pass, or if tests are actually using the new relative path.

tensorflow_datasets/testing/dummy_dataset/checksums.txt
```diff
@@ -0,0 +1 @@
+http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz 4542 f7ae60f92e00ec6debd23a6088c31dbd2371eca3ffa0defaefb259924204aec6
```
Conchylicultor, line 1: We also want to use a dummy url here (e.g. http://dummy.org/data.txt). Otherwise, the tests might still pass as url is registered in checksums.get_all_url_infos()

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -218,6 +218,13 @@ def canonical_version(self):
     else:
       return self.VERSION
 
+  @utils.classproperty
+  @classmethod
+  @utils.memoize()
+  def code_dir(cls) -> str:
```
Conchylicultor, line 224: After thinking more about this, it would be nice to rename this `code_path` and return a pathlib.Path.

It would allow to write `builder.code_path.name` or `builder.code_path.parent / 'checksums.txt'` which seems nicer.

I also think it should returns the python file, not the directory. Indeed, there are some few cases where getting the path will be helpful, like documenting the source code location, or for the refactoring script.

tensorflow_datasets/testing/dummy_dataset/dummy_dataset_checksums_test.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+from tensorflow_datasets import testing
+from dummy_dataset_checksums import DummyDatasetChecksums
+
+
+class DummyDatasetChecksumTest(testing.DatasetBuilderTestCase):
+  DATASET_CLASS = DummyDatasetChecksums
+  EXAMPLE_DIR = os.path.join(DummyDatasetChecksums.code_dir, "fake_examples")
```
Conchylicultor, line 9: As for checksums, this should also be automatically detected. The user shouldn't have to manually set this.

PR #2149, opened 2020-07-04
Update core/download docstrings

sharanramjee:
## Description

Updated docstrings and fixed old pylint errors in core/download files.

### DIFF
--- a/tensorflow_datasets/core/download/checksums.py
+++ b/tensorflow_datasets/core/download/checksums.py
@@ -38,7 +38,7 @@ _CHECKSUM_DIRS = [
 _CHECKSUM_SUFFIX = '.txt'
 
 
-class UrlInfo(object):  # TODO(tfds): Use dataclasses
+class UrlInfo:  # TODO(tfds): Use dataclasses
   """Small wrapper around the url metadata (checksum, size).
 
   Attributes:
@@ -51,7 +51,11 @@ class UrlInfo(object):  # TODO(tfds): Use dataclasses
     self.checksum = checksum
 
   def asdict(self) -> Dict[str, Any]:
-    """Returns the dict representation of the dataclass."""
+    """Returns the dict representation of the dataclass.
+
+    Returns:
+      Dict representation of the dataclass.
+    """
     # TODO(tfds): Replace by `dataclasses.asdict(self)`
     return {
         'size': self.size,
@@ -59,6 +63,11 @@ class UrlInfo(object):  # TODO(tfds): Use dataclasses
     }
 
   def __eq__(self, other) -> bool:
+    """Returns the boolean of whether the two UrlInfo objects are equal.
+
+    Returns:
+      Boolean of whether the two UrlInfo objects are equal.
+    """
     return (
         type(self) == type(other) and  # pylint: disable=unidiomatic-typecheck
         self.size == other.size and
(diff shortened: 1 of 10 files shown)


tensorflow_datasets/core/download/checksums.py
```diff
@@ -51,24 +51,43 @@ def __init__(self, size: int, checksum: str):
     self.checksum = checksum
 
   def asdict(self) -> Dict[str, Any]:
-    """Returns the dict representation of the dataclass."""
+    """Returns the dict representation of the dataclass.
```
Conchylicultor, line 54: What is the point of writing the same sentence twice ? It seems to me that this detailed description doesn't add any new information to the one-line description. It feels quite redundant.
In this case, it seems best to keep the one-line description

tensorflow_datasets/core/download/download_manager_test.py
```diff
@@ -112,6 +128,7 @@ def _extract(path, method, dest):
         extractor._Extractor, 'extract', side_effect=_extract).start()
 
   def setUp(self):
+    """Sets up the DownloadManager for the tests."""
```
Conchylicultor, line 131: Similarly for `__eq__`,... `setUp` is already documented in the parent class.

tensorflow_datasets/core/download/download_manager.py
```diff
@@ -241,12 +249,22 @@ def __getstate__(self):
 
   @property
   def _downloader(self):
+    """Returns the downloader object.
```
Conchylicultor, line 252: Here, a one-line docstring is nice, but typing annotation would be even more helpful

```py
@property
def _downloader(self) -> downloader.Downloader:
```
Is more helpful than
```py
@property
def _downloader(self):
"""Returns the downloader object."""
```

You don't have to change it there but in the future, adding typing annotation is more helpful than adding docstring.

tensorflow_datasets/core/download/checksums.py
```diff
@@ -38,7 +38,7 @@
 _CHECKSUM_SUFFIX = '.txt'
 
 
-class UrlInfo(object):  # TODO(tfds): Use dataclasses
```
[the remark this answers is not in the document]
sharanramjee: Alright. However, the TFDS website says that TFDS is supported for Python 3.5–3.8

tensorflow_datasets/core/download/checksums.py
```diff
@@ -51,24 +51,43 @@ def __init__(self, size: int, checksum: str):
...
     return {
         'size': self.size,
         'checksum': self.checksum,
     }
 
   def __eq__(self, other) -> bool:
+    """Returns the boolean of whether the two UrlInfo objects are equal.
+
+    Returns:
+      Boolean of whether the two UrlInfo objects are equal.
```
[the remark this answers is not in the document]
sharanramjee: Fixed,

tensorflow_datasets/core/download/checksums.py
sharanramjee: Agreed.

tensorflow_datasets/core/download/download_manager.py
sharanramjee: Sounds good.

tensorflow_datasets/core/download/download_manager_test.py
sharanramjee: Fixed.

tensorflow_datasets/core/download/checksums.py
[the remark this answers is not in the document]
Conchylicultor: Externally, we are only supporting Python 3. Internally, generation is Python 3 only, but some user still use Python2 to read datasets

PR #2168, opened 2020-07-09
Typo

lingt-xyz:
Change one dataset type typo: "translation" to "translate"


lingt-xyz:
@googlebot I signed it!

PR #2181, opened 2020-07-11
Fix librispeech dataset path to transcript

carynbear:
Transcript path is incorrect because it includes the directory path twice.

i.e. Path is supposed to be `example/path/to/Librispeech/*/*/*/*.txt` 
and not `example/path/to/example/path/to/Librispeech/*/*/*/*.txt`

### DIFF
--- a/tensorflow_datasets/audio/librispeech.py
+++ b/tensorflow_datasets/audio/librispeech.py
@@ -199,7 +199,7 @@ def _generate_librispeech_examples(directory):
   transcripts_glob = os.path.join(directory, "LibriSpeech", "*/*/*/*.txt")
   for transcript_file in tf.io.gfile.glob(transcripts_glob):
     path = os.path.dirname(transcript_file)
-    with tf.io.gfile.GFile(os.path.join(path, transcript_file)) as f:
+    with tf.io.gfile.GFile(transcript_file) as f:
       for line in f:
         line = line.strip()
         key, transcript = line.split(" ", 1)


tensorflow_datasets/audio/librispeech.py
```diff
@@ -199,7 +199,7 @@ def _generate_librispeech_examples(directory):
   transcripts_glob = os.path.join(directory, "LibriSpeech", "*/*/*/*.txt")
   for transcript_file in tf.io.gfile.glob(transcripts_glob):
     path = os.path.dirname(transcript_file)
-    with tf.io.gfile.GFile(os.path.join(path, transcript_file)) as f:
```
Conchylicultor, line 202: Could you split this PR in 2 (one for the new dataset and one for the librispeech.py fix) ?

Also I think it still works even without the fix as `transcript_file` is an absolute path and `os.path.join('/a/b/c', '/d/e/f') == '/d/e/f'`

Jaidon-Smith:
It is a shame this fix is still not on master. I was trying to generate librispeech on dataflow and was getting this error.

`"/home/jsjsrobert500/env/lib/python3.7/site-packages/tensorflow_datasets/core/tfrecords_writer.py", line 351, in _get_boundaries_per_bucket_shard raise AssertionError("Not a single example present in the PCollection!")AssertionError: Not a single example present in the PCollection! [while running 'train_other500_write/GetBoundaries']`


That one line change fixed the problem.

NikhilBartwal (changes requested):
@carynbear Thank you for the contribution!
There has not been any activity on this PR for about an year now, is this still a WIP?
If it is not, please consider closing this PR as it might lead to work duplication Thanks!

Issue #2190, opened 2020-07-13
Error loading imdb_reviews on Julyter Lab and Windows 10 machine


wbadry:
```python
import numpy as np
import tensorflow as tf
import tensorflow_hub as hub
import tensorflow_datasets as tfds

print("Version: ", tf.__version__)
print("Eager mode: ", tf.executing_eagerly())
print("Hub version: ", hub.__version__)
print("GPU is", "available" if tf.config.experimental.list_physical_devices("GPU") else "NOT AVAILABLE")

```

```
Version:  2.2.0
Eager mode:  True
Hub version:  0.8.0
GPU is available
```

**The datasets** list can be loaded with no issues. When trying to **load imdb_reviews** or any other dataset, I encounter this error

```python
#tfds.list_builders()
imdb, info = tfds.load('imdb_reviews', with_info=True, as_supervised=True)
```

```python
---------------------------------------------------------------------------
UnimplementedError                        Traceback (most recent call last)
c:\python37\lib\site-packages\tensorflow_datasets\core\utils\py_utils.py in try_reraise(*args, **kwargs)
    398   try:
--> 399     yield
    400   except Exception:   # pylint: disable=broad-except

c:\python37\lib\site-packages\tensorflow_datasets\core\registered.py in builder(name, **builder_init_kwargs)
    243       prefix="Failed to construct dataset {}".format(name)):
--> 244     return builder_cls(name)(**builder_kwargs)
    245 

c:\python37\lib\site-packages\wrapt\wrappers.py in __call__(self, *args, **kwargs)
    602             return self._self_wrapper(self.__wrapped__, self._self_instance,
--> 603                     args, kwargs)
    604 

c:\python37\lib\site-packages\tensorflow_datasets\core\api_utils.py in disallow_positional_args_dec(fn, instance, args, kwargs)
     68     _check_required(fn, kwargs)
---> 69     return fn(*args, **kwargs)
     70 

c:\python37\lib\site-packages\tensorflow_datasets\core\dataset_builder.py in __init__(self, data_dir, config, version)
    205     else:  # Use the code version (do not restore data)
--> 206       self.info.initialize_from_bucket()
    207 

c:\python37\lib\site-packages\tensorflow_datasets\core\dataset_info.py in initialize_from_bucket(self)
    422     tmp_dir = tempfile.mkdtemp("tfds")
--> 423     data_files = gcs_utils.gcs_dataset_info_files(self.full_name)
    424     if not data_files:

[...]

PR #2193, opened 2020-07-13
tf.io.gfile fails for gcs paths 

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/2190
CC: @Conchylicultor

### DIFF
--- a/tensorflow_datasets/core/utils/gcs_utils.py
+++ b/tensorflow_datasets/core/utils/gcs_utils.py
@@ -37,8 +37,10 @@ GCS_DATASET_INFO_DIR = 'dataset_info'
 GCS_DATASETS_DIR = 'datasets'
 
 
-_is_gcs_disabled = False
-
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/core/utils/gcs_utils.py
```diff
@@ -37,8 +37,10 @@
 GCS_DATASETS_DIR = 'datasets'
 
 
-_is_gcs_disabled = False
```
Conchylicultor, line 40: I think you could replace this by `_is_gcs_disabled = os.name == 'nt'`

rsepassi:
Merging shortly. Thank you!

tensorflow_datasets/core/utils/gcs_utils.py
vijayphoenix: That Global variable seemed unnecessary

Issue #2161, opened 2020-07-08
Add a TFDS user CLI


Conchylicultor:
It would be nice to have a user CLI which is installed at the same time as TFDS to pre-generate dataset:

```sh
tfds new my_dataset  # Generate the base template & cie
cd my_dataset/
tfds build  # Generate my_dataset
```

`tfds new` would create the following folder structure:
```
my_dataset/
    fake_data/
    __init__.py
    dataset.py
    dataset_test.py
    make_fake_data.py
    checksum.txt
```

Some technical details:
* Script would be implemented in python using argparse
* Is added through the `entry_points={'command'}` section in `setup.py`

This first require to implement the one folder==one dataset model.

acharles7:
can I work on this?

vijayphoenix:
Hi @acharles7,
Thank you for your interest.

FYI, I am already working on the bigger part of this issue, i.e, `one folder per one dataset`.

acharles7:
Thanks for letting me know, Let me know if you need any help:)

Conchylicultor:
Thank you for your interest!

@acharles7, @vijayphoenix I think this bug and https://github.com/tensorflow/datasets/issues/2139 are complementary and both could be worked in parallel.

@acharles7, If you want to get started, you could start by splitting this task in small self-contained PRs:

create a new `scripts/cli/main.py` folder and add a dummy script.

* **PR 1**: Add a dummy `scripts/cli/main.py` script:
   * It should be registered in `setup.py` `entry_points`.
   * We want to use both `absl.app` with `argparse`. See https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/documentation/build_catalog.py#L150 for an example
   * The script could be executed, but wouldn't do anything

Then each new command could be added in a new PR:

* **PR 2**: Add the `tfds new <dataset>` command:
  * Each command should be implemented as a separate file, (here `scripts/cli/new.py`)
  * This would generate the `<dataset>/` folder structure
  * The implementation can be inspired by: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/create_new_dataset.py
  * We want to use `pathlib.Pathlib` instead of `GFile`

acharles7:
@Conchylicultor Thanks, working on it (PR1), I'll let you if I have any questions.

acharles7:
Seems like @vijayphoenix already made architecture for CLI. Thanks
So, I'll add required new commands for this cli.

[...]

PR #2194, opened 2020-07-14
Add new tfds cli command

vijayphoenix:
Related to #2161


tensorflow_datasets/scripts/cli/new_dataset.py
```diff
@@ -0,0 +1,256 @@
...
+  checksum_path = os.path.join(root_dir, 'checksums.txt')
+  with tf.io.gfile.GFile(checksum_path.format(**data), 'w') as f:
+    f.write(_CHECKSUM_FILE.format(**data))
+
+
+def create_single_dataset(ds_name: str) -> None:
+  root_dir = os.path.abspath(ds_name)
+  tf.io.gfile.makedirs(os.path.join(ds_name, 'fake_examples'))
+
+  data = dict(
```
Conchylicultor, line 217: We should use dataclasses now

tensorflow_datasets/scripts/cli/new_dataset.py
```diff
@@ -0,0 +1,256 @@
...
+  for ds_name in args.dataset_name:
+    create_single_dataset(ds_name)
+
+
+def add_new_dataset_parser(subparsers: argparse._SubParsersAction) -> None:  # pylint:disable = protected-access
+  """Add subparser for `new` command"""
+  new_ds_parser = subparsers.add_parser(
+      'new', help='Generated new dataset(s) files')
+
+  new_ds_parser.add_argument(
```
Conchylicultor, line 248: I think it would be helpful to add a `--dst_dir` flag of type `pathlib.Path` which default to Path.cwd()

tensorflow_datasets/scripts/cli/new_dataset.py
vijayphoenix: Dataclasses seems to complicate the template generation as we format the string using `**data`.

Conchylicultor: But not if we use f-string, no ?

```
file_path.write_text(textwrap.dedent(f"""\
from tensorflow.dataset.{data.name} import {data.cls_name}
"""))
```

vijayphoenix:
Merged by 0127735eb2ba851a92de9a3759420f7c4fada7c4

PR #2218, opened 2020-07-21
Add an option for the ImageNet 2012 Challenge TEST split to TFDS.

tfds-copybara:
Add an option for the ImageNet 2012 Challenge TEST split to TFDS.


Conchylicultor:
Thanks for the warning shekkizh. So you know if all split have been updated ? And do you know what are the difference between the two tests archives ?

shekkizh:
My understanding is that only the test image file was updated. This is the warning from the ImageNet download page.
image

Issue #2138, opened 2020-06-30
Config based loading


vijayphoenix:
Currently, reading a dataset requires the features to be defined inside the `_info` method.
If the structure of the feature (key names, feature types,...) was saved along the generated tf-records, it would make it possible to a dataset without requiring the dataset code to be present.
This is related to https://github.com/tensorflow/datasets/issues/1186

**Advantage:**
- Would unlock backward compatibility (the previous dataset can still be read from the config, even if the dataset code don't exist anymore)
- Would help with cross-language support (other languages could load the dataset from the config)
- It would allow TFDS to read datasets without requiring the original generation code.

**Design questions:**

**1) Which format to save the feature connectors?**
We might want to save this as a JSON file.

**2) Should this be saved in the dataset_info.json or Separately?**
We want to save this information separately because of backward compatibility issues.

**3) Where to save?**
We can save this info as `feature_info.json` inside `data_dir/dataset/version`.

**4) What to save?**
We will dump the `FeatureConnector` of the dataset while writing the files.

**5) How to read?**
We will read the `feature_info.json` instead of using the features in the `_info` method.

vijayphoenix:
Completed by https://github.com/tensorflow/datasets/commit/358b069bbb83ae6f7a0cb7bd65940c0c0addf283

PR #2221, opened 2020-07-22
Load and dump FeatureConnectors

vijayphoenix:
Related to https://github.com/tensorflow/datasets/issues/2138


vijayphoenix:
TODO:
- [ ] Implement `from_json` methods and add tests
- [ ] Implement `to_json` methods for `text` and `translation` features.

tensorflow_datasets/core/features/feature.py
```diff
@@ -205,6 +213,44 @@ def dtype(self):
...
+    feature_dict = {}
+    for feature_name, feature in value['content']:
+      subclass = cls._REGISTERED_FEATURES.get(feature_name)
+      if subclass is None:
+        raise ValueError(
+            f'Invalid feature name {feature_name} with feature: {feature}\n'
+            f'Supported: {list(cls._REGISTERED_FEATURES)}'
+        )
+      feature_dict.update(subclass.from_json(feature))
+    return feature_dict
```
Conchylicultor, line 229: Just for info, you're returning a `dict`, but the return type should be `FeatureConnector`

tensorflow_datasets/core/features/audio_feature.py
```diff
@@ -107,6 +107,19 @@ def repr_html(self, ex: np.ndarray) -> str:
         ' controlsList="nodownload" />'
     )
 
+  @classmethod
+  def from_json(cls, value) -> 'FeatureConnector':
+    pass
+
+  def to_json(self):
+    return {
+        'type': type(self).__name__,
```
Conchylicultor, line 116: The `type` field seems to be duplicated across all features. As this is how feature class are detected, I think it is important to automate this field.

I think we should have two method `from_json` and `from_json_content`. Here is my proposal:

The `FeatureConnector` implement the 4 methods:

```py
def from_json(cls, value) -> 'FeatureConnector':
  subclass = cls._REGISTERED_FEATURES.get(value['type'])
  return subclass.from_json_content(value['content'])

def to_json(self, value):
  return {
      'type': type(self).__name__,
      'content': self.to_json_content(),
  }
  
  

def from_json_content(cls, value):
  if value:
    raise ValueError(
        f'FeatureConnector {cls} should overwrite `from_json_content`'
    )
  return cls()

def to_json_content(cls, value):
  return {}
```

Then subclasses would only implement the `to_json_content` and `from_json_content` optionally.

```py
def to_json(self):
  return {
        'file_format': self._file_format,
        'shape': list(self._shape),
        'dtype': self._dtype.name,
        'sample_rate': self._sample_rate,
    }
```

tensorflow_datasets/core/features/audio_feature_test.py
```diff
@@ -66,6 +66,21 @@ def test_numpy_array_float(self):
         ],
     )
 
+  def test_to_json(self):
```
Conchylicultor, line 69: Would it be possible to some-how automate this test inside `assertFeature` ?

Rather than checking individual fields, I think we could check that exporting/loading the feature yield the same results:

```py
def assertFeature(feature, *kwargs):
  with subtest('feature'):
    _assert_feature(feature, *kwargs)
  with subtest('exported_imported'):
    new_feature = FeatureConnector.from_json(feature.to_json())
    _assert_feature(new_feature, *kwargs)
```

This way all features would be tested automatically. And we don't need to test the intermediate representation. If the reconstructed feature works, it's good !

tensorflow_datasets/core/features/audio_feature.py
Conchylicultor: Maybe the `from_json_content` in the base class could be simplified even further:

```py
# In features.py
def from_json_content(cls, value):
  return cls(**value)
```

This way, many FeatureConnector wouldn't have to implement `from_json_content` at all.

tensorflow_datasets/core/features/audio_feature_test.py
vijayphoenix: Added a similar test.
```
  def _assert_feature(self, feature):
    new_feature = feature.from_json_content(feature.to_json_content())
    self.assertEqual(str(feature), str(new_feature))
```

vijayphoenix:
@Conchylicultor Should I add `to_json` and `from_json` methods for `TextEncoder`, `Tokenizer`, `TextEncoderConfig`?

Conchylicultor:
> @Conchylicultor Should I add `to_json` and `from_json` methods for `TextEncoder`, `Tokenizer`, `TextEncoderConfig`?
> (As we are removing the encoders from all datasets in PR #2244)

No, as those features are deprecated, no need to update them.

tensorflow_datasets/core/features/audio_feature_test.py
Conchylicultor: > Added a similar test.

This test is good, but feel fragile and don't really test that the feature is correctly exported/imported.

For instance, with ClassLabel, your implementation would not test that label names are correctly restored.

tensorflow_datasets/core/features/features_dict.py
```diff
@@ -171,6 +171,20 @@ def get_serialized_info(self):
         for feature_key, feature in self._feature_dict.items()
     }
 
+  @classmethod
+  def from_json_content(cls, value) -> 'FeatureConnector':
+    features = dict()
+    for feature_key, feature_value in value.items():
+      subclass = cls._REGISTERED_FEATURES.get(feature_value['type'])
```
Conchylicultor, line 178: Subclasses should not use the parent internal variable. Don't the following works ?

```py
return cls({
    k: FeatureConnector.from_json(v)
    for k, v in value.items()
})
```

tensorflow_datasets/core/features/text_feature.py
```diff
@@ -174,3 +174,12 @@ def repr_html(self, ex: bytes) -> str:
     ex = html.escape(ex)
     ex = textwrap.shorten(ex, width=1000)  # Truncate long text
     return ex
+
+  @classmethod
+  def from_json_content(cls, _) -> 'FeatureConnector':
```
Conchylicultor, line 179: Same comment here

tensorflow_datasets/core/features/sequence_feature.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, feature, length=None, **kwargs):
     # Convert {} => FeaturesDict, tf.int32 => Tensor(shape=(), dtype=tf.int32)
     self._feature = features_dict.to_feature(feature)
     self._length = length
+    self._kwargs = kwargs
```
Conchylicultor, line 91: This is always empty. I would replace this by something like:

```
assert not kwargs, 'Json export/import should be updated'
```

tensorflow_datasets/core/features/sequence_feature.py
```diff
@@ -204,6 +205,30 @@ def __repr__(self):
...
+        kwargs.update({key: value})
+    return cls(features, length, **kwargs)
+
+  def to_json_content(self):
+    if isinstance(self.feature, features_dict.FeaturesDict):
+      value = {k: v.to_json() for k, v in self.feature.items()}
+    else:
+      value = self.feature.to_json()
+      value['arg_'] = 'feature'
+    return {'length': self._length, ** value, **self._kwargs}
```
Conchylicultor, line 231: I don't understand why this need to be so complicated. Why not

```py
def to_json_content(self) -> Json:
  return {
      'length': self._length,
      'feature': self.feature.to_json(),
  }

def from_json_content(cls, value: Json) -> 'Sequence':
  return cls(FeatureConnector.from_json(value['feature']), length=value['length'])
```

Conchylicultor:
Just a few minor polishing.

tensorflow_datasets/core/features/sequence_feature.py
vijayphoenix: Done

Issue #2199, opened 2020-07-15
ImportError: cannot import name 'ContextManager'


SinaChavoshi:
**Short description**
tfds 3.2.0 requires later version of typing 3.6.0>= which includes support for "ContextManager" (ref https://docs.python.org/3/library/typing.html#typing.ContextManager). However this is not specified in requirements file resulting in failure python 3.5.2 environment. 

**Environment information**
* Operating System: ubuntu 18.0
* Python version: 3.5.2 
* `tensorflow-datasets` version: 3.2.0
* `tensorflow` version: 2.2

**Reproduction instructions**

setup an environment with python 3.5.2. 

```shell
pip3 install tensorflow 
python -c "import tensroflow_datasets as tfds" 
```

**Link to logs**
If applicable, <link to gist with logs, stack trace>

**Expected behavior**
What you expected to happen.
tfds should pull required dependancies

**Additional context**

vijayphoenix:
Fixed by #2222

PR #2222, opened 2020-07-22
Set minimal supported Python version in setup.py

lgeiger:
As mentioned in #2199 TFDS requires Python 3.6 or newer. This PR correctly sets [`python_requires`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) in `setup.py` in order to raise a proper proper error message when trying to install TFDS on older versions of Python.

Adresses #2199 
/cc @Conchylicultor

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -187,6 +187,7 @@ setup(
     },
     scripts=[],
     install_requires=REQUIRED_PKGS,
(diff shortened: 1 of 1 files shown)


Conchylicultor:
Thank you for your contribution. Merging now

Issue #2133, opened 2020-06-29
[GSoC] Better auto-sharding in multi-worker pipelines


Conchylicultor:
Currently, when the dataset only has a single shard with many workers (`info.splits['train'].num_shards < read_config.input_context.num_input_pipelines`).
We're raising an error in: `tensorflow_datasets/core/tfrecords_reader.py`

```
ValueError: Cannot shard the pipeline with given `input_context`.`num_shards=1` but `num_input_pipelines=2`. 
This means that some workers won't read any data. To shard the data, you may want to use the subsplit API
instead: https://www.tensorflow.org/datasets/splits
```

```py
options = tf.data.Options()
options.experimental_distribute.auto_shard_policy = tf.data.experimental.AutoShardPolicy.OFF

read_config = tfds.ReadConfig(
   input_context=input_context,
   options=options,
)
ds = tfds.builder(..., split='train', read_config=read_config)
```
Currently, the user should manually use the subsplit API instead: `tfds.load(..., split=f'train[{from_}%:{to}%]')`
Where `from_` and `to` are different values per workers.

This isn't a very user-friendly workflow. We likely should:

* Provide an util to auto-generate subsplits: `tfds.subsplit('train', num_subsplits=2, subsplit_id=0) == 'train[:50%]')`.
* Auto-apply the subsplit API if `num_shards < num_pipelines` and `input_context` is provided.

PR #2278, opened 2020-08-05
Add subsplit API for Better auto-sharding

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/2133
TODO: 
- [x] Auto-apply the subsplit API if `num_shards` < `num_pipelines` and `input_context` is provided.

Usage Example:

```python
ds = tfds.load('mnist',
               split=tfds.subsplit(split_name='train', num_subsplits=2),
               read_config=read_config)

```


Conchylicultor:
For info, your second item is already implemented in https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/tfrecords_reader.py#L190

It only supports case where the dataset as multiple shards. For other cases, it's possible that it should be fixed from the TF side.

Issue #2157, opened 2020-07-07
[data request] BAIR Robot Pushing Dataset (original resolution)


martinengelcke:
* Name of dataset: BAIR Robot Pushing Dataset
* URL of dataset: https://sites.google.com/view/sna-visual-mpc/
* License of dataset: NA
* Short description of dataset and use case(s): The "small" version of this dataset is already available at https://www.tensorflow.org/datasets/catalog/bair_robot_pushing_small but only supports 64x64 images. It would be great to be able to load this dataset at the original resolution or a custom resolution as the 64x64 resolution is very limiting.

Folks who would also like to see this dataset in `tensorflow/datasets`, please thumbs-up so the developers can know which requests to prioritize.

And if you'd like to contribute the dataset (thank you!), see our [guide to adding a dataset](https://github.com/tensorflow/datasets/blob/master/docs/add_dataset.md).

rushabh-v:
I would like to contribute it.

Should I implement it for custom resolution or If fixed then what resolution should the dataset support, optimally? 
The paper [Self-Supervised Visual Planning with Temporal Skip Connections](https://arxiv.org/pdf/1710.05268.pdf) has used 64 x 64 frames only.

martinengelcke:
Thanks for your interest!

Looking at the paper code (https://github.com/febert/visual_mpc/blob/master/python_visual_mpc/video_prediction/read_tf_record.py), my first impression was that the original data supported up to 128x128. The documentation for bair_robot_pushing_small also sounds like frames at resolutions higher than 64x64 are in principle available, but I am not 100% sure this is actually true.

A data loader for custom resolution would be ideal but involve more effort. As a start, 128x128 - if possible - would be great and make visual details a lot more discernible.

rushabh-v:
I see [here](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/testing/fake_data_generation/bair_robot_pushing.py#L54-L56) that the fake dataset for `bair_robot_pushing_small` was created for size `64 x 64`. 
So, should I just replace that data with `128 x 128` and then explicitly set the path of `fakedata` in the testing of one of the dataset builders or should I generate a different fake dataset for it?

martinengelcke:
Maybe generate a separate one?

rushabh-v:
I have opened a PR. Can you please review it @martinengelcke ? https://github.com/tensorflow/datasets/pull/2279

martinengelcke:
[...]

PR #2279, opened 2020-08-06
add bair robot pushing 128 x 128 dataset

rushabh-v:
# Add Dataset

* Dataset Name: bair_robot_pushing
* Issue Reference: https://github.com/tensorflow/datasets/issues/2157
  
## Description

<description>
  
## Checklist
* [x] Address all TODO's
* [ ] Run `download_and_prepare` successfully
* [ ] Add passing test(s)
* [x] Add test data
* [x] If using additional dependencies (e.g. `scipy`), use [lazy_imports](https://www.tensorflow.org/datasets/add_dataset#extra_dependencies) (if applicable)
* [x] [Lint](https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style) code


martinengelcke:
Thanks a lot @rushabh-v for putting together this PR. Looks good at first glance. How about changing the class name from BairRobotPushing to BairRobotPushing128 to be more explicit? That would also match the filename where the class if defined. Any issues with running `download_and_prepare`?

rushabh-v:
Sure, I have updated the name to "bair_robot_pushing128".

Just allow me some time for `download_and_prepare`. I tried to do it on google colab but it exceeded the memory limits. So, I am trying it locally now!

rushabh-v:
Sorry for the delay @martinengelcke. 

I tried running `download_and_prepare`. But it gave an error that size 12288 can not be reshaped into 128 x 128. And `12288 = (64 * 64 * 3)`. Am I making any mistake?

martinengelcke:
Hm. It sounds like images from DATA_URL = "http://rail.eecs.berkeley.edu/datasets/bair_robot_pushing_dataset_v0.tar" are 64x64x3 and therefore can't be reshaped?

In #2157 you said it looks like they are 512x640x3. How did you obtain that number?

rushabh-v:
I tried to analyze a tfrecord file from the original dataset, by converting it to frames and obtained that number. I have lost the script I used, so I am trying that again (in case I made any mistake).

rushabh-v:
I checked in the paper and they also used `64 x 64` frames. Search for "resolution" [here](https://www.arxiv-vanity.com/papers/1610.00696/). What I don't understand though is that why the name of the current dataset is `bair_robot_pushing_small`. Should I open a PR to rename it to `bair_robot_pushing`? @martinengelcke

rushabh-v:
As a confirmation, I used the script [here](https://github.com/edenton/svg/blob/master/data/convert_bair.py) to convert the tfrecord files of the dataset into frames. And found that they are indeed `64 x 64`. Trying to convert them to `128 x 128` resulted into `ValueError: not enough image data`.

martinengelcke:
Ah ok, nicely done for figuring this out.

Yeah I agree, the name of the dataset is indeed misleading. I think creating a PR to rename the dataset is a good idea. I closed #2157 and you can also close this PR.

Thanks for your work on this :)

rushabh-v:
Closing the PR! Will open a new one to rename `bair_robot_pushing_small` to `bair_robot_pushing`.

PR #2284, opened 2020-08-07
Show Supported Python versions

vijayphoenix:
We should show the supported Python version in [pypi classifiers](https://pypi.org/project/tensorflow-datasets/) and [README.md](README.md)


setup.py
```diff
@@ -194,6 +194,11 @@
         'Intended Audience :: Developers',
         'Intended Audience :: Science/Research',
         'License :: OSI Approved :: Apache Software License',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3 :: Only',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
```
Conchylicultor, line 201: Does this mean we'll have to update this every time a new version came out ? Is there a way to indicate `3.6+` instead of listing all individual versions ?

Conchylicultor: Or otherwise, we can we just keep the `Programming Language :: Python :: 3` ?

vijayphoenix: I could not find any `3.6+` option [here](https://pypi.org/classifiers/)

So, unfortunately, if we want users to know, tfds only supports 3.6 or higher, we will have to list all python versions.

Conchylicultor: Thanks for confirming. In this case, it seems best to only list Python 3 rather than individual versions. I'll do the change when merging internally.

PR #2280, opened 2020-08-06
Fix typo in dataset markdown builder

ChAnYaNG97:
According to the [code](https://github.com/tensorflow/datasets/blob/cf76046da843b997caa1e33855b546c67bb40c5a/tensorflow_datasets/core/dataset_builder.py#L768) and [comments](https://github.com/tensorflow/datasets/blob/cf76046da843b997caa1e33855b546c67bb40c5a/tensorflow_datasets/core/dataset_builder.py#L280) here.

PR #2285, opened 2020-08-07
Fix wrong default path of the manual download directory in documentation

MiWeiss:
The manual download dir in the documentation is not correct, an 's' is missing: `download/manual` should be `downloads/manual`

Note: [dataset_builder.py#L768](https://github.com/michaelsresearch/datasets/blob/master/tensorflow_datasets/core/dataset_builder.py#L768) seems to be the corresponding code that defines the default dir.


MiWeiss:
Just found this: Someone else was faster ;-)
https://github.com/tensorflow/datasets/pull/2280

MiWeiss:
@googlebot I signed it!

Issue #2299, opened 2020-08-13
Add script to export tensorflow_data_validation


Conchylicultor:
Currently, there is an option to generate TFDV statistics for [FACET](https://pair-code.github.io/facets/) during `download_and_prepare`. However, it would be best to separate those two steps.

```python
builder = tfds.builder('dataset')
builder.download_and_prepare()
builder.generate_statistics()  # Generate TFDV statistics using tfdv in `dataset/config/version/statistics.json`
```

Then we should add a new script in `scripts/generate_statistics.py` which generate statistics for one dataset and export them to `gs://`. We would automatically generate statistics for all datasets.
When user pre-load the `DatasetInfo`, it would also fetch the `statistics.json` from GCS, so user could visualize FACET for all default TFDS datasets with the existing `tfds.show_statistics`.

devil-cyber:
hi, is this issue still open can I work on this?

Conchylicultor:
I think this could still be worked on. As a first step, I think we could add a standalone script/function `generate_statistics(builder, dst_dir)` which generate the statistics and save them in `dst_dir / statistics.json`.

This way we don't add any more complexity in the core API. Then we can think more about better exposing the statistics directly inside the core API.

PR #2301, opened 2020-08-13
Add generate_statistics function

vijayphoenix:
Related to https://github.com/tensorflow/datasets/issues/2299

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -19,6 +19,7 @@ import abc
 import functools
 import inspect
 import itertools
+import json
 import os
 import sys
 from typing import Any, Optional
@@ -164,7 +165,6 @@ class DatasetBuilder(object):
   # the list of files that should be present. It will be
   # displayed in the dataset documentation.
   MANUAL_DOWNLOAD_INSTRUCTIONS = None
-
 
   def __init__(self, *, data_dir=None, config=None, version=None):
     """Constructs a DatasetBuilder.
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/dataset_builder.py
```diff
@@ -383,33 +384,45 @@ def download_and_prepare(self, *, download_dir=None, download_config=None):
...
-                download_config.compute_stats)
-          else:  # Mode is forced or stats do not exists yet
-            logging.info("Computing statistics.")
-            self.info.compute_dynamic_properties()
           self.info.download_size = dl_manager.downloaded_size
-          # Write DatasetInfo to disk, even if we haven't computed statistics.
           self.info.write_to_directory(self._data_dir)
     self._log_download_done()
 
+  def generate_statistics(self, compute_stats=None):
```
Conchylicultor, line 391: Calling `generate_statistics` should always generate statistics, so I would remove `compute_stats` kwargs.

Like for `builder.download_and_prepare`, `generate_statistics` should auto-detect whether statistics have already been downloaded.

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -383,33 +384,45 @@ def download_and_prepare(self, *, download_dir=None, download_config=None):
...
+        statistics over the generated data. Defaults to `AUTO`.
+
+    Note: Requires `tensorflow_data_validation`.
+    """
+    compute_stats = download.ComputeStatsMode(compute_stats or
+                                              download.ComputeStatsMode.AUTO)
+
+    splits = list(self.info.splits.values())
+    statistics_already_computed = bool(
+        splits and splits[0].statistics.num_examples)
```
Conchylicultor, line 405: To check whether statistics are already computed, we should check for the presence or not of statistics.json

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -383,33 +384,45 @@ def download_and_prepare(self, *, download_dir=None, download_config=None):
...
+
+    if (compute_stats == download.ComputeStatsMode.SKIP or
+        compute_stats == download.ComputeStatsMode.AUTO and
+        statistics_already_computed):
+      logging.info("Skipping computing stats for mode %s.", compute_stats)
+    else:
+      logging.info("Computing statistics.")
+      import tensorflow_data_validation  # pylint: disable=g-import-not-at-top,import-outside-toplevel,unused-import  # pytype: disable=import-error
+      self.info.compute_dynamic_properties()
```
Conchylicultor, line 414: If we want to better layer this feature. Currently, statistics are split between visualization, dataset_info, dataset_builder.

I would organize the code as follow, we should try to move most of the code inside a new `statistics.py` objet:

* `core/statistics.py` with:

```py
class StatisticsInfo:

  def __bool__() -> bool:  # Return True if statistics are present, false otherwise.

  def compute_and_save():

  def _repr_html():  # Display the FACET display

  ...
```

* `core/dataset_info`:

```py
class DatasetInfo:

  def __init__():
    self.statistics = StatisticsInfo(self)
```

* `core/dataset_builder`:

```py
def generate_statistics():
  if not info.statistics:
    info.statistics.compute_and_save()
```

To help with the implementation, I think both the current statistics code and the new stats can be kept for now, so you can just develop the new feature in some `info._statistics`. Once complete, we can delete the old code and have `info.statistics` returns `info._statistics`.

tensorflow_datasets/core/statistics.py
```diff
@@ -0,0 +1,133 @@
...
+                       "doesn't seem to have been generated. Please ensure "
+                       "that the data was downloaded for this split and re-run "
+                       "download_and_prepare."), self._info.name, split_name)
+        raise
+
+    # Set splits to trigger proto update in setter
+    self._info._set_splits(splits)  # pylint: disable = protected-access
+
+  def _repr_html_(self):  # Display the FACET display
+    pass
```
vijayphoenix, line 81: This should have `show_statistics` code, right?

tensorflow_datasets/core/statistics.py
```diff
@@ -0,0 +1,133 @@
...
+import tensorflow.compat.v2 as tf
+
+from tensorflow_datasets.core import dataset_builder
+from tensorflow_datasets.core import lazy_imports_lib
+from tensorflow_datasets.core import naming
+from tensorflow_datasets.core import splits as splits_lib
+from tensorflow_datasets.core import utils
+
+
+class StatisticsInfo:
```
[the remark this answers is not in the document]
vijayphoenix: No problem 🙂
Will work on something else in the meantime.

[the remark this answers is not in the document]
Conchylicultor: As I started looking into this, it seems that the reason why the `SplitInfo` is complicated was because the way `DatasetInfo` was written.

Currently, we need to keep `info.splits` and `info._info_proto` in sync, so updating `info.splits` require to notify `info._info_proto`. This makes the code hard to follow.
It's my fault for implementing this like that.

We should update `DatasetInfo` to be a `dataclasses.dataclass(frozen=True)`, and rather than using the proto as ground truth, use the DatasetInfo attributes. So we can update `info.splits` without having to update the proto.

So 
```py
class DatasetInfo:
  ...

  @property
  def as_proto(self):
    return self._info_proto

  @property
  def name(self):
    return self.as_proto.name
```
would become:
```py
@dataclasses.dataclass(frozen=True)
class DatasetInfo:
  ...

  name: str

  @property
  def as_proto(self):
    return dataset_info_pb2.DatasetInfo(
        name=self.name,
    )
```

There wouldn't be `_info_proto `, and `as_proto` would only be called when saving in `write_to_directory`, instead of for every attributes

Some implementations details:
* With `(frozen=True)`, because info is mutable, we can use `object.__setattr__(self, 'name', value)` when we need to update dataset info, like inside `read_from_directory`.
* The current __init__ would be changed as post_init to cleanup the arguments:


```py
@dataclasses.dataclass(frozen=True)
class _DatasetInfo:
  
  def __post_init__(self):
    object.__setattr__(self, 'version', self.builder.version)
    if self.description:
      object.__setattr__(self, 'description', utils.dedent(self.description))
```

You can look into this if you want but it's possible that I missed something again which would make this more complicated than expected.

PR #2317, opened 2020-08-18
Fix bug where xquad and tydi qa kept the same context for all questions.

tfds-copybara:
Fix bug where xquad and tydi qa kept the same context for all questions.


vijayphoenix:
Merged by 1a35ca63 and https://github.com/tensorflow/datasets/commit/d36337da4e1cfcc4b10524e498181ad725738d16

PR #2322, opened 2020-08-19
Load dataset without reading its generation code

vijayphoenix:
Related https://github.com/tensorflow/datasets/issues/2138


Conchylicultor (changes requested):
Thank you for starting this.

tensorflow_datasets/core/load.py
```diff
@@ -332,6 +373,32 @@ def load(
   return ds
 
 
+def _get_builder_kwargs(path: str) -> Tuple[str, str, Optional[str], str]:
```
Conchylicultor, line 376: It feels strange that the Builder require the parent dir.
Is it not possible to overwrite `def _build_data_dir(self, given_data_dir):` to return `(given_data_dir, given_data_dir)` instead ?
Similarly, instead of extracting the version and config from the path, we should try to load those info from the dataset_info.json

Conchylicultor (changes requested):
Thank you!

tensorflow_datasets/core/load.py
Conchylicultor: Maybe at first you can try to fully overwrite __init__ and not call super() at all. Then we can see which modification would be required to DatasetBuilder.__init__ to support both regular and config-based loading.

tensorflow_datasets/core/load.py
```diff
@@ -111,6 +113,23 @@ def __init__(self, name, is_abstract=False, in_development=False):
     ValueError.__init__(self, error_string)
 
 
+class ConfigBuilder(dataset_builder.FileReaderBuilder):
```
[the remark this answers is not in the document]
vijayphoenix: Should I move this to `dataset_builder.py`?

tensorflow_datasets/core/load.py
vijayphoenix: @Conchylicultor, I could not find any config related in `dataset_info.py`.
Am I missing out on something?

Conchylicultor (approved):
Thank you Vijay! Looks good!

Conchylicultor (changes requested):
I forgot to ask but could you add tests to `ReadOnlyBuilder` ?

Conchylicultor (approved):
Thank you for the quick change

Conchylicultor (changes requested):
Actually it would be nice if there was also a test for a Builder with config to make sure `info.full_names` & cie work properly too. Sorry for the noise

vijayphoenix:
I was also going to add it but got distracted by Kokoro test fails :sweat_smile:

Conchylicultor (approved):
Awesome! I've imported this internally and this should be merged soon!

vijayphoenix:
Merged by https://github.com/tensorflow/datasets/commit/262bbda5479301815b475760854228aed0d4af77

Conchylicultor:
Note: I've splitted internally the PR in smaller parts, so the main code still hasn;t been merged.

PR #2334, opened 2020-08-24
fix return type of tfds.load in doc

HermanBovens:
Was referring to `tfds.data.Dataset` instead of `tf.data.Dataset`.


HermanBovens:
@googlebot I signed it!

Conchylicultor (approved):
Thanks for the fix

Issue #2352, opened 2020-08-28
Add a script to detect dead urls


Conchylicultor:
Some download urls are not valid anymore.

The checksums contains the list of all downloaded urls. We could add a script which iterate over all url and test if the website is still online with some request query.

To get all urls:

1. Use `tfds.core.download.checksums.get_all_url_infos()` to get urls in `checksums/`
2. For new datasets self-contained in folder, checksums are now in `<my_dataset>/checksums.tsv` to extract dataset in those, Iterate over all `builder_cls` (`[tfds.builder_cls(name) for name in tfds.list_builder()]`, the use `builder_cls.url_infos` to get the url list)

PR #2354, opened 2020-08-28
Script to detect Dead URLs

vijayphoenix:
Fix https://github.com/tensorflow/datasets/issues/2352

TODO:
- [ ] Detect Dead URLs in folder datasets


tensorflow_datasets/scripts/dead_url_checker.py
```diff
@@ -0,0 +1,43 @@
+r"""Dump the list of all dead urls in a `.txt` file.
```
Conchylicultor, line 1: This should go in `cleanup/` which contain our maintenance scripts.

tensorflow_datasets/scripts/dead_url_checker.py
```diff
@@ -0,0 +1,43 @@
...
+
+def _is_dead_url(url: str) -> bool:
+  return requests.head(url).status_code != 200
+
+
+def main(_):
+  dead_urls = []
+  for url in tfds.core.utils.tqdm(
+      tfds.core.download.checksums.get_all_url_infos()):
+    if _is_dead_url(url):
```
Conchylicultor, line 33: There is a lot of URLs to check, we should use ThreadPoolExecutor from https://docs.python.org/3/library/concurrent.futures.html to parallelize the checks

tensorflow_datasets/scripts/dead_url_checker.py
```diff
@@ -0,0 +1,43 @@
...
+      dead_urls.append(url)
+
+  path = os.path.join(FLAGS.tfds_dir, 'dead_urls.txt')
+  with tf.io.gfile.GFile(path, 'w') as f:
```
Conchylicultor, line 37: Let's not save the urls. Only print it for now.

Conchylicultor (changes requested):
Thank you for the fast implementation

Conchylicultor (approved):
Thank you Vijay!

PR #2387, opened 2020-09-02
Fix test discovery

vijayphoenix:
Since we are adding `.vscode` folder to `datasets/tensorflow_datasets` instead of `datasets`, the following changes are needed to fix test discovery.


tensorflow_datasets/.vscode/settings.json
```diff
@@ -19,11 +19,10 @@
...
-        "tensorflow_datasets",
         "-vv",
         "--continue-on-collection-errors",
-        "--ignore=tensorflow_datasets/testing/test_utils.py",
-        "--ignore=tensorflow_datasets/scripts/documentation/build_api_docs_test.py"
+        "--ignore=testing/test_utils.py",
+        "--ignore=scripts/documentation/build_api_docs_test.py"
     ],
     "files.watcherExclude": {
         "**/.git/**": true,
```
vijayphoenix, line 28: As `.git/` and `docs/` files are in the `datasets` folder, so maybe we can remove here.

Conchylicultor:
Thank you for catching this. I mistakenly moved the `.vscode/` in the wrong directory.

@vijayphoenix, As you have more experience working with VSCode, do you think it would be better to have `.vscode/`, `.pylintrc`,... in `datasets/.vscode` or to keep it in `datasets/tensorflow_datasets/` ?

I feel having it at the top level might be more discoverable ?

vijayphoenix:
Yes, it would be better to move all the additional files in the `datasets` folder.

vijayphoenix:
Fixed by https://github.com/tensorflow/datasets/commit/3e3fc7ccf8e41e1b00ed5be56cf51f96908b31a7

PR #2388, opened 2020-09-02
Fix MockFs exists function bug

vijayphoenix:
**Bug**:
Returns True, only if the file exists

**Expected**:
Return True, if any file/directory exists

### DIFF
--- a/tensorflow_datasets/testing/test_utils.py
+++ b/tensorflow_datasets/testing/test_utils.py
@@ -152,11 +152,15 @@ class MockFs(object):
       raise FileNotFoundError('Cannot rename unknown file: {}'.format(from_))  # pytype: disable=name-error
     self.files[to] = self.files.pop(from_)
 
+  def _exists(self, path):
+    """Returns True, if any file/directory exists"""
+    return any(map(lambda f: '..' not in os.path.relpath(f, path), self.files))
+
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/testing/test_utils.py
```diff
@@ -152,11 +152,15 @@ def _rename(self, from_, to, overwrite=False):
       raise FileNotFoundError('Cannot rename unknown file: {}'.format(from_))  # pytype: disable=name-error
     self.files[to] = self.files.pop(from_)
 
+  def _exists(self, path):
+    """Returns True, if any file/directory exists"""
+    return any(map(lambda f: '..' not in os.path.relpath(f, path), self.files))
```
Conchylicultor, line 157: Checking for `..` in relpath seems quite quite hacky. Is there no better way to fix the issue ?

Conchylicultor (changes requested):

tensorflow_datasets/testing/test_utils.py
vijayphoenix: Some other alternatives I considered were
https://stackoverflow.com/questions/3328012/how-can-i-tell-if-a-file-is-a-descendant-of-a-given-directory

PR #2397, opened 2020-09-05
fix broken imoprts in bair_robot_pushing.py

rushabh-v:
While testing https://github.com/tensorflow/datasets/pull/2279 I found that the imports in `bair_robot_pushing.py` are being used improperly and that leads to attributing errors on `_generate_examples` and `_info` calls. This PR fixes those issues.

### DIFF
--- a/tensorflow_datasets/video/bair_robot_pushing.py
+++ b/tensorflow_datasets/video/bair_robot_pushing.py
@@ -24,7 +24,8 @@ import os
 
 from absl import logging
 import numpy as np
-import tensorflow.compat.v2 as tf
+import tensorflow as tf
+from tensorflow import compat
 
 import tensorflow_datasets.public_api as tfds
 
@@ -89,17 +90,17 @@ class BairRobotPushingSmall(tfds.core.GeneratorBasedBuilder):
 
   def _generate_examples(self, filedir):
     logging.info("Reading data from %s.", filedir)
-    files = tf.io.gfile.listdir(filedir)
+    files = compat.v2.io.gfile.listdir(filedir)
     logging.info("%d files found.", len(files))
 
     # For each file
-    for filename in sorted(tf.io.gfile.listdir(filedir)):
+    for filename in sorted(compat.v2.io.gfile.listdir(filedir)):
       filepath = os.path.join(filedir, filename)
 
       # For each video inside the file
       for video_id, example_str in enumerate(
-          tf.compat.v1.io.tf_record_iterator(filepath)):
-        example = tf.train.SequenceExample.FromString(example_str)
+          compat.v1.io.tf_record_iterator(filepath)):
+        example = compat.v2.train.SequenceExample.FromString(example_str)
 
         # Merge all frames together
         all_frames = []


Conchylicultor:
Could you report the error you're encountering ? I believe the import are correct and `import tensorflow.compat.v2 as tf` is valid. Which version of TF are you using ?

tensorflow_datasets/video/bair_robot_pushing.py
```diff
@@ -89,17 +90,17 @@ def _split_generators(self, dl_manager):
...
     logging.info("%d files found.", len(files))
 
     # For each file
-    for filename in sorted(tf.io.gfile.listdir(filedir)):
+    for filename in sorted(compat.v2.io.gfile.listdir(filedir)):
       filepath = os.path.join(filedir, filename)
 
       # For each video inside the file
       for video_id, example_str in enumerate(
-          tf.compat.v1.io.tf_record_iterator(filepath)):
```
rushabh-v, line 101: The problem is that `tensorflow.compat.v2` was imported as `tf`. And here `tf.compat.v1` has been called which is it points to `tensorflow.compat.v2.compat.v1` which leads to import errors.

tensorflow_datasets/video/bair_robot_pushing.py
```diff
@@ -89,17 +90,17 @@ def _split_generators(self, dl_manager):
...
-        example = tf.train.SequenceExample.FromString(example_str)
```
rushabh-v, line 102: Here it was correctly pointing to `tensorflow.compat.v2.train`.

rushabh-v: But again [here](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/video/bair_robot_pushing.py#L60-L61) `tf.float32` points to `tensorflow.compat.v2.float32`

tensorflow_datasets/video/bair_robot_pushing.py
Conchylicultor: `tensorflow.compat.v2.compat.v1` is valid too. Make sure you're using a recent version of TF.

tensorflow_datasets/video/bair_robot_pushing.py
Conchylicultor: Could you share the stacktrace ? I think your TF installation might be broken.

rushabh-v: Actually I was running it on google cloud. But it didn't allow me to use TF > 2.0.0b1 there. (maybe because I was using python3.5 ?)
Allow me some time I will try doing it again with the latest TF and report back to you!

Conchylicultor: Using a `beta` version is likely the issue. Note that TFDS do not support Python 3.5 anymore.

Conchylicultor:
Closing this as it should works with recent TFDS/TF version

tensorflow_datasets/video/bair_robot_pushing.py
rushabh-v: locally with `TF 2.2.0`. I was able to access `tf.compat.v2.compat.v1` and `tf.compat.v1.float32`. So it looks like a problem with older versions.
Closing the PR.

PR #2426, opened 2020-09-13
rename `bair_robot_pushing_small` to `bair_robot_pushing`

rushabh-v:
## Description
Previous discussion at https://github.com/tensorflow/datasets/pull/2279#issuecomment-691482621

The original dataset is having `64 x 64` frames only. So, the name bair_robot_pushing_small should be renamed as `bair_robot_pushing`


rushabh-v:
cc @martinwicke

rushabh-v:
I ran download_and_prepare on it. Not completely but it started downloading the dataset so I considered the renaming to be valid. Is that Okay?

Here is the colab where I tested it https://colab.research.google.com/drive/1tH_spnksFbzd3q7GscvRpZFxfd5x7Ecw?usp=sharing

rushabh-v:
cc @martinengelcke

rushabh-v:
could you review this PR, please? @martinengelcke

Conchylicultor:
I agree the name could have been more consistent, but is it really an issue ? Couldn't the description be updated instead to clarify the confusion ?

Updating this would break users which are already using this dataset.

Also even if the dataset used in the first paper was indeed using 64x64 images, it seems that images exists at an higher resolution, so potentially there could be another bair_robot_pushing dataset.

tensorflow_datasets/video/__init__.py
```diff
@@ -15,7 +15,7 @@
 
 """Video datasets."""
 
-from tensorflow_datasets.video.bair_robot_pushing import BairRobotPushingSmall
+from tensorflow_datasets.video.bair_robot_pushing import BairRobotPushing
```
NikhilBartwal, line 18: As @Conchylicultor pointed out, it would be best for now to only change the description so that it doesn't break the current users' code

tensorflow_datasets/video/bair_robot_pushing_test.py
```diff
@@ -20,7 +20,7 @@
 
 
 class BairRobotPushingTest(testing.DatasetBuilderTestCase):
-  DATASET_CLASS = bair_robot_pushing.BairRobotPushingSmall
```
NikhilBartwal, line 23: same as above

tensorflow_datasets/video/bair_robot_pushing.py
```diff
@@ -45,8 +45,8 @@
 """
 
 
-class BairRobotPushingSmall(tfds.core.GeneratorBasedBuilder):
-  """Robot pushing dataset from BAIR (Small 64x64 version)."""
+class BairRobotPushing(tfds.core.GeneratorBasedBuilder):
```
NikhilBartwal, line 48: This change suffices the PR as of now

rushabh-v:
Thanks for reminding me @NikhilBartwal. I almost forgot.

@Conchylicultor I was probing if there is a dataset with a higher resolution. But didn't find anything. I am not sure!

PR #2430, opened 2020-09-14
Fix a typo in the overview docs

mariosasko:
This PR fixes a typo in the [overview](https://www.tensorflow.org/datasets/overview) (section Find available datasets) docs. Instead of: ```To get the list of available builders, uses `tfds.list_builders()` ...``` , it should be: ```To get the list of available builders, use `tfds.list_builders()` ...```


Conchylicultor (approved):
Thank you for fixing this

Issue #2420, opened 2020-09-11
More consistent `checksums.tsv`


Conchylicultor:
Currently, dataset checksums are saved as `checksums.tsv` but are not Tab-Separated-Values (but Space-Separated-Value).  This feel inconsistent.

We should update:
https://github.com/tensorflow/datasets/blob/3e90c815a61db5236063817cf699f2bd4d7eef58/tensorflow_datasets/core/download/checksums.py#L154
to save and load using tab instead of space.

* Current checksums should be updated to use tab.
* We should keep a backward compatibility mode with space while files gets updated.

Also I noticed
* `url_infos_from_path` and `_get_url_infos` seems redundant as they both do the same thing. This could be cleaned up

NikhilBartwal:
@Conchylicultor I would like to take this issue.
As far as I understood we need to use `tab` instead of `space` in the following :

For retrieving:
https://github.com/tensorflow/datasets/blob/3e90c815a61db5236063817cf699f2bd4d7eef58/tensorflow_datasets/core/download/checksums.py#L154
 
For storing:
https://github.com/tensorflow/datasets/blob/3e90c815a61db5236063817cf699f2bd4d7eef58/tensorflow_datasets/core/download/checksums.py#L204

However, I don't quite get how to keep it backward compatible with space?

Conchylicultor:
Thanks for looking into this. To keep it backward compatible, you can first try to load it as tab, then fallback to space if this fail

NikhilBartwal:
@Conchylicultor I have fixed the `checksums `script and aded an `update_checksums `script for updating all checksum files (both .txt and .tsv).
Could you have a look at the PR? Thanks!

vijayphoenix:
The issue seems to be fixed

PR #2433, opened 2020-09-15
Add core tab support for checksums

NikhilBartwal:
PR in reference to Issue #2420
`core/download/checksums.py' has been fixed to now use `tabs` for storage and retrieval of checksums

### DIFF
--- a/tensorflow_datasets/core/download/checksums.py
+++ b/tensorflow_datasets/core/download/checksums.py
@@ -151,14 +151,9 @@ def _parse_url_infos(checksums_file: Iterable[str]) -> Dict[str, UrlInfo]:
     if not line or line.startswith('#'):
       continue
     # URL might have spaces inside, but size and checksum will not.
-    url, size, checksum = line.rsplit(' ', 2)
+    url, size, checksum = line.rsplit('\t', 2)
     url_infos[url] = UrlInfo(size=int(size), checksum=checksum)
   return url_infos
-
-
-def url_infos_from_path(checksums_path: str) -> Dict[str, UrlInfo]:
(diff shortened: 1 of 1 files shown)


Conchylicultor (changes requested):
Thank you, but this will fail for legacy checksums files.
Could you make sure this is backward compatible with old checksum files which still uses space ?

NikhilBartwal:
@Conchylicultor Sorry for that, that part had been left out by mistake while creating different branches :(

Conchylicultor (approved):
Thank you!

tensorflow_datasets/core/download/checksums.py
```diff
@@ -151,16 +151,14 @@ def _parse_url_infos(checksums_file: Iterable[str]) -> Dict[str, UrlInfo]:
...
-    url, size, checksum = line.rsplit(' ', 2)
+    try:
+      url, size, checksum = line.rsplit('\t', 2)
+    except ValueError:
+      url, size, checksum = line.rsplit(' ', 2)
     url_infos[url] = UrlInfo(size=int(size), checksum=checksum)
   return url_infos
 
 
-def url_infos_from_path(checksums_path: str) -> Dict[str, UrlInfo]:
```
Conchylicultor, line 159: Note that this function is called in other part of the code so this is breaking TFDS. I've removed this changed internally.

NikhilBartwal: @Conchylicultor Do i need to change the PR or is it alright?

Conchylicultor: No need to update this PR

PR #2434, opened 2020-09-15
Add update_checksums.py script

NikhilBartwal:
PR in reference to Issue #2420
`update_checksums.py` script is added for updating all type of checksum files

### DIFF
--- a/tensorflow_datasets/scripts/update_checksums.py
+++ b/tensorflow_datasets/scripts/update_checksums.py
@@ -1,0 +1,108 @@
+# coding=utf-8
+# Copyright 2020 The TensorFlow Datasets Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Script to update all the checksums.tsv files with tabs instead of spaces"""
+
+import os
+from absl import app
+from typing import List
+
+import tensorflow.compat.v2 as tf
+
+_DATASET_TYPES = list(['audio',
+                        'image',
+                        'image_claffication',
+                        'onject_detection',
+                        'question_answering',
+                        'structured',
+                        'summarization',
+                        'text',
+                        'translate',
+                        'video'])
+
+_ROOT_DIR = os.path.normpath(os.path.dirname(os.path.dirname(__file__)))
+
+_TXT_CHECKSUM_DIR = os.path.join(_ROOT_DIR, 'url_checksums')
+_TXT_CHECKSUM_SUFFIX = '.txt'
+_TSV_CHECKSUM_SUFFIX = '.tsv'
+
+
+def get_tsv_checksums_paths() -> List[str]:
+  """Returns the paths of all the .tsv checksum files of folder datasets"""
+  paths = list()
+  for dataset_type in _DATASET_TYPES:
+    #Check whether the dataset type folder exists
+    dataset_dir = os.path.join(_ROOT_DIR, dataset_type)
+    if tf.io.gfile.isdir(dataset_dir):
+
+      #Find the paths of all datasets in every type of dataset
+      dataset_paths = [os.path.join(dataset_dir, dir)
+                  for dir in tf.io.gfile.listdir(dataset_dir)
+                  if tf.io.gfile.isdir(os.path.join(dataset_dir, dir)) and
+                  dir!='_pycache_']
+
+      #Getting the path of eack checksum file for every dataset
+      for dataset_path in dataset_paths:
+        checksum_path = os.path.join(dataset_path, 'checksums.tsv')
+        if os.path.isfile(checksum_path):
+          paths.append(checksum_path)
+    else:
+      continue
+  return paths
+
+def get_txt_checksums_paths() -> List[str]:
+  """Returns the paths of all the .txt checksum files relative to the
+  root directory"""
+
+  #print(_TXT_CHECKSUM_DIR)
+  paths = list([os.path.join(_TXT_CHECKSUM_DIR, path)
+                for path in tf.io.gfile.listdir(_TXT_CHECKSUM_DIR)
+                if path.endswith(_TXT_CHECKSUM_SUFFIX)])
+  return paths
+
+
+def rewrite_checksums(paths : List[str]) -> None:
+  """Update all .txt checksum files to use tabs instead of spaces"""
+  for path in paths:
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -0,0 +1,108 @@
...
+import os
+from absl import app
+from typing import List
+
+import tensorflow.compat.v2 as tf
+
+_DATASET_TYPES = list(['audio',
+                        'image',
+                        'image_claffication',
+                        'onject_detection',
```
Conchylicultor, line 27: object_detection

tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -0,0 +1,108 @@
...
+                        'question_answering',
+                        'structured',
+                        'summarization',
+                        'text',
+                        'translate',
+                        'video'])
+
+_ROOT_DIR = os.path.normpath(os.path.dirname(os.path.dirname(__file__)))
+
+_TXT_CHECKSUM_DIR = os.path.join(_ROOT_DIR, 'url_checksums')
```
Conchylicultor, line 37: Rather than hardcoding path here. Is it not possible to reuse the constant defined in `checksums.py` ?

tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -0,0 +1,108 @@
...
+    if content == []:
+      continue
+    elif len(content[0].split('\t'))==3:
+      continue
+    else:
+      content = [info.rsplit(' ', 2) for info in content]
+      with tf.io.gfile.GFile(path, 'w') as f:
+        for info in content:
+          url, size, checksum = info
+          f.write('{}    {}    {}\n'.format(url, size, checksum))
```
Conchylicultor, line 94: 1 tab is not the same character as 4 spaces.

tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -0,0 +1,108 @@
...
+  paths = list([os.path.join(_TXT_CHECKSUM_DIR, path)
+                for path in tf.io.gfile.listdir(_TXT_CHECKSUM_DIR)
+                if path.endswith(_TXT_CHECKSUM_SUFFIX)])
+  return paths
+
+
+def rewrite_checksums(paths : List[str]) -> None:
+  """Update all .txt checksum files to use tabs instead of spaces"""
+  for path in paths:
+    print('Processing file: {}'.format(path))
```
Conchylicultor, line 79: Use f-string rather than .format

tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -0,0 +1,108 @@
...
+
+_TXT_CHECKSUM_DIR = os.path.join(_ROOT_DIR, 'url_checksums')
+_TXT_CHECKSUM_SUFFIX = '.txt'
+_TSV_CHECKSUM_SUFFIX = '.tsv'
+
+
+def get_tsv_checksums_paths() -> List[str]:
+  """Returns the paths of all the .tsv checksum files of folder datasets"""
+  paths = list()
+  for dataset_type in _DATASET_TYPES:
```
Conchylicultor, line 45: To list datasets, how about using `tfds.list_builder` with `tfds.builder_cls(name)`, and `builder.code_path.parent / 'checksums.tsv'`

Conchylicultor (changes requested):

tensorflow_datasets/scripts/update_checksums.py
NikhilBartwal: @Conchylicultor The `_ROOT_DIR` used in `checksums.py` uses the following code
`_ROOT_DIR = os.path.normpath(os.path.join(os.path.dirname(__file__), '../..'))`
and thus, gets the absolute path upto the `../tensorflow_datasets` directory as the file is located in `../tensorflow_datasets/core/download`
But the `update_checksums` script is located in the `../tensorflow_datasets/scripts` and thus, has one level less in its path.
Therefore, I don't think we can use the same code for both of them unless we put the `update_checksums` script inside the `core` directory or the `scripts/cleanup` directory to make the path levels equal

tensorflow_datasets/scripts/update_checksums.py
NikhilBartwal: @Conchylicultor I tried using `\t` as it should it, but it resulted in unequal spaces when the script was run. Should I use `\t` nonetheless?

tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -0,0 +1,97 @@
...
+                        'question_answering',
+                        'structured',
+                        'summarization',
+                        'text',
+                        'translate',
+                        'video'])
+
+_ROOT_DIR = os.path.normpath(os.path.dirname(os.path.dirname(__file__)))
+
+_TXT_CHECKSUM_DIR = os.path.join(_ROOT_DIR, 'url_checksums')
```
acharles7, line 38: @NikhilBartwal We are using the `Pathlib` module for path manipulation. It would be great if we use the `Pathlib` module instead of `os` to make this script consistent with other scripts.

tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -33,9 +33,9 @@
                         'translate',
                         'video'])
 
-_ROOT_DIR = os.path.normpath(os.path.dirname(os.path.dirname(__file__)))
+_ROOT_DIR = Path(__file__).parent.parent
 
-_TXT_CHECKSUM_DIR = os.path.join(_ROOT_DIR, 'url_checksums')
+_TXT_CHECKSUM_DIR = Path.joinpath(_ROOT_DIR, 'url_checksums')
```
acharles7, line 38: `_TXT_CHECKSUM_DIR = _ROOT_DIR / 'url_checksums'`

tensorflow_datasets/scripts/update_checksums.py
```diff
@@ -0,0 +1,97 @@
...
+
+def get_txt_checksums_paths() -> List[str]:
+  """Returns the paths of all the .txt checksum files relative to the
+  root directory"""
+
+  #print(_TXT_CHECKSUM_DIR)
+  paths = list([Path.joinpath(_TXT_CHECKSUM_DIR, path)
+                for path in tf.io.gfile.listdir(_TXT_CHECKSUM_DIR)
+                if path.endswith(_TXT_CHECKSUM_SUFFIX)])
+  return paths
```
acharles7, line 62: why list()? you can simply do
```python
  paths = [_TXT_CHECKSUM_DIR / path)
            for path in tf.io.gfile.listdir(_TXT_CHECKSUM_DIR)
            if path.endswith(_TXT_CHECKSUM_SUFFIX)]
  return paths

```

NikhilBartwal: @acharles7 That was a mistake as the code was written for the os module :p

NikhilBartwal:
@Conchylicultor Could you have a look at the changes? Thanks!

Conchylicultor (approved):
Thank you. Looks good.

vijayphoenix:
Closing as the files have been updated
Fixed #2420

Issue #2423, opened 2020-09-11
Extraction bug for Github urls


Conchylicultor:
`dl_manager.download_and_extract` do not auto-detect the archive extension for github url (e.g. https://github.com/brendenlake/SCAN/archive/master.zip).

Currently, dataset creators are require to hardcode the extension format: https://github.com/tensorflow/datasets/blob/b9deaa07f567ad3211850a1e5c1ae993a1a03409/tensorflow_datasets/text/scan.py#L160

Fails (`data_dir` is downloaded but not extracted):
```py
data_dir = dl_manager.download_and_extract('https://github.com/brendenlake/SCAN/archive/master.zip')
```

Work:
```py
data_dir = dl_manager.download_and_extract(
        tfds.download.Resource(
            url='https://github.com/brendenlake/SCAN/archive/master.zip',
            extract_method=tfds.download.ExtractMethod.ZIP,
       ),
)
```

vijayphoenix:
This bug occurs because we fail to pass the filename with extension.

We get the following response headers when we make a `get` request to the GitHub URLs. 
(In this example `https://github.com/mogutt/README/archive/master.zip`)
```
{
   "headers":{
      "Content-Disposition":"attachment; filename=README-master.zip",
      ...
   },
   "url":"https://codeload.github.com/mogutt/README/zip/master",
   ....
}
```

So, when we pass the above response to the `_get_filaname` function:

https://github.com/tensorflow/datasets/blob/bd47f95c35096232e7ace131ae725c4ebdc7b3d3/tensorflow_datasets/core/download/downloader.py#L47-L55

We get `master` as a return value.

SuryashankarDas:
Right now if the the content-disposition is not enclosed within quotes, we cant get a match. How about we do something like, if the match is null after the first case, we check for a match without quotes? It worked out when I tried.

Conchylicultor:
Thank you for fixing this bug.

As follow up, we should cleanup the `tfds.download.Resource` as in https://github.com/tensorflow/datasets/blob/b9deaa07f567ad3211850a1e5c1ae993a1a03409/tensorflow_datasets/text/scan.py#L160,
and check the download are still working.

PR #2438, opened 2020-09-15
update get_filename to match for non-quoted strings

SuryashankarDas:
This references issue #2423 [bug]

### DIFF
--- a/tensorflow_datasets/core/download/downloader.py
+++ b/tensorflow_datasets/core/download/downloader.py
@@ -47,7 +47,7 @@ def get_downloader(*args: Any, **kwargs: Any) -> '_Downloader':
 def _get_filename(response: Response) -> str:
   content_disposition = response.headers.get('content-disposition', None)
   if content_disposition:
-    match = re.findall('filename="(.+?)"', content_disposition)
+    match = re.findall('[^=;\'"%]+\.[^=; \'"]+', content_disposition)
     if match:
       return match[0]
   return utils.basename_from_url(response.url)
--- a/tensorflow_datasets/core/download/downloader_test.py
+++ b/tensorflow_datasets/core/download/downloader_test.py
@@ -146,13 +146,17 @@ class GetFilenameTest(testing.TestCase):
     self.assertEqual(res, 'baz.zip')
 
   def test_headers(self):
-    cdisp = ('attachment;filename="hello.zip";'
-             'filename*=UTF-8\'\'hello.zip')
-    resp = _FakeResponse('http://foo.bar/baz.zip', b'content', headers={
-        'content-disposition': cdisp,
-    })
-    res = downloader._get_filename(resp)
-    self.assertEqual(res, 'hello.zip')
+    cdisps = [['attachment;filename*=UTF-8\'\'hello.zip','hello.zip'],
+             ['Attachment; filename=example.html','example.html'],
+             ['INLINE; FILENAME= "an example.html"','an example.html'],
+             ["attachment;filename*= UTF-8''%e2%82%ac%20rates.zip",'20rates.zip'],
+             ["attachment;filename='EURO rates';filename*=utf-8''%e2%82%ac%20rates",'baz.zip']]
+    for (cdisp,value) in cdisps:
+      resp = _FakeResponse('http://foo.bar/baz.zip', b'content', headers={
+          'content-disposition': cdisp,
+      })
+      res = downloader._get_filename(resp)
+      self.assertEqual(res, value)
 
 
 if __name__ == '__main__':


tensorflow_datasets/core/download/downloader.py
```diff
@@ -48,6 +48,8 @@ def _get_filename(response: Response) -> str:
   content_disposition = response.headers.get('content-disposition', None)
   if content_disposition:
     match = re.findall('filename="(.+?)"', content_disposition)
+    if (not match):
```
vijayphoenix, line 51: Unnecessary brackets.

vijayphoenix (changes requested):
Thanks for your contribution.
Also, please add some tests in `downloader_test.py`.

tensorflow_datasets/core/download/downloader.py
```diff
@@ -48,6 +48,8 @@ def _get_filename(response: Response) -> str:
   content_disposition = response.headers.get('content-disposition', None)
   if content_disposition:
     match = re.findall('filename="(.+?)"', content_disposition)
+    if not match:
+        match = re.findall('filename=(.+)', content_disposition)
```
Conchylicultor, line 52: How about only using `filename=(.+?)`, then returning `match[0].strip('"')` ?

Conchylicultor (changes requested):
Thank you!

tensorflow_datasets/core/download/downloader.py
SuryashankarDas: `filename=(.+?)` does not work at all, `filename="(.+?)"` works only with quotes, `filename=(.+) and strip will now work for certain test cases ('attachment;filename="hello.zip";''filename*=UTF-8\'\'hello.zip')

SuryashankarDas: So in short we need to check both cased. As for adding tests in `download_test.py` should I just edit the current test and add a new csid without quotes (and loop through the csids to assert if it is working)?

Conchylicultor: For testing, I would add a parametrised test which check the various examples from: https://tools.ietf.org/html/rfc6266#section-5

SuryashankarDas:
Earlier the filename present in` filename*` in the content description was being completely ignored. Now it will effectively check it too.

Conchylicultor:
Thank you for the update. Could you add tests which check the various examples from: https://tools.ietf.org/html/rfc6266#section-5

SuryashankarDas:
I have never written any unit tests before, so is this fine or was I supposed to use some python module like `parameterized` or `pytest`? Btw check the csid value pairs in the latest commit to check if its alright.

vijayphoenix:
You can refer to the following example https://github.com/tensorflow/datasets/blob/10f49a9d0ce0008f617bc3f0b1e5e661fa2ef5cc/tensorflow_datasets/core/read_only_builder_test.py#L39

Conchylicultor (commented):
Thank you for adding unittest. This looks good.

Have to tried to run `pytest tensorflow_datasets/core/download/downloader_test.py` to make sure the tests are passing ?

SuryashankarDas:
Yup its working
```
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-3.6.4, py-1.9.0, pluggy-0.7.1
rootdir: /content/datasets, inifile: pytest.ini
plugins: typeguard-2.7.1
collected 11 items                                                             

datasets/tensorflow_datasets/core/download/downloader_test.py .......s.. [ 90%]
s                                                                        [100%]

===================== 9 passed, 2 skipped in 0.27 seconds ======================
```

tensorflow_datasets/core/download/downloader.py
```diff
@@ -47,7 +47,7 @@ def get_downloader(*args: Any, **kwargs: Any) -> '_Downloader':
 def _get_filename(response: Response) -> str:
   content_disposition = response.headers.get('content-disposition', None)
   if content_disposition:
-    match = re.findall('filename="(.+?)"', content_disposition)
+    match = re.findall('[^=;\'"%]+\.[^=; \'"]+', content_disposition)
```
Conchylicultor, line 50: There is no more `filename`. Is it expected ?

SuryashankarDas: Yup. We are searching for files with any extentions (it needs to have a dot) and searching by `filename=` ignores `filename*=`.

tensorflow_datasets/core/download/downloader_test.py
```diff
@@ -146,13 +146,17 @@ def test_no_headers(self):
...
-             'filename*=UTF-8\'\'hello.zip')
-    resp = _FakeResponse('http://foo.bar/baz.zip', b'content', headers={
-        'content-disposition': cdisp,
-    })
-    res = downloader._get_filename(resp)
-    self.assertEqual(res, 'hello.zip')
+    cdisps = [['attachment;filename*=UTF-8\'\'hello.zip','hello.zip'],
+             ['Attachment; filename=example.html','example.html'],
+             ['INLINE; FILENAME= "an example.html"','an example.html'],
+             ["attachment;filename*= UTF-8''%e2%82%ac%20rates.zip",'20rates.zip'],
```
Conchylicultor, line 152: The filename is not properly decoded right ? The resulting filename should be `€ rates.zip`. This could be fixed in a future PR if you prefer

Conchylicultor (changes requested):

tensorflow_datasets/core/download/downloader_test.py
SuryashankarDas: Do we need to decode it? The original bug was related to the extention not being detected causing the extraction to fail.

Conchylicultor: Which is why this could be fixed in a separate issue, but this is still an issue nonetheless.
I think it would be best to raise an error in this case, or not infer the filename, rather than silently recording a wrong filename.

SuryashankarDas: Oh got it. So a separate case for `filename*=` to decode it.

Conchylicultor: I've imported the code and made some changes to return None when only the encoded filename is provided.
We might want to fix the filename*= case later if we encounter a case where we need it.

Anyway, this will be merged soon. Thank you for your contribution

SuryashankarDas: I got the encoded filename part done, so should i commit to this PR or should i do it in a new PR (after this is merged)?

Conchylicultor: Please open a new PR after this one has been merged, as I've already imported merged this one. It should be merged on github soon.

PR #2455, opened 2020-09-19
Support reading config for partial versioning

### LINKED ISSUE #2446: Support reading config for partial versioning 'my_dataset:3.*.*'
Currently, the dynamic finding dataset does not work when the version is partial:

```py
load.find_builder_dir('my_dataset:3.*.*')  # Currently returns None
```

Instead, we should update the `_get_last_version(builder_dir)` as: `version_str = _get_version_str(builder_dir, version_str)` to search match the `version_str` across the existing versions:

```py
/my_dataset/1.0.0
/my_dataset/1.0.1
/my_dataset/1.1.0
/my_dataset/2.0.1

_get_version_str('/my_dataset/', None) -> 2.0.1   # Last version as previously
_get_version_str('/my_dataset/', '1.*.*') -> 1.1.0  # Last compatible version
_get_version_str('/my_dataset/', '*.*.*') -> 2.0.1  # Last compatible version
_get_version_str('/my_dataset/', '1.3.*') -> None  # No compatible version found
_get_version_str('/my_dataset/', '2.3.5') -> None  # Not found
```

vijayphoenix:
Fix #2446

### DIFF
--- a/tensorflow_datasets/core/load.py
+++ b/tensorflow_datasets/core/load.py
@@ -382,7 +382,7 @@ def find_builder_dir(
 
   # If version not given, extract the version
   if not version_str:
-    version_str = _get_last_version(builder_dir)
+    version_str = _get_version_str(builder_dir, requested_version=version_str)
 
   if not version_str:  # No version given nor found
     return None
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/load.py
```diff
@@ -382,7 +382,7 @@ def find_builder_dir(
 
   # If version not given, extract the version
   if not version_str:
```
Conchylicultor, line 384: `if not version_str` means that _get_version_str will never be called with `version_str is not None`. Haven't the tests catched this error ?

Conchylicultor (changes requested):
Thank you Vijay

tensorflow_datasets/core/load.py
vijayphoenix: Added some tests.
It failed only was `version_str` was a regex.

Conchylicultor (approved):
Thank you Vijay

Conchylicultor (approved):
Nice!

Issue #2460, opened 2020-09-21
CleanUp Dataset Code


vijayphoenix:
**Short description**
Recently, the bug #2423 related to the auto-detection of filename and extension was fixed.

So, we can remove some unnecessary code from the datasets given below.

Steps:
* Remove `tfds.download.resource` from the datasets given below.
* Download and prepare the datasets.
* Add the link to generated `dataset_info.json`

For reference, see https://github.com/tensorflow/datasets/commit/9a78f9eba3ce2ca2d69083b1de33e98722e08ef7

* [x] audio/tedlium
* [x] audio/spoken_digit/spoken_digit
* [x] image/arc
* [x] image/bccd/
* [x] image_classification/cbis_ddsm
* [x] image_classification/deep_weeds
* [x] image_classification/dtd
* [x] image_classification/geirhos_conflict_stimuli
* [ ] image_classification/inaturalist
* [x] image_classification/oxford_flowers102
* [x] image_classification/oxford_iiit_pet
* [ ] image_classification/sun
* [x] image_classification/visual_domain_decathlon
* [ ] object_detection/open_images
* [x] summarization/aeslc
* [x] text/gpt3
* [x] text/irc_disentanglement
* [x] text/scan

SuryashankarDas:
If no one is already working on it, I would like to work on this. btw, should i squash it all in a single commit or do it one by one?

Keshav15:
I would Also  like to work on this issue

vijayphoenix:
Please fix only 3-5 datasets per pull request.
Feel free to distribute the work among yourself.

Ak0303:
I would also like to work on this !!

vijayphoenix:
This issue seems to be fixed.

PR #2478, opened 2020-09-22
Cleanup code for Visual Domain Decathlon and other

NikhilBartwal:
* Dataset Name: Visual Domain Decathlon, Geirhos Conflict Stimuli
* Issue Reference: #2460 
* `Visual Domain Decathlon dataset_info.json` Gist: [here](https://gist.github.com/NikhilBartwal/0eef5b39a9fab4edd17536eb6b2b49c8)
* `Geirhos Conflict Stimuli dataset_info.json` Gist: [here](https://gist.github.com/NikhilBartwal/8b9fdf73b4b88c6c241d20772aba1850)

### DIFF
--- a/tensorflow_datasets/image_classification/visual_domain_decathlon.py
+++ b/tensorflow_datasets/image_classification/visual_domain_decathlon.py
@@ -107,9 +107,7 @@ class VisualDomainDecathlon(tfds.core.GeneratorBasedBuilder):
     if self.builder_config.name == 'imagenet12':
       devkit_path, images_archive = dl_manager.download_and_extract([
           _URL_PREFIX_VGG + 'decathlon-1.0-devkit.tar.gz',
-          tfds.download.Resource(
-              url=_URL_PREFIX_IMAGENET + 'decathlon-1.0-data-imagenet.tar',
-              extract_method=tfds.download.ExtractMethod.NO_EXTRACT),
+          _URL_PREFIX_IMAGENET + 'decathlon-1.0-data-imagenet.tar',
       ])
     else:
       devkit_path, data_path = dl_manager.download_and_extract([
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/image_classification/visual_domain_decathlon.py
```diff
@@ -107,9 +107,7 @@ def _split_generators(self, dl_manager):
     if self.builder_config.name == 'imagenet12':
       devkit_path, images_archive = dl_manager.download_and_extract([
           _URL_PREFIX_VGG + 'decathlon-1.0-devkit.tar.gz',
-          tfds.download.Resource(
-              url=_URL_PREFIX_IMAGENET + 'decathlon-1.0-data-imagenet.tar',
-              extract_method=tfds.download.ExtractMethod.NO_EXTRACT),
```
Conchylicultor, line 112: Removing `NO_EXTRACT` looks like a bug to me. Now `images_archive` will linked to the extracted dir, rather than the archive. I'm quite surprised this is working. Could you investigate why this is not crashing ? Are you sure that you're generating using the modified code ?

Conchylicultor (changes requested):

tensorflow_datasets/image_classification/visual_domain_decathlon.py
NikhilBartwal: @Conchylicultor That indeed, looks like a bug. I generated the default dataset during testing using `tfds.load('visual_domain_decathlon')` due to which it executes the else block which puts the path in `image_archive` manually, therefore, preventing the bug from happening.
https://github.com/tensorflow/datasets/blob/c7439f84c2a77891e07a8e154cbde21f6ddcc8e9/tensorflow_datasets/image_classification/visual_domain_decathlon.py#L119
Currently, I do not have the bandwidth required to try downloading the complete `imagenet12` dataset.
However, we could use `tfds.download` instead of `tfds.download_and_extract` in case of `imagenet12` builder config, thus allowing to follow the given program flow

Conchylicultor: I think keeping `NO_EXTRACT` in this case if fine, as this allow to download urls in parallel, while a separate `dl_manager.download` will download sequentially.
Thank you for confirming. I'll only merge the other dataset. So nothing to do from your end.

PR #2479, opened 2020-09-22
Cleanup Code for Open Images and others

NikhilBartwal:
* Dataset Name: Open Images, Sun, iNaturalist
* Issue Reference: #2460 
Datasets are generating properly!

### DIFF
--- a/tensorflow_datasets/image_classification/inaturalist.py
+++ b/tensorflow_datasets/image_classification/inaturalist.py
@@ -78,16 +78,11 @@ class INaturalist2017(tfds.core.GeneratorBasedBuilder):
   def _split_generators(self, dl_manager):
     output_files = dl_manager.download_and_extract({
         "trainval_images":
-            tfds.download.Resource(
-                url=urllib.parse.urljoin(_URL, "train_val_images.tar.gz"),
-                extract_method=tfds.download.ExtractMethod.NO_EXTRACT),
+            urllib.parse.urljoin(_URL, "train_val_images.tar.gz"),
         "trainval_annos":
             urllib.parse.urljoin(_URL, "train_val2017.zip"),
         "test_images":
-            tfds.download.Resource(
-                url=urllib.parse.urljoin(_URL, "test2017.tar.gz"),
-                extract_method=tfds.download.ExtractMethod.NO_EXTRACT),
-
+            urllib.parse.urljoin(_URL, "test2017.tar.gz"),
     })
     return [
         tfds.core.SplitGenerator(
--- a/tensorflow_datasets/image_classification/sun.py
+++ b/tensorflow_datasets/image_classification/sun.py
@@ -34,7 +34,7 @@ year={2010},
 volume={},
 number={},
 pages={3485-3492},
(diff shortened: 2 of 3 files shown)


tensorflow_datasets/image_classification/inaturalist.py
```diff
@@ -78,16 +78,11 @@ def _info(self):
   def _split_generators(self, dl_manager):
     output_files = dl_manager.download_and_extract({
         "trainval_images":
-            tfds.download.Resource(
-                url=urllib.parse.urljoin(_URL, "train_val_images.tar.gz"),
-                extract_method=tfds.download.ExtractMethod.NO_EXTRACT),
```
Conchylicultor, line 83: Are you sure removing `tfds.download.Resource` is working ? I would expect that the `.tar.gz` is no extracted, so `dl_manager.iter_archive` would crash.

Conchylicultor (changes requested):

tensorflow_datasets/image_classification/inaturalist.py
NikhilBartwal: @Conchylicultor I did notice that later on. However, the dataset gives a `download error` using the modified code as well as the original code as follows:

`tensorflow_datasets.core.download.downloader.DownloadError: Failed to get url http://people.vision.caltech.edu/~gvanhorn/datasets/inaturalist/fgvc4_competition/test2017.tar.gz. HTTP code: 403.`

Trying to download the dataset manually using the link in the code gives a permission error

`You don't have permission to access /~gvanhorn/datasets/inaturalist/fgvc4_competition/train_val_images.tar.gz on this server.`

I think we would need to find some other alternate place to get the dataset from

Conchylicultor: It means the website has been updated. And the dataset should be updated (but in another PR). We should not merge changes which are not working

Keshav15: @Conchylicultor  when we try to visit 
http://people.vision.caltech.edu/~gvanhorn/datasets/inaturalist/fgvc4_competition/
then we have 3 files there but not (train_val_images.tar.gz)
so exactly what should be done?

NikhilBartwal: @Conchylicultor Yes, I think we should leave the `inaturalist` dataset and merge the other two

Conchylicultor: It's depends, are those datasets working  with those changes ?

Conchylicultor (changes requested):

NikhilBartwal:
@Conchylicultor The changes in the Open images dataset are compatible with the original code whereas, the `paths` variable in the Sun dataset was best used using the original `tfds.download.resource` method, which has been fixed now. I thinkl the `Open Images` and the `Sun` dataset PRs can be merged now. The `inaturalist` is best with the original code until we find another website to get the dataset from

vijayphoenix:
Closing the PR for now https://github.com/tensorflow/datasets/pull/2479#discussion_r498882091

PR #2502, opened 2020-09-30
CBIS-DDSM: fix random patch cropping minimum position (min_x, min_y)

lazatsoc:
## Description

There was an issue in the calculation of the minimum position (x,y) that a patch can be sampled from, causing all the abnormalities to be located in the upper-left corner of the patch. It also lead to the creation of several patches with size smaller than the indicated (224,224).
  
## Checklist

@jpuigcerver

### DIFF
--- a/tensorflow_datasets/image_classification/cbis_ddsm.py
+++ b/tensorflow_datasets/image_classification/cbis_ddsm.py
@@ -709,8 +709,8 @@ def _sample_positive_patches(image,
     # Determine the region where random samples should be sampled from.
     max_h, min_h = max(abnorm_h, patch_size[0]), min(abnorm_h, patch_size[0])
     max_w, min_w = max(abnorm_w, patch_size[1]), min(abnorm_w, patch_size[1])
-    min_y = abnorm_y - int(max_h - min_h) - int((1.0 - min_overlap_threshold) * min_h)
-    min_x = abnorm_x - int(max_w - min_w) - int((1.0 - min_overlap_threshold) * min_w)
+    min_y = abnorm_y - max_h + min_overlap_threshold * min_h
+    min_x = abnorm_x - max_w + min_overlap_threshold * min_w
     max_y = abnorm_y + abnorm_h - int(min_overlap_threshold * min_h)
     max_x = abnorm_x + abnorm_w - int(min_overlap_threshold * min_w)
     # Ensure that all sampled batches are within the image.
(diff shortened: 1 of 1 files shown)


jpuigcerver:
Thanks for your contribution!

Do you have some examples of the patches generated before and after this patch, for a given image?

tensorflow_datasets/image_classification/cbis_ddsm.py
```diff
@@ -709,8 +709,8 @@ def _sample_positive_patches(image,
     # Determine the region where random samples should be sampled from.
     max_h, min_h = max(abnorm_h, patch_size[0]), min(abnorm_h, patch_size[0])
     max_w, min_w = max(abnorm_w, patch_size[1]), min(abnorm_w, patch_size[1])
-    min_y = abnorm_y - int((1.0 - min_overlap_threshold) * max_h)
-    min_x = abnorm_x - int((1.0 - min_overlap_threshold) * max_w)
+    min_y = abnorm_y - int(max_h - min_h) - int((1.0 - min_overlap_threshold) * min_h)
```
Conchylicultor, line 712: Could this be simplified into `- max_h + min_overlap_threshold * min_h` ?

Conchylicultor (changes requested):
From internal discussion, 
> The proposed change doesn't make much sense to me. IIUC, the new min_y/min_x are situated even closer to the upper-left corner, thus I don't see how this would help to not getting patches close to that region.

Could you give more context about what is this change solving ?

tensorflow_datasets/image_classification/cbis_ddsm.py
lazatsoc: OK so the simplification is of course correct, sorry about that.

Regarding the comment from the internal discussion, I will try to clarify:
In the initial implementation, min_overlap_threshold = 0.9 and abnorm_x and abnorm_y are the coordinates of the upper left corner of the abnormality bounding box. So the **minimum window position** is min_x = abnorm_x - 22.4. This is too close to the abnormality x: All possible random windows will start at a maximum 22 pixel distance from the ibnormality.
Instead, the actual minimum window x and y should indeed come closer the the upper-left corner of the mammogram which will place the abnormality further down-right in the patch..

Conchylicultor: @jpuigcerver FYI

Conchylicultor:
This was merged by https://github.com/tensorflow/datasets/commits?author=lazatsoc

takifouhal:
finally ! , thank you.

PR #2516, opened 2020-10-02
some fixes on class numbers

serereuk:
# Add Dataset

* Dataset Name: plant_village
* Issue Reference: https://data.mendeley.com/datasets/tywbtsjrjv/1

  
## Description

I fix a typo ( Number of classes should be 39 not 38) after reading that reference site.

```
In this data-set, 39 different classes of plant leaf and background images are available.  
```


serereuk:
This one was mistake. I didn't read ReadMe carefully... 38 was right :)

Issue #2252, opened 2020-07-31
SSL CA failures when using a dataset from gs://


Flamefire:
**Short description**

When trying to load a dataset I get an error "Problem with the SSL CA cert (path? access rights?)" and a subsequent error when `six.reraise` is called

**Environment information**
* Operating System: RHEL 7
* Python version: 3.7.4
* `tensorflow-datasets` version: 3.2.1
* `tensorflow` version: 2.2.0

**Reproduction instructions**

- Run in python `tfds.builder("imagenet2012").info`

**Link to logs**

```
Traceback (most recent call last):
  File "/home/s3248973/.local/lib/python3.7/site-packages/tensorflow_datasets/core/utils/py_utils.py", line 399, in try_reraise
    yield
  File "/home/s3248973/.local/lib/python3.7/site-packages/tensorflow_datasets/core/registered.py", line 244, in builder
    return builder_cls(name)(**builder_kwargs)
  File "/home/s3248973/.local/lib/python3.7/site-packages/tensorflow_datasets/core/api_utils.py", line 69, in disallow_positional_args_dec
    return fn(*args, **kwargs)
  File "/home/s3248973/.local/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_builder.py", line 206, in __init__
    self.info.initialize_from_bucket()
  File "/home/s3248973/.local/lib/python3.7/site-packages/tensorflow_datasets/core/dataset_info.py", line 423, in initialize_from_bucket
    data_files = gcs_utils.gcs_dataset_info_files(self.full_name)
  File "/home/s3248973/.local/lib/python3.7/site-packages/tensorflow_datasets/core/utils/gcs_utils.py", line 71, in gcs_dataset_info_files
    return gcs_listdir(posixpath.join(GCS_DATASET_INFO_DIR, dataset_dir))
  File "/home/s3248973/.local/lib/python3.7/site-packages/tensorflow_datasets/core/utils/gcs_utils.py", line 64, in gcs_listdir
    if is_gcs_disabled() or not tf.io.gfile.exists(root_dir):
  File "/sw/installed/TensorFlow/2.1.0-fosscuda-2019b-Python-3.7.4/lib/python3.7/site-packages/tensorflow_core/python/lib/io/file_io.py", line 280, in file_exists_v2
    pywrap_tensorflow.FileExists(compat.as_bytes(path))
[...]

PR #2547, opened 2020-10-08
Remove dependency on dm-tree 

Flamefire:
As discussed in https://github.com/tensorflow/datasets/issues/2252#issuecomment-692589592 using dm-tree is problematic due to its usage of Bazel to build itself on architectures where it is not available as a binary package.

As only 2 functions are used which are very similar and pretty easy to implement this is done here, tests added and used by the 2 other functions.

### DIFF
--- a/tensorflow_datasets/core/as_dataframe.py
+++ b/tensorflow_datasets/core/as_dataframe.py
@@ -18,7 +18,6 @@
 
 import typing
 from typing import Callable, Dict, List, Optional, Tuple
-import collections
 
 import dataclasses
 import numpy as np
@@ -29,6 +28,7 @@ from tensorflow_datasets.core import dataset_utils
 from tensorflow_datasets.core import features
 from tensorflow_datasets.core import lazy_imports_lib
 from tensorflow_datasets.core.utils.type_utils import TreeDict
+from tensorflow_datasets.core.utils.py_utils import _flatten, _flatten_with_path
 
 try:
   import pandas  # pylint: disable=g-import-not-at-top
@@ -139,31 +139,6 @@ class StyledDataFrame(DataFrame):
       return super()._repr_html_()
     return self.__styler._repr_html_()  # pylint: disable=protected-access
 
-
-def _flatten_with_path(structure: TreeDict):
-  """Convert a TreeDict into a flat list of paths and their values
-
-  A path is a list of keys to get to the value
-  Yields the result elements instead of creating a list
-  """
-  if isinstance(structure, collections.Mapping):
-    for key in sorted(structure):
-      for sub_path, sub_value in _flatten_with_path(structure[key]):
-        yield [key] + sub_path, sub_value
-  else:
-    yield [], structure
-
-def _flatten(structure: TreeDict):
-  """Convert a TreeDict into a flat list of values without the keys
-
-  Yields the result elements instead of creating a list
-  """
-  if isinstance(structure, collections.Mapping):
-    for key in sorted(structure):
-      for sub_value in _flatten(structure[key]):
-        yield sub_value
-  else:
-    yield structure
 
 def _make_columns(
     specs: TreeDict[tf.TypeSpec],
--- a/tensorflow_datasets/core/as_dataframe.py
+++ b/tensorflow_datasets/core/as_dataframe.py
@@ -18,6 +18,7 @@
 
 import typing
 from typing import Callable, Dict, List, Optional, Tuple
+import collections
 
 import dataclasses
 import numpy as np
@@ -28,7 +29,6 @@ from tensorflow_datasets.core import dataset_utils
 from tensorflow_datasets.core import features
 from tensorflow_datasets.core import lazy_imports_lib
 from tensorflow_datasets.core.utils.type_utils import TreeDict
-import tree
 
 try:
   import pandas  # pylint: disable=g-import-not-at-top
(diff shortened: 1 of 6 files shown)


tensorflow_datasets/core/as_dataframe.py
```diff
@@ -140,14 +140,39 @@ def _repr_html_(self) -> str:
...
+  Yields the result elements instead of creating a list
+  """
+  if isinstance(structure, collections.Mapping):
+    for key in sorted(structure):
+      for sub_path, sub_value in _flatten_with_path(structure[key]):
+        yield [key] + sub_path, sub_value
+  else:
+    yield [], structure
+
+def _flatten(structure: TreeDict):
```
Conchylicultor, line 156: Could `tf.nest.flatten` be used instead ?

tensorflow_datasets/core/as_dataframe_test.py
```diff
@@ -37,6 +37,53 @@ def _as_df(ds_name: str, **kwargs) -> pandas.DataFrame:
   return df
 
 
+def test_flatten_with_path():
+  """Test that the flatten function works as expected"""
+  from from tensorflow_datasets.core.as_dataframe import _flatten_with_path
+  assert list(_flatten_with_path('value')) == [([], 'value')]
+  assert list(_flatten_with_path({'key': 'value'})) == [('key', 'value')]
+  assert list(_flatten_with_path({'key1': 'value', 'key2': 'value2'})) == [
```
Conchylicultor, line 45: I would invert key2 and key1, to make sure that keys are correctly sorted

tensorflow_datasets/core/as_dataframe.py
```diff
@@ -140,14 +140,39 @@ def _repr_html_(self) -> str:
     return self.__styler._repr_html_()  # pylint: disable=protected-access
 
 
+def _flatten_with_path(structure: TreeDict):
```
Conchylicultor, line 143: Could this be moved into `py_utils` ? And the tests in `py_utils_test`. Note: pytest support both unittest-style and pytest-style tests, so even if `py_utils_test.py` currently uses `unittest-style`, you can keep the pytest-style function which are better.

tensorflow_datasets/core/as_dataframe_test.py
```diff
@@ -37,6 +37,53 @@ def _as_df(ds_name: str, **kwargs) -> pandas.DataFrame:
...
+      'key2': 'value2',
+    }
+  }
+  assert list(_flatten_with_path(complex_dict)) == [
+    (['key'], 'value'),
+    (['key2'], 'value2'),
+    (['nested', 'subkey'], 'subvalue'),
+    (['nested', 'sub_nested', 'subsubkey1'], 'subsubvalue1'),
+    (['nested', 'sub_nested', 'subsubkey2'], 'subsubvalue2'),
+  ]
```
Conchylicultor, line 66: I'll also add a test which check that `flatten_with_path()` sorted order is consistent with `tf.nest.flatten`.

Conchylicultor (changes requested):
Thank you!

tensorflow_datasets/core/as_dataframe.py
Flamefire: I guess yes. Although I'd argue that this is faster due to handling less cases and yielding instead of manifesting the whole list. I don't mind either way, your decision

tensorflow_datasets/core/as_dataframe_test.py
Flamefire: Sure, can do. To be sure: Consistency with tf.nest.flatten or _flatten (if that should be kept)?

tensorflow_datasets/core/as_dataframe.py
Conchylicultor: I haven't benchmarked, but  as `tf.nest.flatten`, `tree.flatten` are implemented in C/C++, I would still expect them to be faster than Python-based implementations.

For consistency, let's not try to have duplicated API. It will make the migration simpler if we replace all `tf.nest.Xyz` by `tree.Xyz`.

tensorflow_datasets/core/as_dataframe_test.py
Conchylicultor: Yes, in theory, it should be the case, as `tf.nest.flatten` also sort the keys. But it's best to guarantee this with a test.

Conchylicultor (changes requested):
Thanks for the update

tensorflow_datasets/core/as_dataframe_test.py
Flamefire: Looks like TF has `flatten_with_tuple_paths_up_to` which is mostly like `flatten_with_path`. I'll see if that can be used too.

Conchylicultor: > Looks like TF has flatten_with_tuple_paths_up_to which is mostly like flatten_with_path. I'll see if that can be used too.

This would have been  nice, but as far as I know, this was part of tf.contrib, which is deprecated and not included in TF anymore.

As far as I know, this is the current symbol for the TF API: https://www.tensorflow.org/api_docs/python/tf/nest ?

Flamefire:
Updated the PR according to the comments. Please check if that is sufficient or any further changes are required.

Conchylicultor (approved):
Thank you. This will be merged soon.

tensorflow_datasets/core/as_dataframe.py
```diff
@@ -28,7 +28,7 @@
 from tensorflow_datasets.core import features
 from tensorflow_datasets.core import lazy_imports_lib
 from tensorflow_datasets.core.utils.type_utils import TreeDict
-import tree
+from tensorflow_datasets.core.utils.py_utils import _flatten, _flatten_with_path
```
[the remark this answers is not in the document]
Flamefire: Sure. I just wasn't sure if those should be public in the first place. E.g. if you want to unconditionally switch to dm-tree (hopefully after they stopped using Bazel) then this functions could be removed again. If you make them public you'd need to keep them or raise the major version.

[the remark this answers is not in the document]
Conchylicultor: The functions are public for modules which are using `py_utils`, but are not exposed in the public API: https://www.tensorflow.org/datasets/api_docs/python/tfds
Those are two different "level" of public.

PR #2552, opened 2020-10-08
Make sure tfds.load supports loading datasets from GCS without the builder class imported

carlthome:
Background: https://github.com/tensorflow/datasets/pull/2551

The other bug I found for getting TFDS 4 to support a GCS data_dir.

`tfds.builder` assumes that the `name` string contains a config and version when the dataset builder class is missing. Unfortunately, `tfds.load` mutates `name` and removes config and version before calling the builder, so no existing versions will be discovered when the class is missing, even when they actually do exist.

I'm curious if my thinking is right on this, because there should already be some test case covering this behavior, I assume?

### DIFF
--- a/tensorflow_datasets/core/load.py
+++ b/tensorflow_datasets/core/load.py
@@ -332,9 +332,8 @@ def load(
   """
   # pylint: enable=line-too-long
 
-  name, name_builder_kwargs = _dataset_name_and_kwargs_from_name_str(name)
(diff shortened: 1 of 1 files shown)


Conchylicultor (approved):
Nice catch. This bug actually affect all `tfds.load` calls. I only tested `tfds.builder`, so missed that `_dataset_name_and_kwargs_from_name_str` was called twice.

PR #2564, opened 2020-10-12
checksums.py: use importlib_resources

### LINKED ISSUE #2561: [Cleanup] Rewrite checksums.py to use `importlib_resources`
Currently, https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/download/checksums.py read checksums using relative dir: https://github.com/tensorflow/datasets/blob/0c308e4b7be9d0c834d0021c6cf6566f6cd57c00/tensorflow_datasets/core/download/checksums.py#L28

This is hacky when more modern https://docs.python.org/3/library/importlib.html#importlib.resources.files exists with a Pathlib-like interface.

```py
path = importlib_resources.files('tensorflow_datasets') / 'url_checksums'
files = [p for p in path.listdir() if str(p).endswith('txt')]
```

It would make TFDS work even when Python is executed inside compressed `.zip` archive.

iamsdas:
Fix #2561

### DIFF
--- a/tensorflow_datasets/core/download/checksums.py
+++ b/tensorflow_datasets/core/download/checksums.py
@@ -19,6 +19,7 @@
 
 import os
 from typing import Any, Dict, Iterable, List, Optional
+import importlib_resources
 
 import dataclasses
 import tensorflow.compat.v2 as tf
@@ -26,10 +27,9 @@ import tensorflow.compat.v2 as tf
 from tensorflow_datasets.core import utils
 
 
-_ROOT_DIR = os.path.normpath(os.path.join(os.path.dirname(__file__), '../..'))
-
 _CHECKSUM_DIRS = [
-    os.path.join(_ROOT_DIR, 'url_checksums'),
+  os.path.normpath(
+    importlib_resources.files('tensorflow_datasets') / 'url_checksums'),
 ]
 _CHECKSUM_SUFFIX = '.txt'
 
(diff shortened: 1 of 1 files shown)


iamsdas:
fixed conflicts with master branch

tensorflow_datasets/core/download/checksums.py
```diff
@@ -19,17 +19,17 @@
...
 
 from tensorflow_datasets.core import utils
 
 
-_ROOT_DIR = os.path.normpath(os.path.join(os.path.dirname(__file__), '../..'))
-
 _CHECKSUM_DIRS = [
-    os.path.join(_ROOT_DIR, 'url_checksums'),
+  os.path.normpath(
+    importlib_resources.files('tensorflow_datasets') / 'url_checksums'),
```
Conchylicultor, line 32: You could use: `utils.resources_path('tensorflow_datasets')` instead which is a wrapper of `importlib_resources.files`.

Conchylicultor (changes requested):
Thank you for your contribution. Could you update the other methods to use pathlib-like methods like `path.read_text()` ?

tensorflow_datasets/core/download/checksums.py
iamsdas: done

tensorflow_datasets/core/download/checksums.py
```diff
@@ -19,16 +19,15 @@
...
```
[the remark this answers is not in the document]
iamsdas: added required changes, could you check it out?

Conchylicultor (changes requested):
Could you run the tests and make sure they are passing ?

```sh
pytest tensorflow_datasets/core/download/
pytest tensorflow_datasets/core/
```

iamsdas:
Fixed a typo `utils.resources_path -> utils.resource_path`. Now it passes all the tests.

PR #2571, opened 2020-10-14
added verify to DownloadConfig/Manager

jackd:
Previous versions of tfds (< 3.0 ?) allowed downloading from servers with invalid TLS certificates, but this is no longer the case, invalidating some existing external `DatasetBuilder`s. This PR allows users to specify `verify=False` in `DownloadConfig` to get around this requirement.

The following demonstrates an existing external package which can upgrade to recent `tfds` versions because with this PR applied.

```bash
pip install tensorflow-graphics --no-deps
```

```python
import tensorflow_graphics.datasets.modelnet40 # pylint: disable=unused-import

builder = tfds.builder('model_net40')
config = tfds.core.download.DownloadConfig(verify=False)
builder.download_and_prepare(download_config=config)
```


jackd:
... this has been bugging me for a while, and I decide to put in a PR an hour after it gets resolved? I guess that's a good thing.

Issue #2422, opened 2020-09-11
Allow users to skip download and use manually downloaded data instead.


Conchylicultor:
Users often encounter issue during download (e.g. VPN, dead link,...). But they may have already downloaded the data themself.

Currently, users are forced to rely on the automatic `DownloadManager` download. Instead, we should add a mode to check whether the download already exists in the `DownloadConfig.manual_dir`, and skip the download if so.

Mini-design doc:
* In order to detect whether the file is present or not in the `manual/` dir, we should update `record_checksums=True` mode to save the real name in `checksums.tsv`.
* When the user `download_and_prepare()` and `dl_manager.download` is called:
   * If checksums are not-found for the url, then try to download the url as usual
   * If `checkusms.tsv` file is present but the `manual/<filename>` not found, try to download the url as usual
   * If `manual/<filename>` is found, should validate the checksums (cached with some `<filename>.INFO`, similar to https://github.com/tensorflow/datasets/blob/3e90c815a61db5236063817cf699f2bd4d7eef58/tensorflow_datasets/core/download/download_manager.py#L310), then return the path.

harsh020:
Hi, @Conchylicultor I am very interested to work on this issue, but I am new here so, would like your help in it.
I went through `DatasetBuilder` class and `DownloadManager` class, what I didn't understand is that the `DatasetBuilder` class already has a `data_dir` which if exists would not execute the entire `download_and_prepare()`, so I am a bit unclear on where you want it to be implemented.
I would be very grateful if you provided some resources about TFDS's `checksum`, `record_checksums` and `checksums.tsv`

Thanks.

harsh020:
Hi, @Conchylicultor I was trying to understand how and what to do, for this issue. I think I figured out what to do. So, what I understood is -

- we need to add a `record_checksum` flag to `store_checksums()` to store the `dataset_name` along with other details.

- and when the user uses `download_and_prepare` we need to check if `dataset_name` is present and also the `manual/<dataset_name>` is preset, if so, no need to download the dataset.

Kindly, let me know, if I figured it out correctly.

Thanks.

Conchylicultor:
Thank you for looking into it.

As a first step: https://github.com/tensorflow/datasets/blob/09e720f0755c3e808a562dbb49de9c0368463116/tensorflow_datasets/core/download/checksums.py#L36

should be updated to also include the dataset name:

```py
@dataclasses.dataclass(eq=True)
class UrlInfo():
  size: int
[...]

PR #2601, opened 2020-10-17
Check Manual downloads

harsh020:
## Part 2 of 2 | Check if file was manually downloaded

PR for issue #2422

Tasks -

* [x] Check if file was downloaded manually
* [ ] Write `.INFO` file


tensorflow_datasets/core/download/download_manager.py
```diff
@@ -444,6 +459,14 @@ def _download(self, resource: Union[str, resource_lib.Resource]):
       resource = resource_lib.Resource(url=resource)
     url = resource.url
 
+    manually_downloaded_path = self._check_manually_downloaded(url)
```
Conchylicultor, line 462: Could you add a test in `download_manager_test.py` ?

tensorflow_datasets/core/download/download_manager.py
```diff
@@ -421,6 +421,21 @@ def _find_existing_path(self, url: str, url_path: str) -> Optional[str]:
...
+    url_info = self._url_infos.get(url)
+    # If url info was recorded with filename
+    manually_downloaded_path = None
+    if url_info and url_info.filename:
+        manually_downloaded_path = os.path.join(
+          self._manual_dir,
+          url_info.filename
+        )
+        # File not manually downloaded
+        if not tf.io.gfile.exists(manually_downloaded_path):
```
Conchylicultor, line 435: Could you also check that the checksums match `url_info.checksums` if `_force_checksums_validation` is True ?

Conchylicultor (changes requested):
Thank you @harsh020 , this looks great!

tensorflow_datasets/core/download/download_manager.py
harsh020: Hi, I have implemented the checksum check for the case when the user hasn't extracted the dataset. I'm working on the case when the user has extracted the dataset.

Thanks.

harsh020:
@Conchylicultor, I was wondering don't we need to also check if the file is extracted (`tf.io.gfile.exist(filename without extension)`) as the `url_info.filename` will just have the `basename` with extension.

Conchylicultor (approved):
Thank you! Merging this soon :)

Conchylicultor:
> @Conchylicultor, I was wondering don't we need to also check if the file is extracted (i.e, `tf.io.gfile.exist(filename without extension)`, maybe the user downloaded and extracted the dataset in the `manual_dir`) as the `url_info.filename` will just have the `basename` with extension.
> 
> Thanks.

I don't think we want the user to manually extract the data because:

1. This add additional work for the user
2. This creates additional complexity as the script might want to explicitly control the extraction (e.g. if using `dl_manager.iter_archive`, or with nested archives)

harsh020:
@Conchylicultor The `INFO` files are still left to write. Should I do it in the next commit?

Thanks.

Conchylicultor:
> @Conchylicultor The `INFO` files are still left to commit (the commit is all set to push). Should I do it in the next commit?
> 
> Thanks.

For manual downloads, I don't think we need `INFO` files. Those are mainly to track automated downloads.

Issue #2617, opened 2020-10-20
Record filenames for all checksum files.


Conchylicultor:
With https://github.com/tensorflow/datasets/pull/2601 and https://github.com/tensorflow/datasets/issues/2422, users can now manually download all datasets. However this requires the `checksums/my_dataset.txt` (legacy) and `my_datasets/checksums.tsv` (new) to record the filename.

This was only added recently, so most checksums files do not contain the checksums. We should add a new `scripts/cleanup/`, similar to https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/scripts/cleanup/url_status_checker.py

The script would query all datasets to record the name in the checksums. No need to download the full file to get the name. Just using this function on a `request.Request` object should be enough: https://github.com/tensorflow/datasets/blob/ad114d874b9b92d6474b44e0aa92bc918a9b3b94/tensorflow_datasets/core/download/downloader.py#L90

`tfds.download.checksums.load_url_infos` and `tfds.download.checksums.save_url_infos` can be used to load/save the files:

https://github.com/tensorflow/datasets/blob/ad114d874b9b92d6474b44e0aa92bc918a9b3b94/tensorflow_datasets/core/download/checksums.py#L149

Once the script has been merged (PR1), then all checksums files could be updated (PR2):

Before:

```txt
<url> <checksums> <size>
```
After:
```txt
<url> <checksums> <size> <filename>
```

harsh020:
Hi, I would like to work on this issue. I will open PR1 tonight (21 Oct).

Thanks.

harsh020:
Hi, will open PR2  as soon as PR1 gets merged.

Thanks.

Conchylicultor:
Yes, will merge this soon.
I just realise they might be an issue with drive.com url which might not be properly extracted. I'll fix those bugs internally, so I'll probably send PR2 myself.

vijayphoenix:
FIxed by 24211da0

PR #2628, opened 2020-10-21
Script to record filenames

harsh020:
## PR 1 of 2 | Script to record filenames in checksums

PR for issue #2617 

* [x] Add script to record filenames in checksums for legacy and new datasets.

### DIFF
--- a/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
+++ b/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
@@ -73,15 +73,15 @@ def _update_url_infos(url_infos) -> Dict[str, checksums.UrlInfo]:
 
 def _save_url_infos(url_infos, paths):
   """Save UrlInfo to checksums."""
-  for url, url_info in url_infos:
+  for url, url_info in url_infos.items():
     path = paths[url]
-    tfds.core.download.checksums.save_url_infos(path, url_info)
+    tfds.core.download.checksums.save_url_infos(path, {url: url_info})
 
 def main(_):
   # Legacy datasets
   url_checksums_path = checksums._checksum_paths().values()
   url_infos = {}
-  url_paths = {}
+  url_paths = {}  # save paths of checksum files
   for url_path in url_checksums_path:
     if isinstance(url_path, str):
       url_path = pathlib.Path(url_path)
@@ -92,16 +92,18 @@ def main(_):
 
   # New datasets - tfds.core.DatasetBuilder
   builder_url_infos = {}
-  builder_paths = {}
+  builder_paths = {}  # save paths of checksum files
   for name in tfds.list_builders():
       if tfds.builder_cls(name).url_infos is not None:
         url, url_info = list(tfds.builder_cls(name).url_infos.items())[0]
         builder_url_infos[url] = url_info
         builder_paths[url] = _get_builder_checksum_path(name)
 
+  # Add filename to checksums
   url_infos = _update_url_infos(url_infos)
   builder_url_infos = _update_url_infos(builder_url_infos)
 
+  # Save checksums with filename
   _save_url_infos(url_infos, url_paths)
   _save_url_infos(builder_url_infos, builder_paths)
 
--- a/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
+++ b/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
@@ -61,12 +61,14 @@ def _update_url_infos(url_infos, filenames) -> Dict[str, checksums.UrlInfo]:
 def main(_):
   with futures.ThreadPoolExecutor(max_workers=100) as executor:
     future_filenames = {}
+    cached_url_infos = {}
     # Collect legacy datasets as well as new datasets
     for url_path in _collect_path_to_update():
       url_infos = checksums.load_url_infos(url_path)
+      cached_url_infos[url_path] = url_infos
       future_filenames[url_path] = executor.map(_request_filename, url_infos)
     for path, future_filename in future_filenames.items():
-      old_url_infos = checksums.load_url_infos(path)
+      old_url_infos = cached_url_infos[path]
       updated_url_infos = _update_url_infos(old_url_infos, future_filename)
       checksums.save_url_infos(path, updated_url_infos)
 
--- a/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
+++ b/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
@@ -35,8 +35,9 @@ def _collect_path_to_update() -> List[tfds.core.utils.ReadWritePath]:
 
   builder_checksums_path = []
   for name in tfds.list_builders():
-    if tfds.builder_cls(name).url_infos != None:
-      builder_checksums_path.append(tfds.builder_cls(name)._checksums_path)
+    builder_cls = tfds.builder_cls(name)
+    if builder_cls.url_infos != None:
+      builder_checksums_path.append(builder_cls._checksums_path)
   return url_checksums_paths + builder_checksums_path
 
 def _request_filename(url) -> (str, str):
@@ -48,26 +49,26 @@ def _request_filename(url) -> (str, str):
     print(f'Success for {url}')
   except requests.exceptions.HTTPError as http_err:
     print(f'HTTP Error {http_err} for {url}.')
-  except Exception as error:
-    print(f'Error {error} for {url}.')
   return (url, filename)
 
-def _update_url_infos(url_infos) -> Dict[str, checksums.UrlInfo]:
+def _update_url_infos(url_infos, filenames) -> Dict[str, checksums.UrlInfo]:
   """Get and update dataset filname in UrlInfo."""
-  with futures.ThreadPoolExecutor(max_workers=100) as executor:
-    all_content = executor.map(_request_filename, url_infos)
-
-  for content in all_content:
-    url, filename = content
+  for url, filename in filenames:
     url_infos[url].filename = filename
   return url_infos
 
+
 def main(_):
-  # Collect legacy datasets as well as new datasets
-  for url_path in _collect_path_to_update():
-    url_infos = checksums.load_url_infos(url_path)  #load checksums
-    url_infos = _update_url_infos(url_infos)        #update checkums
-    checksums.save_url_infos(url_path, url_infos)   #save checksums
+  with futures.ThreadPoolExecutor(max_workers=100) as executor:
+    future_filenames = {}
+    # Collect legacy datasets as well as new datasets
+    for url_path in _collect_path_to_update():
+      url_infos = checksums.load_url_infos(url_path)
+      future_filenames[url_path] = executor.map(_request_filename, url_infos)
+    for path, future_filename in future_filenames.items():
+      old_url_infos = checksums.load_url_infos(path)
+      updated_url_infos = _update_url_infos(old_url_infos, future_filename)
+      checksums.save_url_infos(path, updated_url_infos)
 
 
 if __name__ == '__main__':
--- a/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
+++ b/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
@@ -74,7 +74,6 @@ def _update_url_infos(url_infos) -> Dict[str, checksums.UrlInfo]:
 def main(_):
   # Legacy datasets
   url_checksums_paths = checksums._checksum_paths().values()
-  print(url_checksums_paths)
   for url_path in list(url_checksums_paths):
     if isinstance(url_path, str):
       url_path = pathlib.Path(url_path)
--- a/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
+++ b/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
@@ -82,12 +82,12 @@ def main(_):
     checksums.save_url_infos(url_path, url_infos)   #save checksums
 
   # New datasets - tfds.core.DatasetBuilder
-  for name in tfds.list_builders()[:20]:
-      if not tfds.builder_cls(name).url_infos:
-        url_infos = tfds.builder_cls(name).url_infos  #load checksums
-        url_infos = _update_url_infos(url_infos)      #update checkums to add filename
-        path = _get_builder_checksum_path(name)       #get checksums path
-        checksums.save_url_infos(path, url_infos)     #save checksums
+  for name in tfds.list_builders():
+    if tfds.builder_cls(name).url_infos != None:
+      url_infos = tfds.builder_cls(name).url_infos  #load checksums
+      url_infos = _update_url_infos(url_infos)      #update checkums to add filename
+      path = _get_builder_checksum_path(name)       #get checksums path
+      checksums.save_url_infos(path, url_infos)     #save checksums
 
 
 if __name__ == '__main__':
--- a/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
+++ b/tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
@@ -19,6 +19,7 @@ import os
 import pathlib
 import requests
 import logging
+import dataclasses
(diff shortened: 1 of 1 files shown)


harsh020:
@Conchylicultor Hi, there seems to be a hurdle, because of
https://github.com/tensorflow/datasets/blob/a8a088364f13a854b47705cc71c8d9d5148f4c78/tensorflow_datasets/core/download/checksums.py#L45-L46
and
https://github.com/tensorflow/datasets/blob/a8a088364f13a854b47705cc71c8d9d5148f4c78/tensorflow_datasets/core/download/checksums.py#L174-L176
the script won't update the filename, as `filenames` are not included while checking equality of two checksums.

Thanks.

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -0,0 +1,94 @@
...
+from absl import app
+from typing import Dict
+from concurrent import futures
+
+import tensorflow as tf
+import tensorflow_datasets as tfds
+from tensorflow_datasets.core.download import checksums
+
+
+TFDS_PATH = tfds.core.utils.tfds_path()
```
Conchylicultor, line 30: `tfds_write_path()` as this makes it compatible with some google-internal

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -0,0 +1,94 @@
...
+  return url_infos
+
+def main(_):
+  # Legacy datasets
+  url_checksums_paths = checksums._checksum_paths().values()
+  for url_path in list(url_checksums_paths):
+    if isinstance(url_path, str):
+      url_path = pathlib.Path(url_path)
+    url_infos = checksums.load_url_infos(url_path)  #load checksums
+    url_infos = _update_url_infos(url_infos)        #update checkums to add filename
```
Conchylicultor, line 81: Rather than duplicating loading/saving checksums, it would be cleaner to re-organize the code as:

```py
# Collect both legacy and new datasets
for checksums_path in _collect_path_to_update():  # -> List[tfds.core.utils.ReadWritePath]
  url_infos = checksums.load_url_infos(checksums_path)
  ...
```

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -0,0 +1,94 @@
...
+    checksums.save_url_infos(url_path, url_infos)   #save checksums
+
+  # New datasets - tfds.core.DatasetBuilder
+  for name in tfds.list_builders():
+    if tfds.builder_cls(name).url_infos != None:
+      url_infos = tfds.builder_cls(name).url_infos  #load checksums
+      url_infos = _update_url_infos(url_infos)      #update checkums to add filename
+      path = _get_builder_checksum_path(name)       #get checksums path
```
Conchylicultor, line 89: It would be best to use `tfds.builder_cls(name)._checksum_path`

Conchylicultor (changes requested):

Conchylicultor:
Thank you! Looks good

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -19,34 +19,25 @@
...
-  if isinstance(path, str):
-    path = pathlib.Path(path)
-  return path
+def _collect_path_to_update() -> List[tfds.core.utils.ReadWritePath]:
+  """Collect checksums paths to update."""
+  url_checksums_paths = list(checksums._checksum_paths().values())
+
+  builder_checksums_path = []
+  for name in tfds.list_builders():
+    if tfds.builder_cls(name).url_infos != None:
```
Conchylicultor, line 38: When the same function is used multiple times, it's best to cache it. Here `builder_cls = tfds.builder_cls(name)`

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -0,0 +1,74 @@
...
+def _request_filename(url) -> (str, str):
+  """Get filename of dataset at `url`."""
+  filename  = ''
+  try:
+    response = requests.get(url, timeout=10)
+    filename = tfds.core.download.downloader._get_filename(response)
+    print(f'Success for {url}')
+  except requests.exceptions.HTTPError as http_err:
+    print(f'HTTP Error {http_err} for {url}.')
+  except Exception as error:
```
Conchylicultor, line 51: Broad except should be avoided as it hides bugs. Which error are you encountering ?

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -0,0 +1,74 @@
...
+    print(f'Error {error} for {url}.')
+  return (url, filename)
+
+def _update_url_infos(url_infos) -> Dict[str, checksums.UrlInfo]:
+  """Get and update dataset filname in UrlInfo."""
+  with futures.ThreadPoolExecutor(max_workers=100) as executor:
```
Conchylicultor, line 57: Note: 
Here you're recreating a new thread pool for each datasets. So each dataset is processed individually. It would be much faster to share the same threadpool across all datasets, like:

```py
with futures.ThreadPoolExecutor(max_workers=100) as executor:
    path_to_new_url_infos = {}
    # Add request to the executor queue
    for url_path in _collect_path_to_update():
      url_infos = checksums.load_url_infos(url_path)
      path_to_new_url_infos[url_path] = executor.map(_request_filename, url_infos)
    # Process results
    for path, url_info_futures in path_to_new_url_infos.items():
      ...
```

Conchylicultor (changes requested):

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
harsh020: Oh! thats my bad. I almost forgot that a broad except will hide all the issues. Sorry, I will remove it.

Thanks.

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
harsh020: I thought to do so first, but then while implementing I realised that for separately processing the results of `Future` instances, I need to either reload the `url_infos` or cache the entire thing. So, instead I concurrently loaded and updated urls within each datasets.
No worries, I will do this in the next commit. 

Thanks.

Conchylicultor: Oh yes, `url_infos` should be cached indeed. This is not an issue. `ckecksums.get_all_url_infos()` already cache all url_infos for all datasets

Conchylicultor: Then this could be: `path_to_url_infos = _collect_path_to_url_infos() -> Dict[path, Dict[url, UrlInfo]]`.

There is still the issue that multiple datasets might use the same url which will currently downloaded twice. Let me know if you're interested by this but that might be over-optimisation.

harsh020: Thats a great idea!

Thanks.

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -0,0 +1,77 @@
...
+
+def _update_url_infos(url_infos, filenames) -> Dict[str, checksums.UrlInfo]:
+  """Get and update dataset filname in UrlInfo."""
+  for url, filename in filenames:
+    url_infos[url].filename = filename
+  return url_infos
+
+
+def main(_):
+  with futures.ThreadPoolExecutor(max_workers=100) as executor:
```
Conchylicultor, line 62: How about my suggestion from the previous message ? I updated the message aftwerwrd so you might not have been notified:

```py
path_to_url_infos = _collect_path_to_url_infos() -> Dict[path, Dict[url, UrlInfo]]

# Flatten all urls to remove duplicates across datasets
all_url_infos = {}
for url_infos in path_to_url_infos.values()
  all_url_infos.update(url_infos)

with futures.ThreadPoolExecutor(max_workers=100) as executor:
  # Process all urls in parallel
  url_to_filename = {
      url: executor.submit(_get_filename, url, url_info)
      for url, url_info in all_url_infos.items()
  }

  # Then update path_to_url_infos with filenames
  for path, url_infos in in path_to_url_infos.items():
    ...
```

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
```diff
@@ -0,0 +1,77 @@
...
+    filename = tfds.core.download.downloader._get_filename(response)
+    print(f'Success for {url}')
+  except requests.exceptions.HTTPError as http_err:
+    print(f'HTTP Error {http_err} for {url}.')
+  return (url, filename)
+
+def _update_url_infos(url_infos, filenames) -> Dict[str, checksums.UrlInfo]:
+  """Get and update dataset filname in UrlInfo."""
+  for url, filename in filenames:
+    url_infos[url].filename = filename
```
Conchylicultor, line 57: If the filename existed before but is updated with a different name, we should log a warning.

Conchylicultor (changes requested):
Thank you for the update.

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
harsh020: Oh! We want to prevent duplicate urls. Sounds great. Working on it, will update as soon as I'm done.

Thanks a lot.

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
harsh020: Sure.
Thanks

tensorflow_datasets/scripts/cleanup/url_filename_recorder.py
harsh020: I am not able to understand one thing. According to the method that you have suggested, I think we need to wait for the async call to to finish, and when we get all the filenames then we can start updating `url_infos` with filenames, this makes the threading useless. So, if we need to accomplish it, I think we need one more dictionary to map `url->List[paths]`. 
I'm working on it. Please do let me know.

Thanks.

harsh020:
@Conchylicultor Hi, thanks for the suggestion. The problem is that our issue requires us to process and update `url_infos` as per the `paths`, but to implement concurrency we have to process and update them as per `urls`. As futures.submit doesn't follow sequence, so we can get the `url` which has fetched filename using `futures.as_completed` and for that `url` we can simultaneously process and update `url_infos`. But since a `url` can have multiple paths(checksums) so, to implement this we further require a dict `url->List[paths]`.
Should I go ahead and implement it, or think of some other way of doing so.

Thanks.

Conchylicultor:
> @Conchylicultor Hi, thanks for the suggestion. The problem is that our issue requires us to process and update `url_infos` as per the `paths`, but to implement concurrency we have to process and update them as per `urls`. As futures.submit doesn't follow sequence, so we can get the `url` which has fetched filename using `futures.as_completed` and for that `url` we can simultaneously process and update `url_infos`. But since a `url` can have multiple paths(checksums) so, to implement this we further require a dict `url->List[paths]`.
> Should I go ahead and implement it, or think of some other way of doing so.
> 
> Thanks.

To get the filenames while they are still processed, you can use the `future.result()` from within the `futures.ThreadPoolExecutor(max_workers=100)`, like:

```py

with futures.ThreadPoolExecutor(max_workers=100) as executor:
  # Process all urls in parallel
  url_to_filename = {
      url: executor.submit(_get_filename, url, url_info)
      for url, url_info in all_url_infos.items()
  }

  # Then update path_to_url_infos with filenames
  for path, url_infos in in path_to_url_infos.items():
    url_infos = {
        dataclasses.replace(url_info, filename=url_to_filename[url_info.filenames].result())
        for url, url_info in  url_infos.items()
    }
```

`url_to_filename[url_info.filenames].result()`

harsh020:
> > @Conchylicultor Hi, thanks for the suggestion. The problem is that our issue requires us to process and update `url_infos` as per the `paths`, but to implement concurrency we have to process and update them as per `urls`. As futures.submit doesn't follow sequence, so we can get the `url` which has fetched filename using `futures.as_completed` and for that `url` we can simultaneously process and update `url_infos`. But since a `url` can have multiple paths(checksums) so, to implement this we further require a dict `url->List[paths]`.
> > Should I go ahead and implement it, or think of some other way of doing so.
> > Thanks.
> 
> To get the filenames while they are still processed, you can use the `future.result()` from within the `futures.ThreadPoolExecutor(max_workers=100)`, like:
> 
> ```python
> with futures.ThreadPoolExecutor(max_workers=100) as executor:
>   # Process all urls in parallel
>   url_to_filename = {
>       url: executor.submit(_get_filename, url, url_info)
>       for url, url_info in all_url_infos.items()
>   }
> 
>   # Then update path_to_url_infos with filenames
>   for path, url_infos in in path_to_url_infos.items():
>     url_infos = {
>         dataclasses.replace(url_info, filename=url_to_filename[url].result())
>         for url, url_info in  url_infos.items()
>     }
> ```
> 
> See the `url_to_filename[url].result()`. I believe calling `.result()` multiple times on the same future returns the same value.

I'm sorry for being annoying, I am currently doing it this way, but I'm concerned that when we do `url_to_filename[url].result()` and if this `url` has not fetched its filename yet, then the `executor` will wait for it to be executed thus wasting time. Will it not?

Thanks.

Conchylicultor:
> I'm sorry for being annoying, I am currently doing it this way (i.e. using `result()` while still processing), but I'm concerned that when we do `url_to_filename[url].result()` and if this `url` has not fetched its filename yet (as `executor.submit()` is not sequential), then the `executor` will wait for it to be executed thus wasting time. Will it not?
> 
> Thanks.

Yes, `url_to_filename[url].result()` will wait if the url hasn't been fetched yet, but in practice this shouldn't be an issue. I assume that task added in `executor.submit` will be processed in the same order as they are added (even if some tasks might finished before others). Here `result()` are called in about same order as they are submitted so it shouldn't matter. Additionally, as the bottleneck is the internet network. Reading/writing url_info files should be almost instantaneous (a few seconds max), so even if it was waiting for all urls to complete, total time should still be much faster as previous implementation.

harsh020:
> > I'm sorry for being annoying, I am currently doing it this way (i.e. using `result()` while still processing), but I'm concerned that when we do `url_to_filename[url].result()` and if this `url` has not fetched its filename yet (as `executor.submit()` is not sequential), then the `executor` will wait for it to be executed thus wasting time. Will it not?
> > Thanks.
> 
> Yes, `url_to_filename[url].result()` will wait if the url hasn't been fetched yet, but in practice this shouldn't be an issue. I assume that task added in `executor.submit` will be processed in the same order as they are added (even if some tasks might finished before others). Here `result()` are called in about same order as they are submitted so it shouldn't matter. Additionally, as the bottleneck is the internet network. Reading/writing url_info files should be almost instantaneous (a few seconds max), so even if it was waiting for all urls to complete, total time should still be much faster as previous implementation.

Ohh! Now I get it clearly. Thanks a lot.

Conchylicultor (approved):
Thank you!

harsh020:
@Conchylicultor Hi. I am unable to understand why the kokoro tests are failing. I just see `tensorflow_datasets/gh_testing/presubmit[default] -> FAILED`. Could you please help me understand it?

Thanks.

vijayphoenix:
@harsh020 Click on the link and navigate all the way till you see the testing logs.

Since Conchylicultor approved the PR, you need not change anything! (The changes might have been already merged internally)

Also, the current test failure seems to be unrelated with your PR  
```
=========================== short test summary info ============================
FAILED tensorflow_datasets/object_detection/waymo_open_dataset_test.py::WaymoOpenDatasetTest::test_download_and_prepare_as_dataset
==== 1 failed, 1532 passed, 339 skipped, 7255 warnings in 431.09s (0:07:11) ====

```

harsh020:
@vijayphoenix Got it. Thanks a lot.

Issue #2474, opened 2020-09-22
WMT14 de-en "AssertionError: Sizes do not match"


jiunoh:
**Short description**
`tfds.load(data_dir='./tensorflow_datasets', name='wmt14_translate/de-en', with_info=True, as_supervised=True)`
this gets
```
**AssertionError: Sizes do not match: 201854 vs 202002** for ../tensorflow_datasets/downloads/extracted/TAR_GZ.statmt.org_wmt14_training-parallel-nc-v9Kk5PK6ZZROqhtS0FWYopB3pKxYzPXQowHgBUqV7eV3U.tgz/training/news-commentary-v9.de-en.de vs ../tensorflow_datasets/downloads/extracted/TAR_GZ.statmt.org_wmt14_training-parallel-nc-v9Kk5PK6ZZROqhtS0FWYopB3pKxYzPXQowHgBUqV7eV3U.tgz/training/news-commentary-v9.de-en.en.
```
again and again.

**Environment information**
* Operating System: <Ubuntu 18.04>
* Python version: <3.6.9>
* `tensorflow-datasets`/`tfds-nightly` version: tfds-nightly, also tried tensorflow-datasets==3.2.1
* `tensorflow`/`tf-nightly` version: tf-nightly, also tried tensorflow-gpu==2.3.0

* Does the issue still exists with the last `tfds-nightly` package (`pip install --upgrade tfds-nightly`) ? yes

**Reproduction instructions**

```
`tfds.load(data_dir='./tensorflow_datasets', name='wmt14_translate/de-en', with_info=True, as_supervised=True)`
```

**Link to logs**

**Expected behavior**
I want the WMT2014 de-en dataset.

**Additional context**
The error always occurs for news-commentary. Also, I got 'All attempts to get a Google authentication bearer token failed' message.
full log:

```
2020-09-22 06:32:16.080181: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudart.so.10.1
in main function
2020-09-22 06:32:18.712241: W tensorflow/core/platform/cloud/google_auth_provider.cc:184] All attempts to get a Google authentication bearer token failed, returning an empty token. Retrieving token from files failed with "Not found: Could not locate the credentials file.". Retrieving token from GCE failed with "Failed precondition: Error executing an HTTP request: libcurl code 6 meaning 'Couldn't resolve host name', error details: Couldn't resolve host 'metadata'".
[...]

PR #2633, opened 2020-10-22
Change splitlines() to split("\n") to avoid AssertionErrors caused by…

lisiyuan656:
[Issue#2474](https://github.com/tensorflow/datasets/issues/2474) seems to be caused by several "\r" separators in the original WMT datasets. Changing splitlines() to split("\n") can solve the problem.

### DIFF
--- a/tensorflow_datasets/translate/wmt.py
+++ b/tensorflow_datasets/translate/wmt.py
@@ -792,7 +792,7 @@ def _parse_parallel_sentences(f1, f2):
     if split_path[-1] == "gz":
       lang = split_path[-2]
       with tf.io.gfile.GFile(path, "rb") as f, gzip.GzipFile(fileobj=f) as g:
-        return g.read().decode("utf-8").splitlines(), lang
+        return g.read().decode("utf-8").split("\n"), lang
 
     if split_path[-1] == "txt":
       # CWMT
@@ -801,7 +801,7 @@ def _parse_parallel_sentences(f1, f2):
     else:
       lang = split_path[-1]
     with tf.io.gfile.GFile(path) as f:
-      return f.read().splitlines(), lang
+      return f.read().split("\n"), lang
 
   def _parse_sgm(path):
     """Returns sentences from a single SGML file."""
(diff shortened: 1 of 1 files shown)


Conchylicultor:
Thank you for the PR!

Could you give more context on the error ? Why is `splitlines()` affecting the number of elements ?
Will this fix also work on Windows ?

lisiyuan656:
I found that there are lines in the datasets which contain "\r" in it. Splitlines() will split this kind in two different lines. Like this one from news-commentary-v9.fr-en.fr: "Horaires de travail.\r Planifier le programme de travail afin d'éviter des conflits entre demandes et responsabilités sortant du cadre du travail ; \r\n".

I didn't test the fix on Windows, but I checked the datasets and they all have "\n" at the end of every line. So I think it should work on Windows.

NikhilBartwal (approved):
@lisiyuan656 The fix does work on windows and it looks like it should solve the assertion error. Have you tried generating the dataset with the fix?

lisiyuan656:
Yeah, it works without problem.

Conchylicultor (approved):
Thank you for the fix. Sorry for the delay

ariG23498:
I have tried with the following snippet. I think this is the same as that of Issue #2474 

```python
datasets, metadata = tfds.load('wmt14_translate/fr-en',
                               with_info=True,
                               as_supervised=True)
```

Stack Trace:
```
AssertionError                            Traceback (most recent call last)
<ipython-input-11-bfee9c36ea57> in <module>()
      3 
      4 datasets, metadata = tfds.load('wmt14_translate/fr-en', with_info=True,
----> 5                                as_supervised=True)

8 frames
/usr/local/lib/python3.6/dist-packages/tensorflow_datasets/translate/wmt.py in _parse_parallel_sentences(f1, f2)
    837     assert len(l1_sentences) == len(l2_sentences), (
    838         "Sizes do not match: %d vs %d for %s vs %s." % (
--> 839             len(l1_sentences), len(l2_sentences), f1_i, f2_i))
    840 
    841     for line_id, (s1, s2) in enumerate(zip(l1_sentences, l2_sentences)):

AssertionError: Sizes do not match: 183761 vs 183787 for /root/tensorflow_datasets/downloads/extracted/TAR_GZ.statmt.org_wmt14_training-parallel-nc-v9Kk5PK6ZZROqhtS0FWYopB3pKxYzPXQowHgBUqV7eV3U.tgz/training/news-commentary-v9.fr-en.fr vs /root/tensorflow_datasets/downloads/extracted/TAR_GZ.statmt.org_wmt14_training-parallel-nc-v9Kk5PK6ZZROqhtS0FWYopB3pKxYzPXQowHgBUqV7eV3U.tgz/training/news-commentary-v9.fr-en.en.
```

Any help would be appreciated :smile:

Conchylicultor:
@ariG23498, try using tfds-nightly

PR #2659, opened 2020-10-29
gpath: add recursive support

### LINKED ISSUE #2634: Add recursive support for `GPath`
Currently, `tf.io.gfile.glob` does not support recursive: https://github.com/tensorflow/tensorflow/issues/39565

This make `GPath().glob('**/x.txt')` inconsistent with python `glob`.

It would be nice to add recursive support. This would be done by a combinaison of `.iterdir()`, `.isdir()` (to recurse) and [`fnmatch.fnmatch()`](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch) to check that the file match the pattern.

iamsdas:
Fix #2634
Note: This is currently supposed be used something like `Gpath('src').rglob('**/*.py')`. currently it will fail if there is any absolute path before `**` in the pattern.

### DIFF
--- a/tensorflow_datasets/core/utils/gpath.py
+++ b/tensorflow_datasets/core/utils/gpath.py
@@ -89,6 +89,16 @@ class _GPath(pathlib.PurePath, type_utils.ReadWritePath):
     """Returns the abolute path."""
     return self._new(posixpath.abspath(self._path_str))
 
+  def rglob(self: _P, pattern: str) -> Iterator[_P]:
+    """
+    Yielding all matching files (of any kind) recursively.
+    """
+    yield from self.glob(pattern)
+    if '**' in pattern:
+      for f in self.iterdir():
+        if f.is_dir():
+          yield from f.rglob(pattern)
+
   def glob(self: _P, pattern: str) -> Iterator[_P]:
     """Yielding all matching files (of any kind)."""
     for f in tf.io.gfile.glob(posixpath.join(self._path_str, pattern)):
--- a/tensorflow_datasets/core/utils/gpath.py
+++ b/tensorflow_datasets/core/utils/gpath.py
@@ -18,6 +18,7 @@
 import os
 import pathlib
 import posixpath
+import fnmatch
 
 import typing
 from typing import Any, AnyStr, Iterator, Optional, Type, TypeVar
@@ -89,15 +90,21 @@ class _GPath(pathlib.PurePath, type_utils.ReadWritePath):
     """Returns the abolute path."""
     return self._new(posixpath.abspath(self._path_str))
 
+  def iglob(self: _P, pattern: str) -> Iterator[_P]:
+    for f in self.iterdir():
+      if fnmatch.fnmatch(f, pattern):
+        yield f
+      if f.is_dir():
+        yield from f.iglob(pattern)
+
   def rglob(self: _P, pattern: str) -> Iterator[_P]:
     """
     Yielding all matching files (of any kind) recursively.
     """
-    yield from self.glob(pattern)
-    if '**' in pattern:
-      for f in self.iterdir():
-        if f.is_dir():
-          yield from f.rglob(pattern)
+    assert '**' in pattern
+    pattern = posixpath.join(self._path_str, pattern)
+    pattern = pattern.partition('*')
+    return list(self._new(pattern[0]).iglob(''.join(pattern[1:])))
 
   def glob(self: _P, pattern: str) -> Iterator[_P]:
     """Yielding all matching files (of any kind)."""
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/utils/gpath.py
```diff
@@ -89,6 +89,16 @@ def resolve(self: _P, strict: bool = False) -> _P:
...
+    """
+    Yielding all matching files (of any kind) recursively.
+    """
+    yield from self.glob(pattern)
+    if '**' in pattern:
+      for f in self.iterdir():
+        if f.is_dir():
+          yield from f.rglob(pattern)
+
   def glob(self: _P, pattern: str) -> Iterator[_P]:
```
Conchylicultor, line 102: To be compatible with `pathlib.Path` API, `Path().glob('**/*.jpeg')` should also be recursive. See https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob for doc.

Conchylicultor (changes requested):
Thank you for starting this. Could you check add tests to make sure this is working ?

tensorflow_datasets/core/utils/gpath.py
iamsdas: I'm not sure I understand what you mean to say over here. Currently, `Gpath().glob('**/*.jpeg)` does work as intended. At the beginning of the function, it uses `self.glob()` to find matching patterns in the given path (which are recursively called at the bottom).

tensorflow_datasets/core/utils/gpath.py
```diff
@@ -89,6 +89,16 @@ def resolve(self: _P, strict: bool = False) -> _P:
     """Returns the abolute path."""
     return self._new(posixpath.abspath(self._path_str))
 
+  def rglob(self: _P, pattern: str) -> Iterator[_P]:
```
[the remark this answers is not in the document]
iamsdas: Although these two examples will work, something like `Gpath('/tmp').glob('a/**/*.txt)` will not (patterns with absolute path before **). This is a WIP. Im currently thinking of a hacky way to get around this by splitting the pattern and adding the part before `**` to the path and removing it from the pattern. Is this ok? I'll add the tests after doing this.

tensorflow_datasets/core/utils/gpath.py
Conchylicultor: You could have a look at the `glob.iglob` implementation to see how they are doing it. There might be smarter way to do this

iamsdas: Is there any reason to not directly use something like `pathlib.Path().rglob(pattern)` and pass it to a for loop to make it our desired type instead? The implementation of `glob.iglob` can get a bit complecated as we are trying to use `Gpath(path).rglob(pattern)` which has this exrta `path`, which is not present in `glob.glob()`.

Conchylicultor: I'm not sure I understand. `pathlib.Path().rglob(pattern)` won't works with GCS right ?
`GPath('gs://some-path/').rglob('*.jpg')`

iamsdas: Oh got it, I had no idea about that.

iamsdas: I'll see if i can do a better implementation, until then I'll send a commit which is supposed to work properly in all the cases.

tensorflow_datasets/core/utils/gpath.py
```diff
@@ -18,6 +18,7 @@
 import os
 import pathlib
 import posixpath
+from fnmatch import fnmatch
```
Conchylicultor, line 21: Direct imports are not allowed (use `import fnmatch`, then `fnmatch .fnmatch()`)

tensorflow_datasets/core/utils/gpath.py
```diff
@@ -89,6 +90,22 @@ def resolve(self: _P, strict: bool = False) -> _P:
     """Returns the abolute path."""
     return self._new(posixpath.abspath(self._path_str))
 
+  def iglob(self: _P, pattern: str) -> Iterator[_P]:
```
Conchylicultor, line 93: Rename `iglob` -> `glob`

Conchylicultor (changes requested):
Thank you!. Could you add tests in `gpath_tests.py` ?

tensorflow_datasets/core/utils/gpath.py
iamsdas: done

HoltSpalding:
Thank You!

Issue #2745, opened 2020-11-17
Wrong labels in coil100 dataset, has 72 instead of 100


duran67:
Hi, I believe the labels are wrong the dataset.

If you examine the "label" feature it says number of classes is 72, it should be 100.
`info.features["label"].num_classes`
`72 `
72 is how many images there are per class, 100 is the number of objects/classes.

If you list the labels it shows the 5-degree steps.
`info.features["label"].names`
`'0' '5' '10' '15' '20' '25' '30' ...`

Should be:
`'obj1', 'obj2', 'obj3', 'obj4', 'obj5' ...`

Maybe solution in [coil100.py](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/image/coil100.py)

`_LABELS = ['obj'+str(x) for x in range(1, 101)]`

And in `_generate_examples` I think label should be: 
	`label = int(file_name.split("__")[0][3:])-1`
instead of:
	`label = file_name.split("_")[2].split(".")[0]`

(and not sure if object_id is needed)

Conchylicultor:
As far as I understand the implementation, it seems label correspond to the angle, while object_id correspond to the object id.
I agree this is confusing.

* object_id should be changed to a `ClassLabel`
* `label` should be renamed `angle_label` 
* There likelly should be an additional `'angle': tf.int64` feature to contain the raw angle.

Don't hesitate to send a PR to update the dataset

harsh020:
@Conchylicultor Hi, I was working  on this issue, I seem to have run into an issue. When I renamed the key `label` to `angle_label`, it seemed that I was able to load the dataset successfully using `load` but I was unable to iterate the dataset using `take`.

Code:

```python
def peek(ds):
  print(ds)
  return ds

ds_train,info= tfds.load("coil100", split='train', with_info=True)
ds_ = ds_train.map(peek, num_parallel_calls=tf.data.experimental.AUTOTUNE)   #works fine

ds = ds_train.shuffle(1024).batch(32).prefetch(tf.data.experimental.AUTOTUNE) 
for d in ds.take(1):   #raises error
    print(d)
```

The peek function prints
```
{'angle_label': <tf.Tensor 'args_0:0' shape=() dtype=int64>, 'image': <tf.Tensor 'args_1:0' shape=(128, 128, 3) dtype=uint8>, 'object_id': <tf.Tensor 'args_2:0' shape=() dtype=string>}
```

The error it raises is
```
[...]

PR #2769, opened 2020-11-23
Update coil100 dataset

harsh020:
## Update coil100 dataset

Fixes issue #2745 

Tasks:
* [x] Rename `label` to `angle_label`
* [x] Convert `object_id` to `ClassLabel`
* [x] Create `angle` key for `raw_angles` of `tf.int64` dtype.  
* [x] Update documentations

### DIFF
--- a/tensorflow_datasets/image/coil100.py
+++ b/tensorflow_datasets/image/coil100.py
@@ -17,6 +17,7 @@
 
 import os
 
+import numpy as np
 import tensorflow.compat.v2 as tf
 import tensorflow_datasets.public_api as tfds
 
@@ -88,7 +89,7 @@ class Coil100(tfds.core.GeneratorBasedBuilder):
       if file_name.endswith(".png"):
         image = os.path.join(data_dir_path, file_name)
         angle_label = file_name.split("_")[2].split(".")[0]
-        angle = tf.cast(int(angle_label), tf.int64)
+        angle = np.int64(angle_label)
         object_id = file_name.split("_")[0]
         yield file_name, {
             "image": image,


harsh020:
@Conchylicultor Hi, currently, I have just committed the changes where I have just renamed `label` to `angle_label`. It is giving the error that I mentioned [here](https://github.com/tensorflow/datasets/issues/2745#issuecomment-731393146).

The tasks of - 
* Convert object_id to ClassLabel
* Create angle key for raw_angles of tf.int64 dtype.
Also raise the same error.

Thank You.

docs/catalog/coil100.md
```diff
@@ -44,12 +44,13 @@ Split  | Examples
 ```python
```
Conchylicultor, line 44: Doc is automatically generated

Conchylicultor (approved):
Thank you.

Note: `Version` should be updated and RELEASE_NOTES added when updating a dataset

tensorflow_datasets/image/coil100.py
```diff
@@ -85,10 +87,12 @@ def _generate_examples(self, data_dir_path):
     for file_name in tf.io.gfile.listdir(data_dir_path):
       if file_name.endswith(".png"):
         image = os.path.join(data_dir_path, file_name)
-        label = file_name.split("_")[2].split(".")[0]
+        angle_label = file_name.split("_")[2].split(".")[0]
+        angle = tf.cast(int(angle_label), tf.int64)
```
Conchylicultor, line 91: Did you actually tested this code ? _generate_examples should yield np.array compatible values, not tf.Tensor.

harsh020: I tested the updated dataset with `load` and it seem to be yielding each example as a `Dict`. 
The signature of each example is as -
```
{
  'angle': <tf.Tensor 'args_0:0' shape=() dtype=int64>, 
  'angle_label': <tf.Tensor 'args_1:0' shape=() dtype=int64>, 
  'image': <tf.Tensor 'args_2:0' shape=(128, 128, 3) dtype=uint8>, 
  'object_id': <tf.Tensor 'args_3:0' shape=() dtype=int64>
}
```

Thank you.

harsh020: Ok! I think you mean instead of `tf.cast` we should use `numpy`conversion for `np.array` compatibility.

harsh020:
@Conchylicultor Have resolved the merge conflict, kindly merge into master as currently the master does not contain `angle` feature.

Thank you.

Issue #2791, opened 2020-11-25
Better `tfds.as_dataframe` visualization


Conchylicultor:
TFDS visualisation is helpful but currently a little too limited:

For example, [COCO](https://www.tensorflow.org/datasets/catalog/coco) visualisation makes it difficult to view the bounding boxes, labels:

![image](https://user-images.githubusercontent.com/9047355/100262249-2df54200-2f4c-11eb-9786-3ba9d90b96f6.png)

Here are a few ideas to improve the current visualisation:
Done:

* [x] `BBoxFeature` inside `Sequence`: Display some boxes in some empty 150x150px image
* [x] Video could also be greatly improved. If fmmpeg is found, it would be nice to generate a proper `.mp4` video which would be more compressed than `.gif` and would have play/pause controls.
* [x] The video framerate should also be greatly reduced. Currently, video like https://www.tensorflow.org/datasets/catalog/abstract_reasoning are much too fast

Todo:

* [x] `ClassLabel`/`Tensor` inside `Sequence`: Display individual `repr_html`, once per line in a `div` with scrollbars

For the implementation, dataframe will automatically call the `. repr_html()` or `repr_html_batch` of the corresponding feature.

For example:

```python
features={
    'image': tfds.features.Image(),
    'objects': tfds.features.Sequence({
        'bbox': tfds.features.BBoxFeature(),
    })
}
```
Will call `tfds.features.Image.repr_html` and `tfds.features.BBoxFeature.repr_html_batch`.

https://github.com/tensorflow/datasets/blob/ad5373e97c37569e9bf024840ba6d6503f8b1116/tensorflow_datasets/core/features/audio_feature.py#L89

harsh020:
@Conchylicultor Hi, I was working on it. Could you please tell me if there is any way to preview **locally** how my changes will be rendered in the frontend?

Thank You.

Conchylicultor:
> Could you please tell me if there is any way to preview locally how my changes will be rendered in the frontend?

Thank you @harsh020 . To preview locally, you can either:
* Test locally in a local jupyter notebook.
* Save the `tfds.as_dataframe(...)._repr_html_()` content in a `.html` file and view it on your browser.

> The implementations will be in repr_html_batch, isn't it?

Yes, for all features wrapped inside `Sequence` (e.g. `Sequence((my_feature))`), `my_feature.repr_html_batch(ex)` will be called during visualization. For other features, `repr_html` is called.

harsh020:
[...]

PR #2798, opened 2020-11-27
[ENC] Better `tfds.as_dataframe` visualization

harsh020:
### Better `tfds.as_dataframe` visualization

Fixes issue #2791 

Tasks:
* [x] `BBoxFeature` inside `Sequence`: Display some boxes in some empty 150x150px image
* [ ] `ClassLabel/Tensor` inside `Sequence`: Maybe display a dropdown list with individual values
* [x] Video could also be greatly improved. If ffmpeg is found, it would be nice to generate a proper `.mp4` video which would be more compressed than .gif and would have play/pause controls.
* [x] The video framerate should also be greatly reduced. Currently, video like https://www.tensorflow.org/datasets/catalog/abstract_reasoning are much too fast

### DIFF
--- a/tensorflow_datasets/core/features/bounding_boxes.py
+++ b/tensorflow_datasets/core/features/bounding_boxes.py
@@ -101,6 +101,7 @@ class BBoxFeature(feature.Tensor):
 
   def repr_html(self, ex: np.ndarray) -> str:
     """Returns the HTML str representation of an Image with BBoxes."""
+    ex = np.expand_dims(ex, axis=0)  # Convert single bounding box to single batch.
     return self._repr_html(ex)
 
   def repr_html_batch(self, ex: np.ndarray) -> str:
--- a/tensorflow_datasets/core/features/bounding_boxes.py
+++ b/tensorflow_datasets/core/features/bounding_boxes.py
@@ -83,7 +83,7 @@ class BBoxFeature(feature.Tensor):
     )
 
   def _build_thumbnail_with_bbox(self, ex: np.ndarray) -> str:
-    """Returns the HTML str representation of an Image with BBoxes."""
+    """Returns blank image with Bboxes drawn on it."""
     PIL_Image = lazy_imports_lib.lazy_imports.PIL_Image
     PIL_ImageDraw = lazy_imports_lib.lazy_imports.PIL_ImageDraw
 
@@ -100,11 +100,9 @@ class BBoxFeature(feature.Tensor):
     return blank_img
 
   def repr_html_batch(self, ex: np.ndarray) -> str:
-    """Returns the HTML str representation of the object."""
+    """Returns the HTML str representation of an Image with BBoxes."""
     img = self._build_thumbnail_with_bbox(ex)
-
     img_str = utils.get_base64(lambda buff: img.save(buff, format='PNG'))
-
     return f'<img src="data:image/png;base64,{img_str}" alt="Img" />'
 
   @classmethod
--- a/tensorflow_datasets/core/features/bounding_boxes.py
+++ b/tensorflow_datasets/core/features/bounding_boxes.py
@@ -90,12 +90,13 @@ class BBoxFeature(feature.Tensor):
     SIZE = 150
     blank_img = PIL_Image.new('RGB', (SIZE, SIZE), (255, 255, 255))
     draw = PIL_ImageDraw.Draw(blank_img)
+    rs = np.random.RandomState(97531)  # freeze random state
 
     for i in range(ex.shape[0]):
       # Rescale coordinates to match size of blank_image
       ymin, xmin, ymax, xmax = ex[i, :]*SIZE
       # Generate random rgb values for Bbox ouline
-      r, g, b = list(np.random.randint(0, 256, size=3))
+      r, g, b = list(rs.randint(0, 256, size=3))
       draw.rectangle(((xmin, ymin), (xmax, ymax)), outline=(r, g, b))
     return blank_img
 
--- a/tensorflow_datasets/core/features/bounding_boxes.py
+++ b/tensorflow_datasets/core/features/bounding_boxes.py
@@ -85,17 +85,11 @@ class BBoxFeature(feature.Tensor):
   def repr_html(self, ex: np.ndarray) -> str:
     """Returns the HTML str representation of an Image with BBoxes."""
     ex = np.expand_dims(ex, axis=0)  # Convert single bounding box to single batch.
-    return self._repr_html(ex)
+    return _repr_html(ex)
 
   def repr_html_batch(self, ex: np.ndarray) -> str:
     """Returns the HTML str representation of an Image with BBoxes (Sequence)."""
-    return self._repr_html(ex)
-
-  def _repr_html(self, ex: np.ndarray) -> str:
-    """Returns the HTML str representation of an Image with BBoxes."""
-    img = _build_thumbnail_with_bbox(ex)
-    img_str = utils.get_base64(lambda buff: img.save(buff, format='PNG'))
-    return f'<img src="data:image/png;base64,{img_str}" alt="Img" />'
+    return _repr_html(ex)
 
   @classmethod
   def from_json_content(cls, value: Json) -> 'BBoxFeature':
@@ -104,6 +98,12 @@ class BBoxFeature(feature.Tensor):
 
   def to_json_content(self) -> Json:
     return dict()
+
+def _repr_html(ex: np.ndarray) -> str:
+  """Returns the HTML str representation of an Image with BBoxes."""
+  img = _build_thumbnail_with_bbox(ex)
+  img_str = utils.get_base64(lambda buff: img.save(buff, format='PNG'))
+  return f'<img src="data:image/png;base64,{img_str}" alt="Img" />'
 
 def _build_thumbnail_with_bbox(ex: np.ndarray):
   """Returns blank image with Bboxes drawn on it."""
--- a/tensorflow_datasets/core/features/bounding_boxes.py
+++ b/tensorflow_datasets/core/features/bounding_boxes.py
@@ -82,7 +82,7 @@ class BBoxFeature(feature.Tensor):
         [bbox.ymin, bbox.xmin, bbox.ymax, bbox.xmax]
     )
 
-  def _build_thumbnail_with_bbox(self, ex: np.ndarray) -> str:
+  def _build_thumbnail_with_bbox(self, ex: np.ndarray):
     """Returns blank image with Bboxes drawn on it."""
     PIL_Image = lazy_imports_lib.lazy_imports.PIL_Image
     PIL_ImageDraw = lazy_imports_lib.lazy_imports.PIL_ImageDraw
--- a/tensorflow_datasets/core/features/bounding_boxes.py
+++ b/tensorflow_datasets/core/features/bounding_boxes.py
@@ -82,24 +82,6 @@ class BBoxFeature(feature.Tensor):
         [bbox.ymin, bbox.xmin, bbox.ymax, bbox.xmax]
     )
 
-  def _build_thumbnail_with_bbox(self, ex: np.ndarray):
-    """Returns blank image with Bboxes drawn on it."""
-    PIL_Image = lazy_imports_lib.lazy_imports.PIL_Image
-    PIL_ImageDraw = lazy_imports_lib.lazy_imports.PIL_ImageDraw
-
-    SIZE = 150
-    blank_img = PIL_Image.new('RGB', (SIZE, SIZE), (255, 255, 255))
-    draw = PIL_ImageDraw.Draw(blank_img)
-    rs = np.random.RandomState(97531)  # freeze random state
-
-    for i in range(ex.shape[0]):
-      # Rescale coordinates to match size of blank_image
-      ymin, xmin, ymax, xmax = ex[i, :]*SIZE
-      # Generate random rgb values for Bbox ouline
-      r, g, b = list(rs.randint(0, 256, size=3))
-      draw.rectangle(((xmin, ymin), (xmax, ymax)), outline=(r, g, b))
-    return blank_img
-
   def repr_html(self, ex: np.ndarray) -> str:
     """Returns the HTML str representation of an Image with BBoxes."""
     ex = np.expand_dims(ex, axis=0)  # Convert single bounding box to single batch.
@@ -111,7 +93,7 @@ class BBoxFeature(feature.Tensor):
 
   def _repr_html(self, ex: np.ndarray) -> str:
     """Returns the HTML str representation of an Image with BBoxes."""
-    img = self._build_thumbnail_with_bbox(ex)
+    img = _build_thumbnail_with_bbox(ex)
     img_str = utils.get_base64(lambda buff: img.save(buff, format='PNG'))
     return f'<img src="data:image/png;base64,{img_str}" alt="Img" />'
 
@@ -122,3 +104,21 @@ class BBoxFeature(feature.Tensor):
 
   def to_json_content(self) -> Json:
     return dict()
+
+def _build_thumbnail_with_bbox(ex: np.ndarray):
+  """Returns blank image with Bboxes drawn on it."""
+  PIL_Image = lazy_imports_lib.lazy_imports.PIL_Image
+  PIL_ImageDraw = lazy_imports_lib.lazy_imports.PIL_ImageDraw
+
+  SIZE = 150
+  blank_img = PIL_Image.new('RGB', (SIZE, SIZE), (255, 255, 255))
+  draw = PIL_ImageDraw.Draw(blank_img)
+  rs = np.random.RandomState(97531)  # freeze random state
+
+  for i in range(ex.shape[0]):
+    # Rescale coordinates to match size of blank_image
+    ymin, xmin, ymax, xmax = ex[i, :]*SIZE
+    # Generate random rgb values for Bbox ouline
+    r, g, b = list(rs.randint(0, 256, size=3))
+    draw.rectangle(((xmin, ymin), (xmax, ymax)), outline=(r, g, b))
+  return blank_img
--- a/tensorflow_datasets/core/features/bounding_boxes.py
+++ b/tensorflow_datasets/core/features/bounding_boxes.py
@@ -17,8 +17,11 @@
 
 import collections
 
+import numpy as np
 import tensorflow.compat.v2 as tf
 
+from tensorflow_datasets.core import utils
+from tensorflow_datasets.core import lazy_imports_lib
 from tensorflow_datasets.core.features import feature
 from tensorflow_datasets.core.utils import type_utils
 
@@ -78,6 +81,31 @@ class BBoxFeature(feature.Tensor):
     return super(BBoxFeature, self).encode_example(
         [bbox.ymin, bbox.xmin, bbox.ymax, bbox.xmax]
     )
+
+  def _build_thumbnail_with_bbox(self, ex: np.ndarray) -> str:
+    """Returns the HTML str representation of an Image with BBoxes."""
+    PIL_Image = lazy_imports_lib.lazy_imports.PIL_Image
+    PIL_ImageDraw = lazy_imports_lib.lazy_imports.PIL_ImageDraw
+
+    SIZE = 150
+    blank_img = PIL_Image.new('RGB', (SIZE, SIZE), (255, 255, 255))
+    draw = PIL_ImageDraw.Draw(blank_img)
+
+    for i in range(ex.shape[0]):
+      # Rescale coordinates to match size of blank_image
+      ymin, xmin, ymax, xmax = ex[i, :]*SIZE
+      # Generate random rgb values for Bbox ouline
+      r, g, b = list(np.random.randint(0, 256, size=3))
+      draw.rectangle(((xmin, ymin), (xmax, ymax)), outline=(r, g, b))
+    return blank_img
+
+  def repr_html_batch(self, ex: np.ndarray) -> str:
+    """Returns the HTML str representation of the object."""
+    img = self._build_thumbnail_with_bbox(ex)
+
+    img_str = utils.get_base64(lambda buff: img.save(buff, format='PNG'))
+
+    return f'<img src="data:image/png;base64,{img_str}" alt="Img" />'
 
   @classmethod
   def from_json_content(cls, value: Json) -> 'BBoxFeature':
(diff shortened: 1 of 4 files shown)


harsh020:
@Conchylicultor Hi, currently it is WIP. Kindly have a look and please let know if I am headed in the right direction.

Thank You.

harsh020:
BTW, currently I have added changes only for `Bboxes`. The result look something like this -

Screen Shot 2020-11-28 at 12 23 24 AM

Thank You.

Conchylicultor (approved):
Nice! Thank you for helping with this. As features are independents and self contained, I'll start by merging this first.

tensorflow_datasets/core/features/feature.py
```diff
@@ -444,12 +445,34 @@ def decode_ragged_example(self, tfexample_data):
...
+
+    return f'<img src="data:image/png;base64,{img_str}" alt="Img" />'
+
   def repr_html(self, ex: np.ndarray) -> str:
     """Returns the HTML str representation of the object."""
     return _repr_html(ex)
 
   def repr_html_batch(self, ex: np.ndarray) -> str:
     """Returns the HTML str representation of the object (Sequence)."""
+    if len(ex.shape) > 1 and ex.shape[1] == 4:
```
Conchylicultor, line 474: This should be implemented in https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/features/bounding_boxes.py instead.

Similarly to `Image.repr_html`:

https://github.com/tensorflow/datasets/blob/26587ec6d4b23a501210a687be667e2f8256f32a/tensorflow_datasets/core/features/image_feature.py#L164

Conchylicultor (changes requested):

harsh020:
@Conchylicultor Thank you for reviewing. I have placed the implementation in `BboxFeature`, but currently I'm overriding the `repr_html_batch` as our requirement specified "`BboxFeature` in `Sequence`". But if we want to display `Bboxes` in image even when they appear independently as  a feature I need to override `rear_html` also. So, please let me know if we need to do that too.

Thank you.

harsh020:
@Conchylicultor Hi, I was working on the video and `GIF` part and the result are as follows - 
* Convert to `mp4` if `ffmpeg` is installed.
* Decrease the `framerate`


output

harsh020:
@Conchylicultor While testing for different browsers, I noted that chrome does not support `mp4`. So videos are just shown as blank in chrome.

Thank you.

Conchylicultor:
> @Conchylicultor Thank you for reviewing. I have placed the implementation in `BboxFeature`, but currently I'm overriding the `repr_html_batch` as our requirement specified "`BboxFeature` in `Sequence`". But if we want to display `Bboxes` in image even when they appear as an independent feature I need to override `rear_html` also. So, please let me know if we need to do that too.
> 
> Thank you.

Yes, it would be nice to also support individual `BBoxFeature` too.

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +197,76 @@ def to_json_content(self) -> Json:
         'ffmpeg_extra_args': self._extra_ffmpeg_args
     }
 
+  def _generate_video_html(self, images, framerate=24, encoding_format='mp4') -> str:
+    """Converts sequence of images into video string."""
+    # CODE REDUNDANT: Could somehow use `_ffmpeg_decode`?
+    imgs = len(str(len(images)))+1  # Find number of digits in len to give names.
+    video_dir = tempfile.mkdtemp()
```
Conchylicultor, line 204: To auto-cleanup the code, could you use `with tempfile. TemporaryDirectory():` contextmanager ?

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +197,76 @@ def to_json_content(self) -> Json:
...
+    ffmpeg_args = [self._ffmpeg_path, '-framerate', str(framerate), '-i',
+                   os.fspath(os.path.join(video_dir, f'img%0{imgs}d.png'))]
+    ffmpeg_stdin = None
+
+    output_pattern = os.path.join(video_dir, 'output.' + encoding_format)
+    ffmpeg_args += self._extra_ffmpeg_args
+    ffmpeg_args.append(output_pattern)
+    try:
+      # try using ffmpeg to convert pngs to mp4
+      process = subprocess.Popen(ffmpeg_args,
```
Conchylicultor, line 218: Could you use the higher level `subprocess.run` instead ?

tensorflow_datasets/core/features/bounding_boxes.py
```diff
@@ -79,6 +82,38 @@ def encode_example(self, bbox):
...
+
+    SIZE = 150
+    blank_img = PIL_Image.new('RGB', (SIZE, SIZE), (255, 255, 255))
+    draw = PIL_ImageDraw.Draw(blank_img)
+
+    for i in range(ex.shape[0]):
+      # Rescale coordinates to match size of blank_image
+      ymin, xmin, ymax, xmax = ex[i, :]*SIZE
+      # Generate random rgb values for Bbox ouline
+      r, g, b = list(np.random.randint(0, 256, size=3))
```
Conchylicultor, line 98: Color should be deterministic (generating the image twice should yield the same result). You should use https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.random.RandomState.html

tensorflow_datasets/core/features/bounding_boxes.py
```diff
@@ -79,6 +82,38 @@ def encode_example(self, bbox):
         [bbox.ymin, bbox.xmin, bbox.ymax, bbox.xmax]
     )
 
+  def _build_thumbnail_with_bbox(self, ex: np.ndarray):
```
Conchylicultor, line 85: As `self` isn't used anywhere, I would make this a simple function rather than method

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +197,76 @@ def to_json_content(self) -> Json:
         'ffmpeg_extra_args': self._extra_ffmpeg_args
     }
 
+  def _generate_video_html(self, images, framerate=24, encoding_format='mp4') -> str:
+    """Converts sequence of images into video string."""
+    # CODE REDUNDANT: Could somehow use `_ffmpeg_decode`?
```
Conchylicultor, line 202: Yes :). You could refactor `_ffmpeg_decode` to use some common `_ffmpeg_run` wrapper.

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +197,76 @@ def to_json_content(self) -> Json:
...
+      def write_buff(buff):
+        buff.write(video)
+      video_str = utils.get_base64(write_buff)
+      return (f'<video height="128" width="128" controls loop plays-inline>'
+              f'<source src="data:image/gif;base64,{video_str}"'
+              f' type="video/mp4" alt="Video"></video>')
+
+    except OSError:
+      # if ffmpeg is not installed, generate GIF using pngs
+      duration = (41*24)/framerate
```
Conchylicultor, line 245: Shouldn' the framerate be the same for both gif and mp4 ?

Conchylicultor (changes requested):
Thank you for the update.

For the chrome issue. Could you look at:
* https://stackoverflow.com/questions/17379971/html5-video-chrome-ffmpeg-mp4-working-in-all-but-chrome
* https://stackoverflow.com/questions/25131906/mp4-file-not-playing-in-chrome
Or similar ?

harsh020:
> * https://stackoverflow.com/questions/25131906/mp4-file-not-playing-in-chrome

* I tried using `h264` video codec, but it is unable to compress images of size `128x128`. But works fine with bigger images.
* I tried using `libx264` but it is generating `Segmentation Fault`.

I am searching for a fix. If you happen to know how to fix it, please do let me know.

Thank you.

harsh020:
@Conchylicultor After a lot of work I finally managed to get a fix for chrome support. I was not able to set `libx264`, moreover adding it would further add `libx264` dependency on `ffmpeg` (i.e, `ffmpeg` must be configured with `libx264`). So I avoided it. I have tested it on chrome, it works just fine.
Kindly, have a look and let me know. 

Thank you.

tensorflow_datasets/core/features/video_feature.py
harsh020: I made this relation to find duration from given framerate (according to the mentioned comment i.e, 41ms ~= 24 imgs/s). As `PIL.Image` uses duration for creating `GIF`.

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -130,19 +130,7 @@ def _ffmpeg_decode(self, path_or_fobj):
     ffmpeg_args += self._extra_ffmpeg_args
     ffmpeg_args.append(output_pattern)
     try:
```
Conchylicultor, line 132: The `try` / `except` should be moved inside `_run_ffmpeg` too

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
...
+
+    with tempfile.TemporaryDirectory() as video_dir:
+      for i, img in enumerate(images):
+        f = os.path.join(video_dir, f'img{i:0{imgs}d}.png')
+        img.save(f, format='png')
+
+      ffmpeg_args = [self._ffmpeg_path, '-framerate', str(framerate), '-i',
+                    os.fspath(os.path.join(video_dir, f'img%0{imgs}d.png'))]
+
+      extra_ffmpeg_args = ['-vf', 'scale=128:128', '-vcodec', 'h264',
```
Conchylicultor, line 200: Could you document what each of those argument represent ?

Conchylicultor (changes requested):

tensorflow_datasets/core/features/bounding_boxes.py
```diff
@@ -79,10 +82,43 @@ def encode_example(self, bbox):
...
+  def repr_html(self, ex: np.ndarray) -> str:
+    """Returns the HTML str representation of an Image with BBoxes."""
+    ex = np.expand_dims(ex, axis=0)  # Convert single bounding box to single batch.
+    return self._repr_html(ex)
+
+  def repr_html_batch(self, ex: np.ndarray) -> str:
+    """Returns the HTML str representation of an Image with BBoxes (Sequence)."""
+    return self._repr_html(ex)
+
+  def _repr_html(self, ex: np.ndarray) -> str:
```
Conchylicultor, line 94: If `self` isn't used anywhere, this should be a simple function. This helps readability by explicitly showing there is no hidden magic/`self` input.

tensorflow_datasets/core/features/video_feature.py
Conchylicultor: And rather than the `finally:`, you could replace `tempfile.mkdtemp()` by `with tempfile.TemporaryDirectory()` context manager.

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
         'ffmpeg_extra_args': self._extra_ffmpeg_args
     }
 
+  def _generate_video_html(self, images, framerate=24, encoding_format='mp4') -> str:
+    """Converts sequence of images into video string."""
+    imgs = len(str(len(images)))+1  # Find number of digits in len to give names.
```
Conchylicultor, line 190: `imgs` -> `nb_digit`

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
...
+
+    with tempfile.TemporaryDirectory() as video_dir:
```
Conchylicultor, line 192: Is there a way to test that `ffmpeg` is installed before writing on disk ? (e.g. run a dummy `ffmpeg --version` command) As this isn't required for GIF generation.

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
...
+        # if ffmpeg is not installed, generate GIF using pngs
+        duration = (41*24)/framerate
+        def write_buff(buff):
+          images[0].save(
+              buff,
+              format='GIF',
+              save_all=True,
+              append_images=images[1:],
+              # Could add a frame_rate kwargs in __init__ to customize this.
+              duration=duration,  # 41ms / img ~= 24 img / sec
```
Conchylicultor, line 232: Originally we picked 24 img/sec but this is too fast. Now the two FFMPEG and GIF framerate should match

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
...
+                           '-pix_fmt', 'yuv420p', '-allow_sw', '1']
+      ffmpeg_stdin = None
+
+      output_pattern = os.path.join(video_dir, 'output.' + encoding_format)
+      ffmpeg_args += extra_ffmpeg_args
+      ffmpeg_args.append(output_pattern)
+      try:
+        _ffmpeg_run(ffmpeg_args, ffmpeg_stdin)
+
+        video = None
```
Conchylicultor, line 210: rm

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
...
+      for i, img in enumerate(images):
+        f = os.path.join(video_dir, f'img{i:0{imgs}d}.png')
+        img.save(f, format='png')
+
+      ffmpeg_args = [self._ffmpeg_path, '-framerate', str(framerate), '-i',
+                    os.fspath(os.path.join(video_dir, f'img%0{imgs}d.png'))]
+
+      extra_ffmpeg_args = ['-vf', 'scale=128:128', '-vcodec', 'h264',
+                           '-pix_fmt', 'yuv420p', '-allow_sw', '1']
+      ffmpeg_stdin = None
```
Conchylicultor, line 202: Rm and set

```
_ffmpeg_run(ffmpeg_args, ffmpeg_stdin=None)
```

instead

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
...
+
+      output_pattern = os.path.join(video_dir, 'output.' + encoding_format)
```
Conchylicultor, line 204: Rm `encoding_format` and set `.mp4` as this is hardcoded in multiple places

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -197,26 +185,80 @@ def to_json_content(self) -> Json:
...
+        video = None
+        video_file = os.path.join(video_dir, 'output.mp4')
+        with tf.io.gfile.GFile(video_file, 'rb') as vf:
+          video = vf.read()
+
+        def write_buff(buff):
+          buff.write(video)
+        video_str = utils.get_base64(write_buff)
+        return (f'<video height="128" width="128" controls loop plays-inline>'
+                f'<source src="data:image/gif;base64,{video_str}"'
```
Conchylicultor, line 219: Why is source `data:image/gif` ? Shouldn't this be video ? Is it why HTML5 was bugging ?

Conchylicultor (changes requested):

tensorflow_datasets/core/features/video_feature.py
harsh020: Oh! Its a typo I made, thanks. No, I don't think its the reason for bugging the chrome support. It was most probably because chrome didn't support that codec.

tensorflow_datasets/core/features/video_feature.py
harsh020: I have added a `framerate` parameter to the and am generating duration from that to keep comparable framerate between the two.

Sure now I will use the original relation to convert `framerate` (used by `mp4`) into `duration` in ms (used by `GIF`). 
```
duration = (1/framerate)*100   #framerate will be passed to the function, currently I'm using 10
```

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -185,27 +177,53 @@ def to_json_content(self) -> Json:
...
+        ffmpeg_args = [self._ffmpeg_path, '-framerate', str(framerate), '-i',
+                      os.fspath(os.path.join(video_dir, f'img%0{num_digits}d.png'))]
 
-      output_pattern = os.path.join(video_dir, 'output.' + encoding_format)
-      ffmpeg_args += extra_ffmpeg_args
-      ffmpeg_args.append(output_pattern)
-      try:
-        _ffmpeg_run(ffmpeg_args, ffmpeg_stdin)
+        extra_ffmpeg_args = [
+          '-vf', 'scale=128:128', # scale output videos to 128x128
```
Conchylicultor, line 215: Isn't the images already scaled to 128 ?

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -185,27 +177,53 @@ def to_json_content(self) -> Json:
...
+          # using native h264 encoder
+          '-vcodec', 'h264',      # output video stream codec - H.264
+          '-pix_fmt', 'yuv420p',  # use yuv420v pixel format (enhance compatibility)
+          '-allow_sw', '1'        # allow software encoding
```
Conchylicultor, line 219: Could you provide some context ? Why is it required ?

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -185,27 +177,53 @@ def to_json_content(self) -> Json:
...
-      output_pattern = os.path.join(video_dir, 'output.' + encoding_format)
-      ffmpeg_args += extra_ffmpeg_args
-      ffmpeg_args.append(output_pattern)
-      try:
-        _ffmpeg_run(ffmpeg_args, ffmpeg_stdin)
+        extra_ffmpeg_args = [
+          '-vf', 'scale=128:128', # scale output videos to 128x128
+          # using native h264 encoder
+          '-vcodec', 'h264',      # output video stream codec - H.264
+          '-pix_fmt', 'yuv420p',  # use yuv420v pixel format (enhance compatibility)
```
Conchylicultor, line 218: enhance compatibility with what ?

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -185,27 +177,53 @@ def to_json_content(self) -> Json:
...
+          '-allow_sw', '1'        # allow software encoding
+        ]
+
+        output_pattern = os.path.join(video_dir, 'output.mp4')
+        ffmpeg_args += extra_ffmpeg_args
+        ffmpeg_args.append(output_pattern)
+
+        _ffmpeg_run(ffmpeg_args)
```
Conchylicultor, line 226: The `video = None` is overwritten be video bellow, so I would remove the line

Conchylicultor (changes requested):
Thank you for the update

tensorflow_datasets/core/features/video_feature.py
```diff
@@ -216,27 +234,9 @@ def write_buff(buff):
           buff.write(video)
         video_str = utils.get_base64(write_buff)
         return (f'<video height="128" width="128" controls loop plays-inline>'
-                f'<source src="data:image/gif;base64,{video_str}"'
+                f'<source src="data:image/video;base64,{video_str}"'
```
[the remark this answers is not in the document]
Conchylicultor: This correspond to the MIME type to indicate to the browser the format of the data: See https://www.iana.org/assignments/media-types/media-types.xhtml#video
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

tensorflow_datasets/core/features/video_feature.py
Conchylicultor: For reference, which codec wasn't supported ?

Conchylicultor: Are you sure it wasn't the MIME ? From https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

> **Important:** Browsers use the MIME type, not the file extension, to determine how to process a URL, so it's important that web servers send the correct MIME type in the response's Content-Type header. If this is not correctly configured, browsers are likely to misinterpret the contents of files and sites will not work correctly, and downloaded files may be mishandled.

tensorflow_datasets/core/features/video_feature.py
harsh020: When using H.264 coded video, it ensures that not so good players can decode the video. Please see:
https://trac.ffmpeg.org/wiki/Slideshow

Conchylicultor: Thank you. Not everyone is familiar with FFMPEG, so I think we should make sure comments are explainatory. For example. we should add https://trac.ffmpeg.org/wiki/Slideshow to the comment

tensorflow_datasets/core/features/video_feature.py
harsh020: >Are you sure it wasn't the MIME ? From https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

Yes, I did correct the MIME type, disabled the extra args and rechecked, it did not work, so pretty much sure that the codec wasn't supported.

> For reference, which codec wasn't supported ?

`mpeg` Please refer: https://stackoverflow.com/questions/17793150/default-ffmpeg-codec-when-nothing-is-specified

Conchylicultor (approved):
Thank you very much for the explanations and documentation! The code looks much better now.

harsh020:
@Conchylicultor Hi, sorry for the delayed response. Actually was looking into the `codec` thing. The official documents  [here](https://ffmpeg.org), do not mention anything about `h264` encoder that we are using. So after googling it I found [this](https://video.stackexchange.com/questions/28294/h264-constrained-baseline-h264-native-vs-h264-libx264), it mentions that native `h264` is used for detecting the input video stream as `H.264` and decode it.
But experimentally I found out that the native `h264` is also converting the format to `H.264`. This is confusing.

BTW, I had also tried using `libx264` before finally using `h264` (prior to committing), but because - 
* Using `libx264` gave `Segmentation Fault` (may be I might not have been able to set it correctly)
* Using `libx264` would create an extra dependency that `ffmpeg` be complied with `x264` package,
so, I avoided using it.

Kindly, suggest me how to proceed.

Thank you.

Conchylicultor:
Our documentation has been updated correctly with video, so it seems to work now: https://www.tensorflow.org/datasets/catalog/ucf101

Unless we receive bugs from user, I think the current code is good. Thanks again for contributing

Issue #2813, opened 2020-12-02
Update documentation to display dataset module


Conchylicultor:
With community datasets coming in, TFDS will automatically document datasets generated outside of TFDS, but publicly available (e.g. in GCS).
We would like to point to the original source code which generated the dataset. Like the current "Source code" section: (e.g. https://www.tensorflow.org/datasets/catalog/accentdb).

We should save the `builber_cls.__module__` inside `dataset_info.json`. This require adding a new `module` field in dataset_info.proto (https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/core/proto)

Once community datasets are available, we could read the original module location and point to it in our documentation.

NikhilBartwal:
@Conchylicultor I would like to take this up. Could you guide me how should i get started with `dataset_info.proto`. Thanks!

Conchylicultor:
@NikhilBartwal Thank you for your interest.
You should:
* have a look at `dataset_info.proto` and add a new string `module` field.
* Regenerate the dataset_info_generated_pb2.py with instruction from https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/core/proto
* Then you should fill this field in the `dataset_info.py`
* Then add test in `dataset_info_test.py` to make sure this field is properly set and restored.

vijayphoenix:
Closed by https://github.com/tensorflow/datasets/pull/2831

PR #2831, opened 2020-12-05
Add Dataset Module field to DatasetInfo

NikhilBartwal:
In reference to Issue #2813 
As suggested by @Conchylicultor, a new `module_field` was added to the `dataset_info` to hold the original module location of the dataset.
The added change was also tested by updating the `INFO_STR` in `dataset_info_test` and then comparing it with the new ly generated `dataset_info`. 

The `mnist dataset_info.json` file used for testing was also updated manually to contain module_name so as it can be tested for `assertEqual` with the newly generated one.
The code performed as expected without generating any failures (except some which were present previous to the changes)


NikhilBartwal:
@Conchylicultor Could you have a look if this is indeed what was needed? Thanks!

Conchylicultor (approved):
Thank you. Looks good

NikhilBartwal:
@Conchylicultor There are a couple more `AssertionErrors` that occur when `dataset_info_test.py` is ran. Should i open separate issues for that or should i wait for this to get merged first?

Conchylicultor:
`test_utils.py` should not be executed as test: https://github.com/tensorflow/datasets/blob/8ee49991924a058b91b410f3bc4309f3a764faa2/oss_scripts/oss_tests.sh#L45

I've imported the code internally, so no need to fix anything from your side!

NikhilBartwal:
@Conchylicultor I ran `dataset_info_test.py` to test whether it is generated correctly (the same script was used to test this PR as well). `test_utils.py` was not run directly but was a part of `dataset_into_test.py`.

This was the test which resulted in the mentioned assertion errors. 
The mentioned code was only a part of the stack trace.

tensorflow_datasets/core/proto/dataset_info_generated_pb2.py
```diff
@@ -37,7 +35,8 @@
...
+  create_key=_descriptor._internal_create_key,
```
Conchylicultor, line 38: Could you share which protobuf version you used to generate the `_pb2.py` ? I think the new version might break users which are still using an older protobuf package version: https://stackoverflow.com/questions/61922334/how-to-solve-attributeerror-module-google-protobuf-descriptor-has-no-attribu

Could you try to generate the proto with an older protobuf version which avoid the `_internal_create_key` ?

Conchylicultor (changes requested):

tensorflow_datasets/core/proto/dataset_info_generated_pb2.py
NikhilBartwal: @Conchylicultor I used the v3.15 by custom download and setup as the v3.11 provided for download as part of the `install_protoc.sh` was not compatible for windows.

I will revert back to an older version (preferably 3.11) and try generating the proto again!

Conchylicultor:
> @Conchylicultor I ran `dataset_info_test.py` to test whether it is generated correctly (the same script was used to test this PR as well). `test_utils.py` was not run directly but was a part of `dataset_into_test.py`.
> 
> This was the test which resulted in the mentioned assertion errors.
> The mentioned code was only a part of the stack trace.

For your issues, I believe this is because you haven't installed required extra-tests deps like `tensorflow-data-validation`. See https://www.tensorflow.org/datasets/contribute#cloning_the_repo

NikhilBartwal:
@Conchylicultor The `_internal_create_key` issue was fixed by reverting back to an older `v3.11.2` and regenerating the proto. The required tests passed successfully.

NikhilBartwal:
> 
> For your issues, I believe this is because you haven't installed required extra-tests deps like `tensorflow-data-validation`. See https://www.tensorflow.org/datasets/contribute#cloning_the_repo
> 
> ```
> pip install -e .[dev]
> ```
@Conchylicultor My bad, i didn't know about that. Thanks for clearing that out!

Conchylicultor (approved):
Awesome! Merging this soon

vijayphoenix:
Merged by https://github.com/tensorflow/datasets/commit/0a1875c7f43a48932225dca707c1decd7031c7b3

PR #2838, opened 2020-12-08
Add support for decoders in ImageFolder

mkuchnik:
This PR adds decoder support to ImageFolder. Currently, ImageFolder decodes all files in the ImageFolder, which prevents accessing the raw file bytes to do custom decoding (e.g., fused crop+decode). The PR pushes decoding logic into a decode_function, which can then be set to a default (if no decoders are passed) or set to utilize the passed decoders.

I did not see a convenient way to add a test case (given the `mock` usage). However, the current tests run and you can verify that the code functions with the following:

```python3
import tensorflow as tf
import numpy as np

data_dir = "my_dataset"
builder = tfds.ImageFolder(data_dir)
decoders = {"image": tfds.decode.SkipDecoding(),
            }
ds1 = builder.as_dataset(split="train", decoders=None, as_supervised=True)
ds2 = builder.as_dataset(split="train", decoders=decoders, as_supervised=True)
ds2 = ds2.map(lambda x, y: (tf.image.decode_jpeg(x), y))

for x1, x2 in zip(ds1, ds2):
    assert np.all(x1[0] == x2[0])
    assert x1[1] == x2[1]
```

### DIFF
--- a/tensorflow_datasets/core/folder_dataset/image_folder.py
+++ b/tensorflow_datasets/core/folder_dataset/image_folder.py
@@ -16,6 +16,7 @@
 """Image Classification Folder datasets."""
 
 import collections
+import functools
 import os
 import random
 from typing import Dict, List, NoReturn, Optional, Tuple
@@ -23,6 +24,7 @@ from typing import Dict, List, NoReturn, Optional, Tuple
 import tensorflow.compat.v2 as tf
 from tensorflow_datasets.core import dataset_builder
 from tensorflow_datasets.core import dataset_info
+from tensorflow_datasets.core import decode
 from tensorflow_datasets.core import features as features_lib
 from tensorflow_datasets.core import splits as split_lib
 from tensorflow_datasets.core.utils import type_utils
@@ -135,15 +137,21 @@ class ImageFolder(dataset_builder.DatasetBuilder):
 
   def _as_dataset(
       self,
-      split,
-      shuffle_files=False,
-      decoders=None,
+      split: str,
+      shuffle_files: bool = False,
+      decoders: Optional[Dict[str, decode.Decoder]] = None,
       read_config=None) -> tf.data.Dataset:
     """Generate dataset for given split."""
     del read_config  # Unused (automatically created in `DatasetBuilder`)
+
     if decoders:
-      raise NotImplementedError(
-          '`decoders` is not supported with {}'.format(type(self).__name__))
+        unrecognized_features = (set(decoders.keys())
+                                 .difference(set(self.info.features.keys())))
+        if unrecognized_features:
+          raise ValueError(
+              'Unrecognized features {}. '
+              'Decoder feature name should be one of {}.'.format(
+                  unrecognized_features, list(self.info.features.keys())))
     if split not in self.info.splits.keys():
       raise ValueError(
           'Unrecognized split {}. Subsplit API not yet supported for {}. '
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/folder_dataset/image_folder.py
```diff
@@ -135,15 +137,21 @@ def download_and_prepare(self, **kwargs):  # -> NoReturn:
...
     del read_config  # Unused (automatically created in `DatasetBuilder`)
+
     if decoders:
-      raise NotImplementedError(
-          '`decoders` is not supported with {}'.format(type(self).__name__))
+        unrecognized_features = (set(decoders.keys())
+                                 .difference(set(self.info.features.keys())))
+        if unrecognized_features:
+          raise ValueError(
+              'Unrecognized features {}. '
```
Conchylicultor, line 152: f-string

tensorflow_datasets/core/folder_dataset/image_folder.py
```diff
@@ -163,20 +171,33 @@ def _as_dataset(
     if shuffle_files:
       ds = ds.shuffle(len(examples))
     ds = ds.map(_load_example, num_parallel_calls=tf.data.experimental.AUTOTUNE)
+    if not decoders:
+      decode_fn = _default_decode_fn
+    else:
+      decode_fn = functools.partial(
+          self.info.features.decode_example, decoders=decoders)
+
+    ds = ds.map(decode_fn, num_parallel_calls=tf.data.experimental.AUTOTUNE)
```
Conchylicultor, line 180: Rather than calling `ds.map` twice. How about moving this logic inside `_load_example` ?

```python
def _load_example(...) -> Dict[str, tf.Tensor]:
  img = tf.io.read_file(path)
  ex = {
      'image': img,
      'label': tf.cast(label, tf.int64),
      'image/filename': path,
  }
  return features.decode_example(ex, decoder=decoder)
```

tensorflow_datasets/core/folder_dataset/image_folder.py
```diff
@@ -135,15 +137,21 @@ def download_and_prepare(self, **kwargs):  # -> NoReturn:
...
+      shuffle_files: bool = False,
+      decoders: Optional[Dict[str, decode.Decoder]] = None,
       read_config=None) -> tf.data.Dataset:
     """Generate dataset for given split."""
     del read_config  # Unused (automatically created in `DatasetBuilder`)
+
     if decoders:
-      raise NotImplementedError(
-          '`decoders` is not supported with {}'.format(type(self).__name__))
+        unrecognized_features = (set(decoders.keys())
```
Conchylicultor, line 148: If decoder is invalid, the error should already be raised in `self.info.features.decode_example(..., decoders=decoders)` no ? 
If the error message isn't explicit enough, we should update `features.decode_example` instead. I think it's best to centralize the place where error is raised so error is consistent across datasets.

tensorflow_datasets/core/folder_dataset/image_folder.py
```diff
@@ -163,20 +171,33 @@ def _as_dataset(
     if shuffle_files:
       ds = ds.shuffle(len(examples))
     ds = ds.map(_load_example, num_parallel_calls=tf.data.experimental.AUTOTUNE)
+    if not decoders:
+      decode_fn = _default_decode_fn
+    else:
+      decode_fn = functools.partial(
```
Conchylicultor, line 177: Would there be a drawback about always decoding through `self.info.features.decode_example` even when `decoders` is `None` ?

Conchylicultor (changes requested):
Thank you! This looks good. Could you also add a test to make sure this works as intended?

tensorflow_datasets/core/folder_dataset/image_folder.py
mkuchnik: `decode_example` in tensorflow_datasets/core/features/image_feature.py does expand_animation=True and does not hardcode channels=3, which is different from the current behavior as far as I can tell.

tensorflow_datasets/core/folder_dataset/image_folder.py
mkuchnik: A lambda function seems to be a better way to fix this. The type signature of _load_example becomes bloated with `Callable` and lambda is 1-line of code.

tensorflow_datasets/core/folder_dataset/image_folder.py
mkuchnik: You are right. Removing this code.

tensorflow_datasets/core/folder_dataset/image_folder.py
mkuchnik: Code should be removed.

mkuchnik:
@Conchylicultor I have added tests, which test the interface.

I am most concerned about what the default behavior should be when no decoder is passed. As noted above, it seems to me that tfds default decoding behavior is to use variable amounts of channels and expand_animations=True. This may be desirable behavior as a default. Users can pass channels=3 and expand_animations=False in a custom decoder.

Conchylicultor (approved):
Thank you for the update. Looks good.

Concerning the default shape, removing explicit channel number makes `tf.decode.image` return a `(None, None, None)` `tf.Tensor`. This makes downstream model fails as most models expect a fixed number of channel. `3` is the most common default which is why we use it.

PR #2867, opened 2020-12-14
Remove deprecated dataset from doc (Fix #2861)

### LINKED ISSUE #2861: `tfds.image_classification.image_folder.ImageLabelFolder` has been deleted
This PR removed a `class` I use in my codebase: https://github.com/tensorflow/datasets/pull/2089 c47aa6ecd0dd62e114102286fec478af1330107f

The `class` is documented here: https://www.tensorflow.org/datasets/catalog/image_label_folder - but the link breaks (because the file was deleted in that PR).

If this isn't a mistake, then dropping it from the documentation would be a good idea. If this is a mistake, then adding the file back would be prudent.

_In the meantime I'll copy the file into my codebase_

EDIT: Found it - https://www.tensorflow.org/datasets/api_docs/python/tfds/folder_dataset/ImageFolder https://github.com/tensorflow/datasets/blob/v4.1.0/tensorflow_datasets/core/folder_dataset/image_folder.py#L40-L163

tfds-copybara:
Remove deprecated dataset from doc (Fix #2861)


vijayphoenix:
Fixed

Issue #1441, opened 2020-02-01
dataset_builder._prepare_split ResourceExhaustedError


PerryXDeng:
**Short description**
[dataset_builder._prepare_split](https://github.com/tensorflow/datasets/blob/e1c9535c9d9b67e99ee962ee0ac4ce14dfe89a2b/tensorflow_datasets/core/dataset_builder.py#L1047) throws `tensorflow.python.framework.errors_impl.ResourceExhaustedError; Too many open files` when used on imagenet_resized 64x64. 

**Environment information**
* Operating System: linux with user privileges, so not possible to change the Python upperbound for number of open files
* Python version:3.6.7
* tensorflow-datasets version: 1.3.2, but the same function exists in 2.0.0 so I suspect the error is also there
* tensorflow-gpu version: 1.15

**Reproduction instructions**
```
import tensorflow_datasets as tfds
data = tfds.load(name="imagenet_resized", split="train", builder_kwargs={'config':'64x64'})
```

**Link to logs**
```Traceback (most recent call last):
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/file_format_adapter.py", line 199, in incomplete_dir
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/dataset_builder.py", line 322, in download_and_prepare
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/dataset_builder.py", line 969, in _download_and_prepare
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/dataset_builder.py", line 837, in _download_and_prepare
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/dataset_builder.py", line 996, in _prepare_split
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/tfrecords_writer.py", line 160, in write
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/shuffle.py", line 194, in add
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/shuffle.py", line 180, in _add_to_mem_buffer
  File "/home/pxd256/.local/lib/python3.6/site-packages/tensorflow_datasets/core/shuffle.py", line 174, in _add_to_bucket
[...]

PR #2890, opened 2020-12-22
Bug fix for Issue 1441

PerryXDeng:
IO parameters in shuffler implementation are modified to be dynamic and customizable rather than constants, which were causing dataset to crash on load for some environments. This Shuffler uses either 

1. 1000,  or
2. half of the process limit on number of open files

, which ever number is smaller, for number of buckets, by default. Related tests are also modified to work with the modifications. 

Fix: https://github.com/tensorflow/datasets/issues/1441

### DIFF
--- a/tensorflow_datasets/core/shuffle.py
+++ b/tensorflow_datasets/core/shuffle.py
@@ -17,6 +17,7 @@
 
 import math
 import os
+import resource
 import struct
 import uuid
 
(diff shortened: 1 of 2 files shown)


NikhilBartwal (approved):

PerryXDeng:
@Conchylicultor please review, thanks

fineguy:
Thank you @PerryXDeng for your contribution!

Sorry that we never got to your PR, but hopefully https://github.com/tensorflow/datasets/commit/8f44c895c4115ac75d4c2ca4847ebf441e5450d4 resolves the issue for you.

PR #2904, opened 2021-01-02
Update URL and version for citrus leaves dataset

NikhilBartwal:
In reference to issue Fix #2903
The website URL used for dataset generation was updated and the appropriate `version` and `release_notes` were added in the script. 
There was no change in the generated checksums.
The dataset is now generated and built successfully without any errors.

PR #2906, opened 2021-01-04
Recording checksums according to dataset structure

### LINKED ISSUE #2905: Modify existing checksum file
**Short description**
Changing the download URL should modify the existing checksum file of a dataset.

For example:
Changing the download URL of the `citrus` dataset should modify `url_checksums/citrus_leaves.txt`.
However, `--register_checksum` flag in `tfds build` creates a new checksum file in `image_classification/checksums.tsv`

**Environment information**
* Operating System: Ubuntu 20.04
* Python version: 3.8
* `tensorflow-datasets`/`tfds-nightly` version: tfds-nightly

**Reproduction instructions**

```
tfds build --dataset citrus_leaves --register_checksums
```

**Expected behavior**
 `url_checksums/citrus_leaves.txt` should be modified

NikhilBartwal:
In reference to issue Fix #2905 
The checksums are now recorded in the dataset folder for the new structure and in the common `url_checksums` folder for other datasets.
It was successfuly tested with `citrus_leaves` and `wine_quality` dataset, storing the checksums at the right place.
[Note: `Citrus leaves` datasets currently has the wrong `url`, which was fixed in #2904 ]

### DIFF
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -289,10 +289,7 @@ class DatasetBuilder(registered.RegisteredDataset):
       return cls.code_path.parent / "checksums.tsv"
     else:
       #Record checksums in the common folder for older structured datasets
-      chksm_path = utils.gpath.WindowsGPath(
-        'f:/tfds/datasets/tensorflow_datasets/url_checksums/' +
-        cls.name +
-        '.txt')
+      chksm_path = utils.tfds_path()/ 'url_checksums'/ (cls.name + '.txt')
       return chksm_path
 
   @utils.classproperty
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -285,8 +285,7 @@ class DatasetBuilder(registered.RegisteredDataset):
     # * To load the checksums (in url_infos)
     # * To save the checksums (in DownloadManager)
     #Checking whether the dataset follow the folder structure
-    path = str(cls.code_path.parent)
-    if path.split('\\')[-1] == cls.name:
+    if cls.code_path.parent.name == cls.name:
       return cls.code_path.parent / "checksums.tsv"
     else:
       #Record checksums in the common folder for older structured datasets
--- a/tensorflow_datasets/core/dataset_builder.py
+++ b/tensorflow_datasets/core/dataset_builder.py
@@ -289,6 +289,7 @@ class DatasetBuilder(registered.RegisteredDataset):
     if path.split('\\')[-1] == cls.name:
       return cls.code_path.parent / "checksums.tsv"
     else:
+      #Record checksums in the common folder for older structured datasets
       chksm_path = utils.gpath.WindowsGPath(
         'f:/tfds/datasets/tensorflow_datasets/url_checksums/' +
         cls.name +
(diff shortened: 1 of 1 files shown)


NikhilBartwal:
@vijayphoenix It would be great if you could have a look! Thanks.

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -284,7 +284,17 @@ def _checksums_path(cls) -> ReadOnlyPath:
     # Used:
     # * To load the checksums (in url_infos)
     # * To save the checksums (in DownloadManager)
-    return cls.code_path.parent / "checksums.tsv"
+    #Checking whether the dataset follow the folder structure
+    path = str(cls.code_path.parent)
+    if path.split('\\')[-1] == cls.name:
```
vijayphoenix, line 289: Try to work directly using Pathlib API

Like `cls.code_path.parent.name == cls.name`

tensorflow_datasets/core/dataset_builder.py
```diff
@@ -284,7 +284,17 @@ def _checksums_path(cls) -> ReadOnlyPath:
...
+      return cls.code_path.parent / "checksums.tsv"
+    else:
+      #Record checksums in the common folder for older structured datasets
+      chksm_path = utils.gpath.WindowsGPath(
+        'f:/tfds/datasets/tensorflow_datasets/url_checksums/' +
+        cls.name +
+        '.txt')
+      return chksm_path
```
vijayphoenix, line 297: - Don't hardcode any value
- Code should be independent of OS

Try rewriting as:
```
utils.tfds_path() / 'url_checksums' / (cls.name + '.txt')
```

vijayphoenix (changes requested):
I think this code should be moved to `_make_download_manager`.

vijayphoenix:
> In reference to issue #2905

As this PR completely fixes the issue, we can link the PR and issue. 
https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

NikhilBartwal:
Thank you for the review @vijayphoenix! I have made the necessary changes.

> Maybe this code should be moved to `_make_download_manager`.

The `_make_download_manager` function uses the `_checksums_path` class property to get the path, so i think it would be better to keep the code inside the `_checksums_path` as its whole purpose is to return the correct path only. What do you think?

> > In reference to issue #2905
> 
> As this PR completely fixes the issue, we can link the PR and issue.
> https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue
> 
> So, the issue will be automatically closed when the PR is merged

I currently, do not have write access to TFDS :(
It would be great if you could provide me that!

vijayphoenix:
> I currently, do not have write access to TFDS :(
> It would be great if you could provide me that!

Access is not required

> In reference to issue #2905

Just edit this to `Fix #2905`

Also, can you add tests for the same?
It will be easier for maintainers to verify the correctness...

Conchylicultor (approved):
Thank you!!

NikhilBartwal:
@Conchylicultor Do you think we need to add some tests or the existing changes will do?

Conchylicultor:
As this is a deprecated feature which should not be used anymore, the current code should be enough

PR #2921, opened 2021-01-06
Fix doc for pip install command for zsh

jatin-code777:
The following command doesn't work on zsh (the z shell) and shows this
```
$ pip install -e .[dev]              
zsh: no matches found: .[dev]
```

This happens because `[` and `]` hold meaning and are special 
characters in zsh and thus need to be either escaped or quoted. 

I have tested that `pip install -e ".[dev]"` works on bash, 
zsh and csh. I believe that this will work with most other shells too.

### DIFF
--- a/docs/contribute.md
+++ b/docs/contribute.md
@@ -45,10 +45,10 @@ Install the development dependencies:
 
 ```sh
 pip install -e .  # Install minimal deps to use tensorflow_datasets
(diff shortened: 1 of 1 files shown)


vijayphoenix (approved):

Conchylicultor (approved):
Thank you, looks good

Issue #2967, opened 2021-01-20
Better `tfds.as_dataframe` visualization for segmentation masks


Conchylicultor:
Currently datasets with segmentation mask are poorly displayed. For example, click "Display examples..." in https://www.tensorflow.org/datasets/catalog/lost_and_found. `instance_id` and `segmentation_label` are displayed as black images. Indeed image values are between `[0, ~5]` while the image is displayed as `[0, 255]`, so looks almost black.

We should display the segmentation as a color image. Each value of the original img would be mapped to a random RGB colour (e.g. 0 displayed as blue, 1 displayed as red,....). The mapping should be deterministic, so colour are consistent between images.

For the implementation:

1. Add a `use_colormap: bool` kwarg to `tfds.features.Image`

   ```python
   tfds.features.Image(..., use_colormap=True),
   ```

1. Update `tfds.features.Image.repr_html` such as if use_colormap is True, the 1 channel input image is converted to 3-channel RGB image using the mapping.

1. Add the `use_colormap=True` on the affected datasets.

PR #2974, opened 2021-01-21
added option to use_colormaps

DumbMachine:
Aim to provide a fix for https://github.com/tensorflow/datasets/issues/2967.

Makes use of the colors available in `matplotlib`.

### DIFF
--- a/tensorflow_datasets/core/features/image_feature.py
+++ b/tensorflow_datasets/core/features/image_feature.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 """Image feature."""
 
 import os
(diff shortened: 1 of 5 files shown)


DumbMachine:
@googlebot I consent.

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -189,14 +209,36 @@ def to_json_content(self) -> Json:
...
-    ex = ex.squeeze(axis=-1)
+    if use_colormap:
+      matplotlib = lazy_imports_lib.lazy_imports.matplotlib  # pylint: disable=invalid-name
+      mcolors = matplotlib.colors
+
+      ex = ex.squeeze(axis=-1)
+      unique_pixels = np.unique(ex)
+      blank_image = np.ones(shape=(*ex.shape[:2], 3), dtype=np.uint8)
+
+      if len(unique_pixels) <= 7:
```
Conchylicultor, line 227: If in image1 there is 5 values, but in image 2 there is 10, then colors won't match between images.

We should make sure colors are consistents so that `32` is always matched with the same color, irrespectively of the image. And independently if `np.unique(ex) == [0, 32]`, `np.unique(ex) == [0, 2, 3, 4 32]`,...

DumbMachine:
Closed in favor of #2975

Conchylicultor (changes requested):
Thank you for starting this. It's a good start

PR #2975, opened 2021-01-21
added option to use_colormaps

DumbMachine:
Aim to provide a fix for #2967.

Maps the pixels to another color space. Example from `lost_and_found` data:
image


Conchylicultor (changes requested):
Thank you! Please see my comments at: https://github.com/tensorflow/datasets/pull/2974

DumbMachine:
Modified the color logic, to choose a (deterministic) index from a pallete of 24 colors: 
```python
  for value in unique_pixels:
    indices = np.where(image == value)
    if value < len(colors):
      colr = matplotlib.colors.to_rgb(colors[value])
    else:
      factor = value // len(colors)
      colr_idx = value - (int(factor) * len(colors))
      colr = matplotlib.colors.to_rgb(colors[colr_idx])

    blank_image[indices] = colr
```

The problem with this section is that in certain cases there are more than 24 colors required, colors after `23` are mapped back starting from 0 to the next 23.  Let me know if this suffices, or if you have something better in mind,

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -186,34 +208,45 @@ def repr_html(self, ex: np.ndarray) -> str:
...
+def _create_thumbnail(ex: np.ndarray, use_colormap: bool) -> PilImage:
   """Creates the image from the np.array input."""
   PIL_Image = lazy_imports_lib.lazy_imports.PIL_Image  # pylint: disable=invalid-name
 
   _, _, c = ex.shape
   postprocess = _postprocess_noop
   if c == 1:
-    ex = ex.squeeze(axis=-1)
+    if use_colormap:
+      mode = None
```
Conchylicultor, line 246: Mode here is overwritten by `mode = 'L'` bellow

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -186,34 +208,45 @@ def repr_html(self, ex: np.ndarray) -> str:
...
+      image_with_colormap = _apply_colormap(ex)
+      ex = image_with_colormap * 255
```
Conchylicultor, line 248: `ex = image_with_colormap * 255` should go inside `_apply_colormap` directly no ?

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -345,3 +378,44 @@ def _get_and_validate_shape(shape, encoding_format):
...
+    raise ValueError(
+        f'Colormap is only available for PNG images. Not {encoding_format}'
+    )
+  if shape[-1] != 1 and use_colormap:
+    raise ValueError('Colormap is only available for gray-scale images')
+
+  return use_colormap
+
+
+def _apply_colormap(image: np.ndarray) -> np.ndarray:
```
Conchylicultor, line 401: Could you add unittests for this function where you test the output on simple images ?

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -345,3 +378,44 @@ def _get_and_validate_shape(shape, encoding_format):
...
+  image = image.squeeze(axis=-1)
+  unique_pixels = np.unique(image)
+  blank_image = np.ones(shape=(*image.shape[:2], 3))
+
+  for value in unique_pixels:
+    indices = np.where(image == value)
+    if value < len(colors):
+      colr = matplotlib.colors.to_rgb(colors[value])
+    else:
+      factor = value // len(colors)
```
Conchylicultor, line 415: Note: To circularly apply the color, you could use the modulo op

`value = value % len(colors)`

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -345,3 +378,44 @@ def _get_and_validate_shape(shape, encoding_format):
...
+
+
+def _apply_colormap(image: np.ndarray) -> np.ndarray:
+  """Apply colormap to 1D images"""
+  matplotlib = lazy_imports_lib.lazy_imports.matplotlib  # pylint: disable=invalid-name
+
+  colors = _COLORS
+  image = image.squeeze(axis=-1)
+  unique_pixels = np.unique(image)
+  blank_image = np.ones(shape=(*image.shape[:2], 3))
```
Conchylicultor, line 408: dtype=np.uint8

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -45,6 +44,13 @@
     'jpeg': [tf.uint8],
 }
 
+_COLORS = [
+    '#AA0DFE', '#3283FE', '#85660D', '#782AB6', '#565656', '#1C8356', '#16FF32',
```
Conchylicultor, line 48: I liked your previous idea of using matplotlib for the colormap. How about using the matplotlib API to generate the colormap ?

https://matplotlib.org/3.1.0/tutorials/colors/colormap-manipulation.html

Something like

```python
@utils.memoize()  # Memoize to only be computed once
def _get_colormap() -> np.ndarray:
  """Returns the `np.array(shape=(255, 3), dtype=uint8)` RGB colors."""
```

Conchylicultor (changes requested):
Thank you for the update

tensorflow_datasets/core/features/image_feature.py
Conchylicultor: And the you can simply use something like:

```python
def _apply_colormap(img: np.ndarray) -> np.ndarray:
  img = img.squeeze(axis=-1)
  cmap = _get_colormap()  # (255, 3) id->color mapping 
  return cmap[img]  # Apply mapping from (x, y) -> (x, y, 3)
```

tensorflow_datasets/core/features/image_feature.py
DumbMachine: Could you reference me to another tests? Would love a lead on how to do this.

tensorflow_datasets/core/features/image_feature_test.py
```diff
@@ -118,7 +116,9 @@ def test_image_shaped(self):
 
     self.assertFeature(
         # Image with statically defined shape
-        feature=features_lib.Image(shape=(32, 64, 3), encoding_format='png'),
+        feature=features_lib.Image(
+            shape=(32, 64, 3), encoding_format='png', use_colormap=False
```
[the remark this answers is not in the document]
DumbMachine: Having it as `True` will throw an error as expected. 
>   if shape[-1] != 1 and use_colormap:
                    raise ValueError('Colormap is only available for gray-scale images')
I believe we will have to add another test for this, is that right?

tensorflow_datasets/core/features/image_feature.py
DumbMachine: Have updated the code, to align with your expectations. Let me know if I have used the decorator properly.

tensorflow_datasets/core/features/image_feature_test.py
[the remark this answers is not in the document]
Conchylicultor: Then could you create a new `self.assertFeature(` with `feature=features_lib.Image(shape=(..., 1), use_colormap=True)` to make sure this works ?

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -345,3 +370,46 @@ def _get_and_validate_shape(shape, encoding_format):
...
+    cmap_list.append(cmap(itr))
+
+  return cmap_list
+
+
+def _apply_colormap(image: np.ndarray) -> np.ndarray:
+  """Apply colormap to 1D images"""
+  image = image.squeeze(axis=-1)
+  unique_pixels = np.unique(image)
+  colors = _get_colormap(len(unique_pixels))
```
Conchylicultor, line 406: The number of pixels might change from images to images, but we want to colors to always be the same across images. We should generate a unique colormap of 255 values.

tensorflow_datasets/core/features/image_feature.py
```diff
@@ -345,3 +370,46 @@ def _get_and_validate_shape(shape, encoding_format):
...
+  blank_image = np.zeros(shape=(*image.shape[:2], 3), dtype=np.uint8)
+
+  for itr, value in enumerate(unique_pixels):
```
Conchylicultor, line 409: It would be much more efficient to use the code snippet I posted in my previous comment:

```python
def _apply_colormap(img: np.ndarray) -> np.ndarray:
  img = img.squeeze(axis=-1)  # (x, y, 1) -> (x, y)
  cmap = _get_colormap()  # (255, 3) id->color mapping 
  return cmap[img]  # Apply mapping: (x, y) -> (x, y, 3)
```

tensorflow_datasets/core/features/image_feature_test.py
```diff
@@ -26,7 +25,6 @@
 
 tf.enable_v2_behavior()
 
-
 randint = np.random.randint
```
Conchylicultor, line 30:
```python
def test_apply_colormap():
  img = np.array([...])
  colored_img = image_feature._apply_colormap(img)
  numpy.testing.assert_array_equal(colored_img, [[...], [...]])
```

Conchylicultor (changes requested):
Almost good. We should now make sure to add the proper unit tests

tensorflow_datasets/core/features/image_feature.py
DumbMachine: Generating a colormap for soo many values would lead to non-unique shades (at least to the human eye) ( or I maybe wrong here )
Is this something we want? If not maybe we could have separate mapping zones ( 1-20 shades, 1-50 shades, and .....  this would mean only those with categories>~20 will suffer from non-unique looking shades )

image
image
image
image

tensorflow_datasets/core/features/image_feature.py
DumbMachine: This would be if we generate 255 pixels in the colormap, will change if we end up making 255 unique values in the map. ref https://github.com/tensorflow/datasets/pull/2975#discussion_r564447448

tensorflow_datasets/core/features/image_feature.py
Conchylicultor: This is a good point. Basically what we want is that numbers close to each others have high contrast. But indeed, many values will have similar colors.
So it would be something like 

```
0 -> black
1 -> red
2 -> green
3 -> blue
...
123 -> slight variation on blue
124 -> slight variation on red
...
```

So an image which has contains values `[1, 3, 0]` would have distinct colors, but an image with  values [3, 123] would have very low contrast (as 3 and 123 correspond to similar colors in the above mapping). In practice it wouldn't really matter as the likelihood that `[3, 123]` appear on the same image is low.

But more important, because the same colors are applied on all images, so value `x` will always be displayed the same across all images.

One way to create the color map would be randomly shuffle the values (using fixed seed for determinism). Another likely smarter way would be to reorder the mapping returned by matplotlib to better distribute the values (to have something more like my example mapping above)

Another based color could be used too, like `gist_ncar` which seems to have more color variations

DumbMachine: Here is the contrast of the colormaps, feels like only the order has changed for the colors.
image
image

I believe our motive is to assign deterministic colors for pixel values where any pixel from any image ( from the same dataset ) will have the same colormap. Is this understanding correct? If that's the case, we can use some metadata like  `dataset_name` as the seed to generate uniform color distribution from the large color space ( 255 values ), this would ensure that the colors are distinguishable. What do you think about this implementation?

Conchylicultor: Let me think of an algorithm to generate the colormap as we want and I'll post it there maybe tomorrow.

In the meantime, could you assume `_get_colormap()` returns a `[255, 3]` vector (you can have `_get_colormap` return a dummy `range(255*3).reshape(255,3)` for now) and add the update the tests ?

Conchylicultor (approved):
As update I've used https://github.com/taketwo/glasbey algorithm to pre-compute a colormap which satisfied the property we wanted (distinct colors).

I've already imported the PR internally to apply the updates. This should be merged soon

PR #2980, opened 2021-01-22
Fix html tag visualization for Ragged Tensor

### LINKED ISSUE #2966: Better `tfds.as_dataframe` ragged tensor visualization
Currently, datasets which contains nested sequences `Sequence(Sequence())` are not correctly displayed.

For example: Click on `Display examples...` in https://www.tensorflow.org/datasets/catalog/arc

We should investigate why nothing seems to be displayed. In theory, calling `tfds.as_dataframe(ds, ds_info)` should call `FeatureConnector.repr_html_ragged(ex)` on the ragged tensor and display its `repr`.

jatin-code777:
Fixes #2966 

On investigating the issue, it turns out that when the data is a RaggedTensor, the `repr_html_ragged` does get called which in turn calls the `_repr_` of ReggedTensor, which contains `<` and `>` symbols, which are then parsed as HTML when rendering.

Hence, the fix was to replace `<` and `>` characters in the HTML with `&lt;` and `&gt;` which is also what is done somewhere in pandas `_repr_html_`

### DIFF
--- a/tensorflow_datasets/core/as_dataframe.py
+++ b/tensorflow_datasets/core/as_dataframe.py
@@ -82,9 +82,14 @@ class ColumnInfo:
     elif sequence_rank > 1:
       repr_fn = feature.repr_html_ragged
 
-    def repr_fn_with_debug(val):  # Wrap repr_fn to add debug info
+    def repr_fn_wrapper(val):  # Wrap repr_fn to add debug info & fix html tags
       try:
-        return repr_fn(val)
+        html_repr =  repr_fn(val)
+        # Replace '<' and '>' characters in the html with '&lt;' and '&gt;'
+        # so that they are not confused as html tags when rendering
+        # Inspired from pandas _repr_html_
+        html_repr = html_repr.replace('<', r'&lt;', 1).replace('>', r'&gt;', 1)
+        return html_repr
       except Exception as e:  # pylint: disable=broad-except
         err_msg = (
             f'HTML formatting of column {name} failed:\n'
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/core/as_dataframe.py
```diff
@@ -82,9 +82,14 @@ def from_spec(
...
 
-    def repr_fn_with_debug(val):  # Wrap repr_fn to add debug info
+    def repr_fn_wrapper(val):  # Wrap repr_fn to add debug info & fix html tags
       try:
-        return repr_fn(val)
+        html_repr =  repr_fn(val)
+        # Replace '<' and '>' characters in the html with '&lt;' and '&gt;'
+        # so that they are not confused as html tags when rendering
+        # Inspired from pandas _repr_html_
+        html_repr = html_repr.replace('<', r'&lt;', 1).replace('>', r'&gt;', 1)
```
Conchylicultor, line 91: This is not the right place to escape this, as `repr_fn` might also returns actual HTML code (e.g. to display images, video,...)

Instead, we should make the update in `def _repr_html()` in `core/features/feature.py`

Conchylicultor, line 91: It would be cleaner to use `html.escape(` instead

Conchylicultor (changes requested):
Thank you for investigating and finding the issue!

jatin-code777:
Thank you for the quick review. I have made the requested changes

I had thought about the wrong placement of escape code myself earlier and was about to change it.
Thanks for the `html.escape` tip!

Conchylicultor (approved):
Thank you for the quick update!! Looks good!

PR #2981, opened 2021-01-23
Updated extractor to show better progress bar with number of files

### LINKED ISSUE #2979: Better progression bar for extraction
Currently, extracting big archive made it looks TFDS is frozen. It would be great to have a more granular progression bar counting the extract number of files, rather than just extracted/non-extracted.

https://github.com/tensorflow/datasets/blob/0540c979cba848d814d8266e807352063b4ef355/tensorflow_datasets/core/download/extractor.py#L61

For the implementation, it would be like the download bar which indicates the combined size of all download: https://github.com/tensorflow/datasets/blob/0540c979cba848d814d8266e807352063b4ef355/tensorflow_datasets/core/download/downloader.py#L138

To detect the number of file, we could use `TarFile.getnames()`, `ZipFile.namelist()`

jatin-code777:
Fixes #2979

I looked into this issue and it required a bit of restructuring to implement.

Previously we mapped extract methods directly to `iter_<type>` but it was difficult to accommodate 'number of files' with it.
So this PR creates a base class which each extraction type inherits from and implements `get_num_files` and `get_iter`.

There are some points I would like comments on

- Currently, I have opened the file once in `get_num_files` and then again when `get_iter` is called. Should I try to optimize this to not open twice but store the opened file when the object is initialized? This might be done by making _ArchiveFile itself a context manager, and close the file when its context ends. This would also require manually calling `__enter__` and `__exit__` of `_open_or_pass` in `_ArchiveFile`

- Currently in the base class, I have used `raise NotImplementedError()`. I can also use `abc.ABC` for the same. Would that be better?

- There is some performance hit to get the number of files, which scales linearly with file size. Should there be an option to opt-out? It is still much less compared to the actual extraction time, so perhaps it is not that much of an issue.

- The other `iter_<type>` functions were removed except for `iter_archive` which is the common one. Its implementation was modified to accommodate the new code. I thought it better not to remove it since it is used by some datasets directly. Is this okay?

- Mentioned inline

Thank you

### DIFF
--- a/tensorflow_datasets/core/download/extractor.py
+++ b/tensorflow_datasets/core/download/extractor.py
@@ -253,6 +253,16 @@ def iter_gzip(arch_f):
     yield ('', gzip_)  # No inner file.
 
 
+class _Bzip2File(_ArchiveFile):
+
+  def get_num_files(self) -> int:
+    return 1
+
+  def get_iter(self):
+    with _open_or_pass(self.arch_f) as fobj:
+      bz2_ = bz2.BZ2File(filename=fobj)
+      yield ('', bz2_)  # No inner file.
+
 def iter_bzip2(arch_f):
   with _open_or_pass(arch_f) as fobj:
     bz2_ = bz2.BZ2File(filename=fobj)
--- a/tensorflow_datasets/core/download/extractor.py
+++ b/tensorflow_datasets/core/download/extractor.py
@@ -237,6 +237,15 @@ def iter_tar(arch_f, stream=False):
 def iter_tar_stream(arch_f):
   return iter_tar(arch_f, stream=True)
 
+class _GzipFile(_ArchiveFile):
+
+  def get_num_files(self) -> int:
+    return 1
+
+  def get_iter(self):
+    with _open_or_pass(self.arch_f) as fobj:
+      gzip_ = gzip.GzipFile(fileobj=fobj)
+      yield ('', gzip_)  # No inner file.
 
 def iter_gzip(arch_f):
   with _open_or_pass(arch_f) as fobj:
--- a/tensorflow_datasets/core/download/extractor.py
+++ b/tensorflow_datasets/core/download/extractor.py
@@ -139,6 +139,18 @@ def _open_or_pass(path_or_fobj):
     yield path_or_fobj
 
 
+class _ArchiveFile:
+
+  def __init__(self, arch_f) -> None:
+    self.arch_f = arch_f
+
+  def get_num_files(self) -> int:
+    raise NotImplementedError()
+
+  def get_iter(self):
+    raise NotImplementedError()
+
+
 def iter_tar(arch_f, stream=False):
   """Iter over tar archive, yielding (path, object-like) tuples.
 
--- a/tensorflow_datasets/core/download/extractor.py
+++ b/tensorflow_datasets/core/download/extractor.py
@@ -69,7 +69,6 @@ class _Extractor(object):
     """Returns `promise.Promise` => to_path."""
     path = os.fspath(path)
     to_path = os.fspath(to_path)
-    self._pbar_path.update_total(1)
     if extract_method not in _EXTRACT_METHODS:
       raise ValueError('Unknown extraction method "%s".' % extract_method)
     future = self._executor.submit(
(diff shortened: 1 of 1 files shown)


tensorflow_datasets/core/download/extractor.py
```diff
@@ -240,4 +276,6 @@ def iter_archive(
     raise ValueError(
         f'Cannot `iter_archive` over {path}. Invalid or unrecognised archive.'
     )
-  return _EXTRACT_METHODS[method](path)  # pytype: disable=bad-return-type
+  archive_file  = _get_archive_file(path, method)
+  kwargs = {'stream': True} if method.name.endswith('STREAM') else {}
+  return archive_file.get_iter(**kwargs)
```
jatin-code777, line 281: I'm not sure if this is the best way to implement this.
Another idea I had was for the `_TarFile` class to have a parameter `stream` and then the mapping could be changed to `_Tarfile(stream=True)` where required in `_EXTRACT_METHODS`

This would be much cleaner, but I'm not sure how to implement this properly. 
Other possibilities are 
- to have a function `_TarFile` which takes an argument `stream` and creates a class inside it bases on the arg and return that class. We should also memorize this so it returns the same class every time for the same parameter. 
- create two classes one for stream and one plain. But this would lead to a lot of code duplication

dhananjayraut:
nice work

jatin-code777:
@Conchylicultor could you please review this PR?

Conchylicultor (changes requested):
Thank you for looking into this!
Maybe we should thing a little bit more of the design. I think we should try to avoid opening twice the archive, as this adds overhead, especially when writing to GCS.

Also some users already use direct calls of `tfds.download.extractor.iter_xyz`, so we do not want to break those.

A suggestion would be to keep the same API, but rather than yielding the attributes directly, return some `_ArchiveIterator` with `__len__` and `__iter__` method defined:

```python
def iter_xyz(arch_f, stream=False) -> _ArchiveIterator:


arch_iter = iter_xyz(...)
print(len(arch_iter))
for elem in arch_iter:
  ...
```
The file would be open in `_ArchiveIterator.__init__` and closed after `_ArchiveIterator.__iter__` is exausted.

jatin-code777:
> closed after `_ArchiveIterator.__iter__` is exausted

In this case, if it happens that a user breaks out of the loop early, our cleanup won't be done.
[PEP 533](https://www.python.org/dev/peps/pep-0533/) has exactly what we want, but unfortunately, it was rejected/deferred.

Also, how do you suggest I do the cleanup, I can call `_open_or_pass(arch_f).__enter__()` explicitly to get the fobj.
If this is done, should I call exit(file close/cleanup) as `_open_or_pass(arch_f).__exit__(None, None, None)`?

Conchylicultor:
> In this case, if it happens that a user breaks out of the loop early, our cleanup won't be done.

But this is the case currently too, no ? In the `iter_xyz()`, the `with open()` is only exited after the iterator

```python
@contextlib.contextmanager
def open_file():
    print('Enter')
    yield
    print('Exit')


def iter_archive():
    with open_file():
        for _ in range(10):
            yield


def f():
    for _ in iter_archive():
        return  # Exit early, open_file never closed !!
```

> If this is done, should I call exit(file close/cleanup) as _open_or_pass(arch_f).__exit__(None, None, None)?

Yes, calling `self._open_cm.__exit__(None, None, None)` on exit is good.

jatin-code777:
> But this is the case currently too, no ?

There is a subtle difference. With the previous design since we use a context manager, the `close` will be called when the program exits. This is not the case with the new suggestion.

But since we only open files in read mode, it should not make much of a difference here. (We need to close file in order to get a guarantee that the writes are applied in case of write mode). Moreover, I'm guessing it's not common to break out of loop midway. So I've implemented your suggestion.

Some points
- The `__len__` has been memoized
- In case of stream mode in tar extraction, the file is `seek`ed to beginning and tar file is opened again. This is more efficient than reopening the file.

Conchylicultor:
> There is a subtle difference. With the previous design since we use a context manager, the close will be called when the program exits. This is not the case with the new suggestion.

I don't think this is true. From my example above, the `print('Exit')` is never called, even after the program exit.

jatin-code777:
I thought more about it and read through docs and tested examples. In the example you gave, the `'Exit'` is printed because it is in a `finally` block and not because of the context manager, as I wrongly stated earlier. So when the program exits, (or the iterator returned by `iter_archive()` becomes unreachable), the code in the `finally` block is executed regardless of whether we break early or not.

To ensure that the file is always closed, this is the change I propose:

Replace
```python
@contextlib.contextmanager
def _open_or_pass(path_or_fobj):
  if isinstance(path_or_fobj, utils.PathLikeCls):
    with tf.io.gfile.GFile(path_or_fobj, 'rb') as f_obj:
      yield f_obj
  else:
    yield path_or_fobj
```
with 
```python
@contextlib.contextmanager
def _open_or_pass(path_or_fobj):
  if isinstance(path_or_fobj, utils.PathLikeCls):
    f_obj = tf.io.gfile.GFile(path_or_fobj, 'rb')
    try:
      yield f_obj
    finally:
      f_obj.close()
  else:
    yield path_or_fobj

```

I feel this is a much better alternative in general, guaranteeing that the file gets closed at some point.

PR #2982, opened 2021-01-23
Added public API to enable tqdm corresponding to the one for closing it

jatin-code777:
Progress bar once disabled, cannot be enabled again. This PR adds an API to do so.

I had some use case for this feature myself and had added it locally. Thought I'd upstream it.


Conchylicultor (approved):
Thank you. Looks good

PR #3005, opened 2021-01-31
Fix ag_news_subset checksum

### LINKED ISSUE #3004: ag_news_subset Checksum Error
In trying to download the ag_news_subset dataset I ran into a checksum error. I found a workaround that I have a few questions about, and I am also not sure if this is known bug or not.

**Problem**: Running the tfds load function results in a checksum mismatch error. Interestingly the only way the checksums differed was that TensorFlow had their expected value in all caps.

**Fix**: My fix was to use the CLI and run the following command where data is just a specific directory I want the dataset in:

```tfds build ag_news_subset --register_checksums --data_dir data```

I am not sure if what I did was bad in some way, but it works in the sense that running tfds loads with download=False successfully loads the dataset.

**Environment information**
- Operating System: Pop!_OS 20.10
- Python version: 3.8
- tensorflow-datasets version: 4.2.0
- tensorflow version: 2.4.1

Really I am just looking to better understand what I did, and whether or not this is a known bug.

jatin-code777:
Fixes #3004 

The checksum hasn't changed, but the currently recorded checksum uses uppercase letters when it should be lowercase, thus giving the mismatch error.


NikhilBartwal (approved):
Thank you for the contribution @jatin-code777 ! 

@Conchylicultor I think this is ready to be merged, would be great if you could have a look.

PR #3047, opened 2021-02-17
Parallelize list_builders function call

### LINKED ISSUE #3036: Parallelise  `tfds.list_builder` when community dataset is used
Currently, when `tfds.list_builder()` is called, TFDS will sequentially try to list datasets in each of the registered data_dir which is very slow when there is many data_dir to look at:

https://github.com/tensorflow/datasets/blob/aa0547927fab3504a9980528c486bbe32a3b598c/tensorflow_datasets/core/community/register_path.py#L144

It would be nice if this sequential loop was parallelised, so all `list(path.iterdir())` are executed in parallel.

NikhilBartwal:
`tfds.list_builders` has been parallelized using `multiprocessing.dummy.pool` with multiple threads as the underlying function is IO bound.

Closes #3036

### DIFF
--- a/tensorflow_datasets/core/community/register_path.py
+++ b/tensorflow_datasets/core/community/register_path.py
@@ -162,6 +162,6 @@ def _iter_builder_names(
   # Note: `data_dir` might contain non-dataset folders, but checking
   # individual dataset would have significant performance drop, so
   # this is an acceptable trade-of
-  pool = ThreadPool(10)
-  for result in pool.starmap(_parallel, ns2data_dir.items()):
-    yield from result
+  with ThreadPool(4) as pool:
+    for result in pool.starmap(_parallel, ns2data_dir.items()):
+      yield from result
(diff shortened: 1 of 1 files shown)


jatin-code777:
Hi @NikhilBartwal, I had already started work on this, you can see my changes committed yesterday on my PR(#3045).

I was hoping to think more on it before pushing a PR, but since I wanted to prevent work 'duplication', I published a draft PR earlier. Perhaps I should have been clearer and commented too. Sorry about that!

I think we should collaborate better since this has happened twice now :)

NikhilBartwal:
Hey @jatin-code777 , indeed this has happened twice now ;p
It is always a good practice to let the other contributors know when you start working on an issue so as to reduce work  'duplication' and collaborate better. 
I will keep the PR open for community review as of now till further approval from the repo members.

Conchylicultor:
Thank you both for your contributions!
Yes, it's best to comment on the issue to avoid duplication, as suggested in our https://www.tensorflow.org/datasets/contribute guide.
Both implementation looks good, but I imported @jatin-code777  one as it was using `concurrent.futures` which is more recent and standard than `multiprocessing.dummy`, and was slightly better formatted.

NikhilBartwal:
@Conchylicultor Thank you for the review! I will keep in mind to use the more recent packages for the contributions from now on. 
I think I should close this PR now that the other one has been imported.

PR #3062, opened 2021-02-26
Fix librispeech wrong path

jatin-code777:
A separate PR for the librispeech.py fix since the original author of #2181 hasn't responded for a while


NikhilBartwal (commented):
Thank you for the contribution @jatin-code777. 
As @Conchylicultor mentioned on the original PR, the previous code should perform the same job. Nonetheless, it only leads to code duplication and is better omitted.

NikhilBartwal (approved):

ashaw596:
Could we get this fixed? I had to manually pull these changes to generate the dataset.

ashaw596:
I don't believe it's the same the code errors when generating on buckets previously.

ashaw596:
@NikhilBartwal @jatin-code777

ashaw596:
@tfds-copybara

Issue #3116, opened 2021-03-18
Add `tfds list`  CLI command


vijayphoenix:
Addition of community dataset will sharply increase the no of available datasets. So it will become difficult to browse through repos or catalog.

So, it would great if we can add a new CLI command to list all the datasets in present in TFDS. This will allows users to easily search a dataset using tools like grep.

**Possible Usage**
```
tfds list # List all datasets
tfds list --type audio # List all audio dataset in TFDS
tfds list --namespace huggingface # List all huggingface community dataset
tfds list --search sun # Lists datasets like sun397, lsun (alternative is to use grep)
```

**Implementation details**

• Use/modify the `tfds.list_builder` API.
• Add a new file `list.py` in `scripts/cli` folder

jatin-code777:
Hi, I have started work on this

NikhilBartwal:
Since @jatin-code777 has already started working on `--type` and `--exclude_community` part, I will be sending a PR for the 
`--search` and `--namespace` flags soon.

Srikeshram:
I am working on this issue. Will you please assign it to me?

NikhilBartwal:
Hey @Srikeshram,  @jatin-code777 is already working on the issue and taking it up would only lead to work duplication. You can take up other bugs and issues which are still unresolved in TFDS. Thank you for the efforts!

PR #3117, opened 2021-03-18
Add `tfds list` command to the CLI

jatin-code777:
Addresses #3116 

### TODO:

- [ ] Add `--type` flag (Lists only only non-community datasets)
- [ ] Add `--exclude-community` flag

Will add other flags and options in subsequent PRs

### DIFF
--- a/tensorflow_datasets/scripts/cli/list.py
+++ b/tensorflow_datasets/scripts/cli/list.py
@@ -1,0 +1,23 @@
+
+"""`tfds list` command."""
+
+import argparse
+
+from tensorflow_datasets.core import load
+
+
+def register_subparser(parsers: argparse._SubParsersAction) -> None:  # pylint: disable=protected-access
+  """Add subparser for `list` command."""
+  new_parser = parsers.add_parser('list', help='Lists all datasets')
+  new_parser.set_defaults(subparser_fn=_list_datasets)
+
+
+def _list_datasets(args: argparse.Namespace) -> None:
+  """Creates the dataset directory. Executed by `tfds new <name>`."""
+  list_datasets()
+
+
+def list_datasets():
+  datasets = load.list_builders(with_community_datasets=True)
+  for dataset in datasets:
+    print(dataset)
--- a/tensorflow_datasets/scripts/cli/main.py
+++ b/tensorflow_datasets/scripts/cli/main.py
@@ -38,6 +38,7 @@ import tensorflow_datasets.public_api as tfds
 # Import commands
 from tensorflow_datasets.scripts.cli import build
 from tensorflow_datasets.scripts.cli import new
+from tensorflow_datasets.scripts.cli import list as list_
 
 FLAGS = flags.FLAGS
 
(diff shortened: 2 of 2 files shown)


NikhilBartwal:
@jatin-code777 Is this PR still in progress?

jatin-code777:
@NikhilBartwal Yes, it is. Since there are no reviews, I suppose I'll go ahead with the same setup. I have also been a little busy, hence the delay.

tensorflow_datasets/scripts/cli/list.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def register_subparser(parsers: argparse._SubParsersAction) -> None:  # pylint: disable=protected-access
+  """Add subparser for `list` command."""
+  new_parser = parsers.add_parser('list', help='Lists all datasets')
+  new_parser.set_defaults(subparser_fn=_list_datasets)
+
+
+def _list_datasets(args: argparse.Namespace) -> None:
+  """Creates the dataset directory. Executed by `tfds new <name>`."""
```
NikhilBartwal, line 16: This should be modified for `tfds list` command

tensorflow_datasets/scripts/cli/list.py
```diff
@@ -0,0 +1,23 @@
...
+  list_datasets()
+
+
+def list_datasets():
+  datasets = load.list_builders(with_community_datasets=True)
```
NikhilBartwal, line 21: Since, `community_datasets` argument is `True` by default, there's not really any need to write it explicitly here.

tensorflow_datasets/scripts/cli/list.py
```diff
@@ -0,0 +1,23 @@
...
+  for dataset in datasets:
```
NikhilBartwal, line 22: I think we should simply print the whole list as it as, just like `tfds.list_builders()` since, printing each dataset in newline will generate a pretty huge list, which might not be ideal for the user to browse.

NikhilBartwal (changes requested):

tensorflow_datasets/scripts/cli/list.py
jatin-code777: As mentioned in the original Issue it should be easily greppable, and following the way this is usually done, each dataset in a new line makes sense.

tensorflow_datasets/scripts/cli/list.py
jatin-code777: Since it doesn't make a difference anyway, I felt like writing this here would be clearer and more explicit. Both approaches are appropriate but being explicit seems cleaner to me. No harm.

tensorflow_datasets/scripts/cli/list.py
jatin-code777: Not sure what you mean but this will be heavily changed for the new flags. This function's purpose is to handle the `args` (including validation) and change control flow from here on. No other function should have to deal with `args`. Please review it again then!

PR #3149, opened 2021-04-08
Add necessary gtzan extras

carlthome:
Just a small tweak. `tfds.load("gtzan")` in a fresh Colab notebook complains that pydub is also needed (for the audio decoding).


Conchylicultor (approved):
Thank you

PR #3157, opened 2021-04-12
Load metadata from metadata.json if there is any

jneeven:
Currently, specifying a dataset version in `tfds.load` will try to load the dataset from files using the `ReadOnlyBuilder` (see https://github.com/tensorflow/datasets/pull/2493). However, since it never specifies any metadata (see [`ReadOnlyBuilder._info`](https://github.com/tensorflow/datasets/blob/55f84b5723649a5bdedcce6b0343d8d0cb609140/tensorflow_datasets/core/read_only_builder.py#L101)), [the part of `DatasetInfo.read_from_directory` that is responsible for loading the metadata](https://github.com/tensorflow/datasets/blob/55f84b5723649a5bdedcce6b0343d8d0cb609140/tensorflow_datasets/core/dataset_info.py#L383) is never triggered, even though metadata might exist.

This PR creates a new `MetadataDict` if a `metadata.json` exist, and then loads the metadata from there. I've also added a unit test to verify that this behaviour previously did not work and now does. While that fixes the problem on my side, @Conchylicultor is there any reason why we use the builder code for loading the dataset at all (if no version is specified)? Since we can evidently load datasets from file just fine, and all the necessary information is present in the directory, why would we not always load from file if they exist? It seems to me that the builder should only be used for actually building a new version of the dataset.

Curious to hear your thoughts!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,11 +15,7 @@
             80
         ],
         "editor.tabSize": 2,
-        "editor.detectIndentation": false,
-        "editor.formatOnSave": false,
-        "editor.codeActionsOnSave": {
-            "source.organizeImports": false
-        }
+        "editor.detectIndentation": false
     },
     "python.formatting.provider": "yapf",
     "python.testing.pytestArgs": [
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,7 +15,11 @@
             80
         ],
         "editor.tabSize": 2,
-        "editor.detectIndentation": false
+        "editor.detectIndentation": false,
+        "editor.formatOnSave": false,
+        "editor.codeActionsOnSave": {
+            "source.organizeImports": false
+        }
     },
     "python.formatting.provider": "yapf",
     "python.testing.pytestArgs": [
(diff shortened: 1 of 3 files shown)


jneeven:
@googlebot I signed it!

Conchylicultor (approved):
Thank you! Looks good

jneeven:
@Conchylicultor thanks for the quick response! Regarding loading from files vs loading from builder, do you think it makes sense to _always_ load from file? That should hopefully prevent bugs and confusion in the future, since there is only one "path" from which a dataset can be loaded (dataset creation of course still needs the builder code).

Conchylicultor:
> @Conchylicultor thanks for the quick response! Regarding loading from files vs loading from builder, do you think it makes sense to _always_ load from file? That should hopefully prevent bugs and confusion in the future, since there is only one "path" from which a dataset can be loaded (dataset creation of course still needs the builder code).

Yes, I agree this would be nice. Unfortunately, not all datasets can be loaded from files currently. For instance, datasets which are using custom `FeatureConnector`.
Another minor use-case is that when updating info (description, homepage,...) currently the documentation is automatically updated without having to regenerate the full dataset.

PR #3194, opened 2021-05-01
Remove redundant code for `pathlib.Path().resolve()`

NikhilBartwal:
`pathlib.Path().resolve()` used to fail for some systems in the earlier versions. However, this was fixed in `py3.7`
This PR removes the redundant code and addresses the TODO


Conchylicultor:
But TFDS is still compatible with 3.6 right ?

NikhilBartwal:
@Conchylicultor I agree. I didn't consider that TFDS was still compatible with `py3.6`. Closing this PR for now. Thanks!

PR #3211, opened 2021-05-09
Add `tfds.features.Optional`

### LINKED ISSUE #3207: Support optional tensors
TF recently added optional support: https://www.tensorflow.org/api_docs/python/tf/experimental/Optional 
It seems to be working inside tf.data:

```python
ds = tf.data.Dataset.range(3)
ds = ds.map(lambda x: tf.experimental.Optional.from_value(x))
list(ds)[0].get_value()
```

We should update our feature connectors to accept None to return optionals. Maybe something like

```python
def _info(self):
  features=FeatureDict({
      'image': tfds.features.Optional(tfds.features.Image()),
  })

def _generate_examples(self):
  yield {
      'image': None,  # Accept `None` or feature
  }
```
Which would returns:
```python
ds = tfds.load()
ds.element_spec = {'image': OptionalSpec(TensorSpec(shape=(None, None, 3), dtype=tf.uint8, name=None),)}
```

NikhilBartwal:
Resolves #3207 
This PR aims to add the new `tfsd.features.Optional` feature which can act as a wrapper for all other `tfds.features` such that the user can even pass `None` as suitable value wduring `_generate_examples()`

[NOTE: This PR is a work in progress]

### DIFF
--- a/tensorflow_datasets/core/features/optional_feature.py
+++ b/tensorflow_datasets/core/features/optional_feature.py
@@ -1,0 +1,54 @@
+# coding=utf-8
+# Copyright 2021 The TensorFlow Datasets Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Optional feature."""
+
+import tensorflow.compat.v2 as tf
+
+from tensorflow_datasets.core.features import feature
+
+
+class Optional(feature.FeatureConnector):
+    """Optional `FeatureConnector` for wrapping other features.
+
+    `Optional` defines a wrapper for any other `tfds.features.FeatureConnector`.
+    At generation time, the user can even pass `None` as a suitable value to
+    the Optional feature
+
+    NOTE: `Optional` does not support Nested features in itself. However, it
+    can be achieved by wrapping a `tfds.features.Sequence` feature.
+
+    During `_generate_examples` the Optional feature can take as input:
+
+      * Any type of input supported by the wrapped feature
+      * `None` to denote absence of any input
+
+    Example:
+    At construction time:
+    ```python
+    features=features.FeatureDict({
+        'image': features.Optional(features.Image()),
+        'label': features.Optional(features.ClassLabel()),
+    })
+    ```
+
+    At generation time:
+    ```python
+    yield{
+        'image': 'path/to/image.jpg',
+        'label': None
+    }
+    ```
+    """


NikhilBartwal:
Hey @Conchylicultor, although I've added the basic fields and methods for `Optional` feature such that it works successfully with wrapping any normal feature and yielding proper values.
However, the main issue comes with encoding and decoding the `None` values in the proper format (compatible with `tf.train.Example`)

Currently, this is what I have in mind (and have been trying for quite some time now ):
* We can check if the example passed to `Optional.encode_example()` is `None` and then return a string (like `'OptionalNone'`). This part is working as TF was successfully encoding and writing `'OptionalNone'` when `None` was yielded.
* The concerning part is when, in the `Optional.decode_example()`, we need to check if the `tf_exampledata` passed to the `decode_example` contains `'OptionalNone'` or some other value, depending upon which we can simply use either `Optional.from_value` or `Optional.empty()`.
Since, the `tfexample_data` is a pure tensor, I'm confused as to how to check its value.
The `Optional.from_value()` seems to be working pefectly fine (in case i remove the checking and yielding all non-None data) and gives `OptionalSpec` as expected.

Let me know your thoughts about encoding and decoding the `None` values if you have any solution or any different approach in mind. Thanks!

NikhilBartwal:
Hey @Conchylicultor, the `Optional` feature, as implemented in this PR is working successfully as expected.
This was checked with editing `mnist` script from this:
https://github.com/tensorflow/datasets/blob/d5a2b6a7ff71c1cb0c262b7fff27ccc3dffdae05/tensorflow_datasets/image_classification/mnist.py#L101
to this:

`            "image": tfds.features.Optional(tfds.features.Image(shape=MNIST_IMAGE_SHAPE)),`

and this:
https://github.com/tensorflow/datasets/blob/d5a2b6a7ff71c1cb0c262b7fff27ccc3dffdae05/tensorflow_datasets/image_classification/mnist.py#L157

to this:
`      record = {"image": image if index%2==0 else None, "label": label}`

(for adding some `None` and some valid values to the dataset)

The dataset generated with `nik = tfds.load('mnist')` successfully gives::
```
>>> nik
<PrefetchDataset shapes: {image: OptionalSpec(TensorSpec(shape=(28, 28, 1), dtype=tf.uint8, name=None),), label: ()}, types: {image: OptionalSpec(TensorSpec(shape=(28, 28, 1), dtype=tf.uint8, name=None),), label: tf.int64}>
>>> nik.element_spec
{'image': OptionalSpec(TensorSpec(shape=(28, 28, 1), dtype=tf.uint8, name=None),), 'label': TensorSpec(shape=(), dtype=tf.int64, name=None)}
>>> list(nik)[0]['image'].has_value()
<tf.Tensor: shape=(), dtype=bool, numpy=False>
>>> list(nik)[1]['image'].has_value()
<tf.Tensor: shape=(), dtype=bool, numpy=True>
>>> list(nik)[1]['image'].get_value()
<tf.Tensor: shape=(28, 28, 1), dtype=uint8, numpy=
array([[[  0],
        [  0],
        [  0],
        [  0],
```

tensorflow_datasets/core/features/optional_feature.py
```diff
@@ -0,0 +1,106 @@
...
+        if data is None:
+            return 'OptionalNone'
+        else:
+            return self.feature.encode_example(data)
+
+    def decode_example(self, tfdata):
+        """Decode example based on whether it is a None value or a valid one."""
+        def is_none():
+            return tf.experimental.Optional.empty(
+                tf.TensorSpec(shape=(28,28,1), dtype=self.feature._dtype, name=None)
```
Conchylicultor, line 90: You hadcoded the special use-case, so this will only work for mnist. We want this feature to work in the general case. Instead the shape/dtype should be inferred from the `feature.get_serialized_info()`

One way to make sure this works with all features would be to add a `subTest` to test wrapping the features in `Optional`:
https://github.com/tensorflow/datasets/blob/2cfb06097c653ec118de539647b895caecbf58e9/tensorflow_datasets/testing/feature_test_case.py#L121
Then run the unittests in `core/features/` and try to support as many use-case as possible.

Conchylicultor (changes requested):

tensorflow_datasets/core/features/optional_feature.py
NikhilBartwal: Ah, it just slipped my mind I guess that we can simply use the `get_tensor_info().shape` to infer the wrapped feature's shape xD

> You hadcoded the special use-case, so this will only work for mnist. 

Yes, actually I wanted to discuss how to infer the shape of the wrapped feature so I just put this for the time being. No issues now. I will be adding the tests soon. Thanks!

PR #3240, opened 2021-05-21
Add missing `person` builder config to LSUN dataset.

AdamHillier:
The LSUN dataset on TDFS includes all [10 available scene categories](http://dl.yf.io/lsun/scenes/) and 19 out of the [20 available object categories](http://dl.yf.io/lsun/objects/).

The missing object category is `person` - likely omitted because it is by far the largest.

This PR adds the person category, which makes the LSUN dataset complete.

### DIFF
--- a/tensorflow_datasets/image/lsun.py
+++ b/tensorflow_datasets/image/lsun.py
@@ -75,6 +75,7 @@ _OBJECTS_CATEGORIES = [
     "dog",
     "horse",
     "motorbike",
+    "person",
     "potted_plant",
     "sheep",
     "sofa",
@@ -96,9 +97,10 @@ class Lsun(tfds.core.GeneratorBasedBuilder):
       tfds.core.BuilderConfig(  # pylint: disable=g-complex-comprehension
           name=category,
           description="Images of category %s" % category,
-          version=tfds.core.Version("3.0.0"),
+          version=tfds.core.Version("3.0.1"),
           release_notes={
               "3.0.0": "New split API (https://tensorflow.org/datasets/splits)",
+              "3.0.1": "Add builder config for missing `person` object category",
           },
       ) for category in (_SCENES_CATEGORIES + _OBJECTS_CATEGORIES)
   ]
--- a/tensorflow_datasets/image/lsun.py
+++ b/tensorflow_datasets/image/lsun.py
@@ -97,10 +97,10 @@ class Lsun(tfds.core.GeneratorBasedBuilder):
       tfds.core.BuilderConfig(  # pylint: disable=g-complex-comprehension
           name=category,
           description="Images of category %s" % category,
-          version=tfds.core.Version("3.0.1"),
+          version=tfds.core.Version("3.1.0"),
           release_notes={
               "3.0.0": "New split API (https://tensorflow.org/datasets/splits)",
-              "3.0.1": "Add builder config for missing `person` object category",
+              "3.1.0": "Add builder config for missing `person` object category, and add `id` to the feature dict",
           },
       ) for category in (_SCENES_CATEGORIES + _OBJECTS_CATEGORIES)
   ]
@@ -111,6 +111,7 @@ class Lsun(tfds.core.GeneratorBasedBuilder):
         description=("Large scale images showing different objects "
                      "from given categories like bedroom, tower etc."),
         features=tfds.features.FeaturesDict({
+            "id": tfds.features.Text(),
             "image": tfds.features.Image(encoding_format="jpeg"),
         }),
         homepage="https://www.yf.io/p/lsun",
@@ -156,6 +157,9 @@ class Lsun(tfds.core.GeneratorBasedBuilder):
     with tf.Graph().as_default():
       path = os.path.join(extracted_dir, file_path, "data.mdb")
       dataset = _make_lmdb_dataset(path)
-      for i, (_, jpeg_image) in enumerate(tfds.as_numpy(dataset)):
-        record = {"image": io.BytesIO(jpeg_image)}
+      for i, (id_bytes, jpeg_image) in enumerate(tfds.as_numpy(dataset)):
+        record = {
+          "id": id_bytes.decode("utf-8"),
+          "image": io.BytesIO(jpeg_image),
+        }
         yield i, record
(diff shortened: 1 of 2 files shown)


tensorflow_datasets/image/lsun.py
```diff
@@ -96,9 +97,10 @@ class Lsun(tfds.core.GeneratorBasedBuilder):
       tfds.core.BuilderConfig(  # pylint: disable=g-complex-comprehension
           name=category,
           description="Images of category %s" % category,
-          version=tfds.core.Version("3.0.0"),
+          version=tfds.core.Version("3.0.1"),
```
NikhilBartwal, line 100: Since, the dataset now has a completely new `BuilderConfig`, I think it would be better to version it `3.1.0`

tensorflow_datasets/url_checksums/lsun.txt
```diff
@@ -32,6 +32,7 @@ http://dl.yf.io/lsun/objects/dining_table.zip	51826252804	4c40a17884c7180a8f75a7
 http://dl.yf.io/lsun/objects/dog.zip	155606131278	dab08a84ce3facb2ac7261f58c1eea7f070a6d118b82f9778b6a1cf56ed15db4	dog.zip
 http://dl.yf.io/lsun/objects/horse.zip	73864289242	2ec7893f3308da7f68cc2bdb1e9a75cde7abf842628feea1c49c30006e9c3a1b	horse.zip
 http://dl.yf.io/lsun/objects/motorbike.zip	44948708338	76cba2c366269bf86ce57ff496a229b0a237d1095e4bddeec02c9c6b9aa1b9a3	motorbike.zip
+http://dl.yf.io/lsun/objects/person.zip	511754125920	4f6c58c308dfde383d77a018eed37a14e328cd7115b40396a38e15db68b00520	person.zip
```
NikhilBartwal, line 35: Have you generated the `checksums` locally on your end? If yes, could you please us tell a bit about the hardware and time utilized in the dataset generation. Since it is a >500GBs dataset, it might be a better option to use Apache Beam here.

NikhilBartwal (approved):
Thank you so much for the contribution @AdamHillier !

tensorflow_datasets/image/lsun.py
AdamHillier: Sure, done in 6686f37

tensorflow_datasets/url_checksums/lsun.txt
AdamHillier: I did generate the checksums myself, yes.

For reference, the person category zip file is ~477 GB to download, and the generated TFRecords add up to 439 GB.

I actually used a GCP virtual machine to do it, one with: 32 vCPU cores, 32 GB RAM, 4 TB disk. However, I only did this because my local internet connection is slow. The overall process took around 15 hours, where the vast majority of time was taken by the initial download phase - only the last 90 mins was dataset generation.

I monitored resource usage it was always (during both the download phase and data generation phase) using < 20% CPU and < 10 GB RAM. Also, 4 TB disk wasn't required - at the end I had > 50% disk space remaining.

NikhilBartwal: Thank you so much for the info @AdamHillier!
As per the given stats (evidently the low CPU usage), I think it would be better to use `Apache Beam` during dataset generation, instead of a casual loop as we might benefit from the underlying parallelization when generating the dataset in a distributed environment like GC Dataflow.
Please feel free to add the required changes for utilizing `Beam` during dataset generation as demonstrated [here](https://www.tensorflow.org/datasets/beam_datasets#instructions).

Also, it would be great if you could share the generated dataset's `dataset_info.json` file as it will help with updating the dataset documentation on the TFDS catalog page. Thanks!

tensorflow_datasets/image/lsun.py
```diff
@@ -156,6 +157,9 @@ def _generate_examples(self, extracted_dir, file_path):
     with tf.Graph().as_default():
       path = os.path.join(extracted_dir, file_path, "data.mdb")
       dataset = _make_lmdb_dataset(path)
-      for i, (_, jpeg_image) in enumerate(tfds.as_numpy(dataset)):
-        record = {"image": io.BytesIO(jpeg_image)}
+      for i, (id_bytes, jpeg_image) in enumerate(tfds.as_numpy(dataset)):
+        record = {
+          "id": id_bytes.decode("utf-8"),
```
NikhilBartwal, line 162: I'm not really sure about changing the structure of the dataset by adding in a new `id` feature as it might cause some issues for the previous users. 

Also, since the different configs consist of images belonging to the same category (of each config), I don't think the `id` feature would be required. Let me know your thoughts on this.

tensorflow_datasets/url_checksums/lsun.txt
AdamHillier: Ah sorry for the delay - here is the `dataset_info.json`: https://gist.github.com/AdamHillier/5a5e9ccf9bc801a9de11ca5114efcb26.

AdamHillier: For Apache Beam, would you like that to apply only to the `person` builder config or to all configs?

NikhilBartwal: Thank you so much for the json file!
Since, majority of the configs are >100GB, I think it would be better to use `Beam` for all the datasets

tensorflow_datasets/image/lsun.py
AdamHillier: Why do you think adding a feature would break an existing use case? Wouldn't an existing user just ignore the presence of the additional feature?

For my own use case, I need to have the `id` feature, because I want to be able to perform certain per-image analysis tasks -- basically, I want to be able to associate arbitrary additional metadata with specific images, and having a static per-image identifier makes that much easier.

NikhilBartwal: Since, the dataset does not currently have supervised keys (which would have returned a tuple, instead of a dict ), I think you are right that the new feature would only add as an additional for the existing user base.

In that case, I think it would be good to add the supervised keys now (as `('id, 'image')`) as shown [here](https://www.tensorflow.org/datasets/add_dataset#_info_dataset_metadata).

I'll send a PR to update the required documentation once this PR gets approved from a maintainer or merged. Thanks!

Conchylicultor (approved):
Looks good. Thank you

PR #3360, opened 2021-07-14
Dataset generation (tfrecords) speedup

### LINKED ISSUE #3357: Dataset generation is painfully slow
/!\ IF YOU WANT PEOPLE TO HELP YOU, PLEASE GIVE AS MUCH DETAIL AS POSSIBLE, INCLUDING THE FULL STACKTRACE AND CODE SNIPPET

**Short description**
Some might say this is not a bug, but the speed of any non-trivial dataset generation is painfully slow. To the point it becomes unusable.

**Environment information**
* Operating System: Debian
* Python version: 3.9
* `tensorflow-datasets`/`tfds-nightly` version: 4.3.0
* `tensorflow`/`tf-nightly` version: 2.7.0

* Does the issue still exists with the last `tfds-nightly` package (`pip install --upgrade tfds-nightly`) ?
Yes.

**Reproduction instructions**
On my computer it takes too long to build:
`movielens`: about 3k sample per second
`kdd_99_cup`: about 800 samples per second (see #3355 for code)
```
tfds build movielens
```

**Expected behavior**
At least an order of magnitude faster.
The optimization should be done in
1. More efficient single process
2. Multi-process  

**Additional context**

eli-osherovich:
## Description
Some low-hanging fruits to fix #3357 
 
My tests show significant throughput improvement:
~100% on KDD 99 Cup 
~50% on Movielens
  
## Checklist
* [x] Address all TODO's
* [x] Add passing test(s)
* [x] Add test data
* [x] If using additional dependencies (e.g. `scipy`), use [lazy_imports](https://www.tensorflow.org/datasets/add_dataset#extra_dependencies) (if applicable)
* [x] [Lint](https://www.tensorflow.org/datasets/add_dataset#5_check_your_code_style) code

### DIFF
--- a/tensorflow_datasets/core/example_serializer.py
+++ b/tensorflow_datasets/core/example_serializer.py
@@ -118,7 +118,7 @@ def _is_string(item):
 def _item_to_np_array(item, dtype, shape):
   """Single item to a np.array."""
   original_item = item
-  item = np.array(item, dtype=dtype.as_numpy_dtype)
+  item = np.asanyarray(item, dtype=dtype.as_numpy_dtype)
   utils.assert_shape_match(item.shape, shape)
   if dtype == tf.string and not _is_string(original_item):
     raise ValueError(
@@ -134,13 +134,12 @@ def _item_to_tf_feature(item, tensor_info):
   if v.dtype == np.bool_:
     v = v.astype(int)
 
-  v = v.flatten()  # Convert v into a 1-d array
   if np.issubdtype(v.dtype, np.integer):
-    return tf.train.Feature(int64_list=tf.train.Int64List(value=v))
+    return tf.train.Feature(int64_list=tf.train.Int64List(value=v.flat))
   elif np.issubdtype(v.dtype, np.floating):
-    return tf.train.Feature(float_list=tf.train.FloatList(value=v))
+    return tf.train.Feature(float_list=tf.train.FloatList(value=v.flat))
   elif tensor_info.dtype == tf.string:
-    v = [tf.compat.as_bytes(x) for x in v]
+    v = [tf.compat.as_bytes(x) for x in v.flat]
     return tf.train.Feature(bytes_list=tf.train.BytesList(value=v))
   else:
     raise ValueError(
--- a/tensorflow_datasets/core/features/tensor_feature.py
+++ b/tensorflow_datasets/core/features/tensor_feature.py
@@ -16,12 +16,12 @@
 """Feature connector."""
 
 import enum
-from typing import Union
+import functools
 import zlib
+from typing import Union
 
 import numpy as np
 import tensorflow.compat.v2 as tf
-
 from tensorflow_datasets.core import utils
 from tensorflow_datasets.core.features import feature as feature_lib
 
@@ -119,6 +119,7 @@ class Tensor(feature_lib.FeatureConnector):
       }
     return serialized_spec
 
+  @functools.lru_cache(maxsize=2048)
   def encode_example(self, example_data):
     """See base class for details."""
     # TODO(epot): Is there a better workaround ?
@@ -131,19 +132,10 @@ class Tensor(feature_lib.FeatureConnector):
                        f'For {self}')
 
     np_dtype = np.dtype(self.dtype.as_numpy_dtype)
-    if isinstance(example_data, tf.Tensor):
-      raise TypeError(
-          f'Error encoding: {example_data!r}. `_generate_examples` should '
-          'yield `np.array` compatible values, not `tf.Tensor`')
-    if not isinstance(example_data, np.ndarray):
-      example_data = np.array(example_data, dtype=np_dtype)
-    # Ensure the shape and dtype match
-    if example_data.dtype != np_dtype:
-      raise ValueError('Dtype {} do not match {}'.format(
-          example_data.dtype, np_dtype))
 
-    shape = example_data.shape
-    utils.assert_shape_match(shape, self._shape)
+    example_data = np.asanyarray(example_data, dtype=np_dtype)
+    # Ensure the shape matches
+    utils.assert_shape_match(example_data.shape, self._shape)
 
     # Eventually encode the data
     if self._encoded_to_bytes:
--- a/tensorflow_datasets/core/utils/tf_utils.py
+++ b/tensorflow_datasets/core/utils/tf_utils.py
@@ -136,7 +136,7 @@ def is_dtype(value):
   return True
 
 
-@functools.lru_cache()
+@functools.cache
 def assert_shape_match(shape1, shape2):
   """Ensure the shape1 match the pattern given by shape2.
 
--- a/tensorflow_datasets/core/utils/tf_utils.py
+++ b/tensorflow_datasets/core/utils/tf_utils.py
@@ -17,6 +17,7 @@
 
 import collections
 import contextlib
+import functools
 from typing import Union
 
 import numpy as np
@@ -135,6 +136,7 @@ def is_dtype(value):
   return True
 
 
+@functools.lru_cache()
 def assert_shape_match(shape1, shape2):
   """Ensure the shape1 match the pattern given by shape2.
 
--- a/tensorflow_datasets/core/utils/tf_utils.py
+++ b/tensorflow_datasets/core/utils/tf_utils.py
@@ -136,7 +136,7 @@ def is_dtype(value):
   return True
 
 
-@functools.cache
+@functools.lru_cache
 def assert_shape_match(shape1, shape2):
   """Ensure the shape1 match the pattern given by shape2.
 


tensorflow_datasets/core/features/tensor_feature.py
```diff
@@ -131,19 +132,10 @@ def encode_example(self, example_data):
                        f'For {self}')
 
     np_dtype = np.dtype(self.dtype.as_numpy_dtype)
-    if isinstance(example_data, tf.Tensor):
```
Conchylicultor, line 134: We want to keep this check as this is a common error

tensorflow_datasets/core/features/tensor_feature.py
```diff
@@ -131,19 +132,10 @@ def encode_example(self, example_data):
...
-      raise TypeError(
-          f'Error encoding: {example_data!r}. `_generate_examples` should '
-          'yield `np.array` compatible values, not `tf.Tensor`')
-    if not isinstance(example_data, np.ndarray):
-      example_data = np.array(example_data, dtype=np_dtype)
-    # Ensure the shape and dtype match
-    if example_data.dtype != np_dtype:
```
Conchylicultor, line 141: We want to keep this check too

tensorflow_datasets/core/features/tensor_feature.py
```diff
@@ -119,6 +119,7 @@ def get_serialized_info(self):
       }
     return serialized_spec
 
+  @functools.lru_cache(maxsize=2048)
```
Conchylicultor, line 122: example_data should be different each time. How is it supposed to help ?

Additionally, I don't think example_data is hashable. Does it actually improve the speed ?

tensorflow_datasets/core/features/tensor_feature.py
```diff
@@ -131,19 +132,10 @@ def encode_example(self, example_data):
...
-      example_data = np.array(example_data, dtype=np_dtype)
-    # Ensure the shape and dtype match
-    if example_data.dtype != np_dtype:
-      raise ValueError('Dtype {} do not match {}'.format(
-          example_data.dtype, np_dtype))
-
-    shape = example_data.shape
-    utils.assert_shape_match(shape, self._shape)
+
+    example_data = np.asanyarray(example_data, dtype=np_dtype)
```
Conchylicultor, line 136: `np.asanyarray` might accidentally cast data. It's best to be explicit and only call `asarray` if the data is not already a np.ndarray, as previously

tensorflow_datasets/core/utils/tf_utils.py
```diff
@@ -135,6 +136,7 @@ def is_dtype(value):
   return True
 
 
+@functools.cache
```
Conchylicultor, line 139: lru_cache as we're still supporting python 3.6

tensorflow_datasets/core/example_serializer.py
```diff
@@ -134,13 +134,12 @@ def _item_to_tf_feature(item, tensor_info):
   if v.dtype == np.bool_:
     v = v.astype(int)
 
-  v = v.flatten()  # Convert v into a 1-d array
```
Conchylicultor, line 137: v = v.flat

Conchylicultor (changes requested):
Thank you for your contribution.

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: It does show improvement. Without running deep tests, I think that datasets with a lot of discrete features will benefit most.
Trivial example: one hot encodings have a lot of repeating 0s and 1s.

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: My understanding is that `asarray` and `asanyarray` behave identically in terms of data type casting. The only difference is that `asaanyarray` will let more types (e.g. `matrix`) to be converted without a copy.  This behavior seems okay since we only need a flat iterator.

tensorflow_datasets/core/utils/tf_utils.py
eli-osherovich: done.

tensorflow_datasets/core/example_serializer.py
eli-osherovich: done.

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: I remove this since `Tensors` are converted without any problem to `np` `arrays`. Why do we need this test?

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: Maybe I miss something, but we just converted it to `np.ndarray`. This test condition should always be false. Shouldn't it?

tensorflow_datasets/core/features/tensor_feature.py
Conchylicultor: We do not want to cast the dtype.

If the user pass `np.array([.1, 0.4, .5])` but the expected dtype is int32, we want to raise an error. With this change, the float values will be silently casted to int32, instead of raising an error: `ValuError('Expected int32, got float32')`

tensorflow_datasets/core/features/tensor_feature.py
Conchylicultor: See my above comment

tensorflow_datasets/core/utils/tf_utils.py
```diff
@@ -135,6 +136,7 @@ def is_dtype(value):
   return True
 
 
+@functools.lru_cache
```
Conchylicultor, line 139: Datasets can have GB or TB of examples, this basically force the full dataset to be kept in memory.

tensorflow_datasets/core/features/tensor_feature.py
```diff
@@ -16,12 +16,12 @@
 """Feature connector."""
```
Conchylicultor, line 16: Could you run unittests to make sure this is not breaking current generation ? Especially all the ones in `core/features/`

Conchylicultor (changes requested):

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: In fact, this function has very little correlation with the dataset size. It is directly related to the number of features. 
If we want to be on the safe side, we can limit the cache size.

tensorflow_datasets/core/utils/tf_utils.py
eli-osherovich: In fact, this function has very little correlation with the dataset size. It is directly related to the number of features. 
If we want to be on the safe side, we can limit the cache size.

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: Hmm, but it is exactly what the old code does... it casts data type.

```python
if not isinstance(example_data, np.ndarray):
      example_data = np.array(example_data, dtype=np_dtype)
```

Do I read it incorrectly?

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: It took some time. Now I do see a lot of errors of unhashable types: TensorShape, list, ...
We can either remove caches or use hashable types...

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: See my comment above. Some datasets use unhashable types (e.g. lists for bounding boxes). It is, of course, fixable. The question is whether we want to convert them to hashable or not.

tensorflow_datasets/core/utils/tf_utils.py
eli-osherovich: Unless each element can have a different shape. Is it possible?

tensorflow_datasets/core/features/tensor_feature.py
eli-osherovich: master:
---
11 failed, 2099 passed, 446 skipped, 2617 warnings

PR (with caching removed):
---
========== 19 failed, 2091 passed, 446 skipped, 2617 warnings 


The additional failed tests are expecting failures that are now gone.

Conchylicultor: Our tests are green at master: https://source.cloud.google.com/results/invocations/a220ad4b-b74d-47aa-aadf-c9bda4a658be/targets, so maybe some dependencies are missing from your set-up.

If tests are too long to run, only run the ones in `core/features/` as those should test the feature encoding/decoding.

> It took some time. Now I do see a lot of errors of unhashable types: TensorShape, list, ...
> We can either remove caches or use hashable types...

This means this is gonna break many users which are not using hashable types, so we should revert this change.

Please fix the PR to have tests pass again (at least the ones in `core/features/`).

eli-osherovich: Thanks @Conchylicultor , without the caching this PR has very little value for me. 
I would suggest to standardize some things to make it 
1. more understandable  + fewer surprises
2. more structured 
2. amendable for more optimizations

E.g.,  shapes and bounding boxes must be tuples, elements must be either int64, float, or byte (types supported by the example proto).


If this direction looks good, I can help with the transition.

tensorflow_datasets/core/utils/tf_utils.py
ameya98: I came across this commit accidentally. Please don't assume the number of features are constant throughout the dataset. This assumption would be incorrect in the case of graph datasets, such as [ogbg_molpcba](https://www.tensorflow.org/datasets/catalog/ogbg_molpcba) on TFDS.

Conchylicultor:
I think this PR has good ideas. I personally in favor of splitting this PR in 2 to first submit the non-controvertial changes (the `v.flat`, `asanyarray`), and move the memoize changes in an other PR, which can be tested separately.

So at least users could benefit of some speed benefits

eli-osherovich:
SGTM

tensorflow_datasets/core/features/tensor_feature.py
Conchylicultor: > Hmm, but it is exactly what the old code does... it casts data type silently.

It will cast data if user pass `list[int]` (-> `np.array`), but not if the user pass a `np.array`

PR #3459, opened 2021-09-04
Add zh-CN and zh-HK to Common Voice

rongcuid:
# Update Dataset

* Dataset Name: Common Voice
* Issue Reference: None
  
## Description

Add `zh-CN` and `zh-HK` to languages of common dataset


Conchylicultor (changes requested):
Thank you for your contribution. Please also update the checksums https://www.tensorflow.org/datasets/add_dataset#checksums
by generating the dataset with `--register_checksums`. (see https://www.tensorflow.org/datasets/cli#tfds_build_download_and_prepare_a_dataset to only generate a single config with `--config <name>`)

PR #3532, opened 2021-10-15
Allow underscore in slice value syntax (eg 50_000)

lucasb-eyer:
(I have not added a test because github UI doesn't allow editing multiple files and I didn't have time for the overhead of fork->clone->...etc)


ccl-core:
Thank you @lucasb-eyer for your contribution!

Your change is in place:
https://github.com/tensorflow/datasets/blob/252535edd5e456609a3b402f3ca41556d6b03b18/tensorflow_datasets/core/tfrecords_reader.py#L67

PR #3873, opened 2022-04-07
update urls from pass v2 to v3

yukimasano:
# Update of PASS dataset from v2 to v3

Hi, I've updated the PASS dataset. The updated version has 131 less images, as they were found to go against the PASS policy of not including humans/human parts or other personally identifiable information. 

All data/meta-data is updated and format is kept the same.
Best, Yuki


ccl-core:
Thank you @yukimasano !

PR #3886, opened 2022-04-16
setup.py: librispeech needs apache_beam

albertz:


fineguy:
Thanks for the PR!

I'm closing it though, because librispeech can be prepared without using Beam.

PR #4217, opened 2022-10-12
Update wikipedia.py data storage link

Splend1d:
## Description
link not found with the following code

```
ds = tfds.load('wikipedia/20220620.zh', split='train')
```
The new header for wikipedia seems to be under ```https://dumps.wikimedia.your.org/pub/wikimedia/dumps```
After changing to this link I was able to download the data.


Splend1d (commented):
The following code fails to downlaod
```
ds = tfds.load('wikipedia/20220620.zh', split='train')
```
and is solvable by changing to the base url

pierrot0:
Wikimedia dataset was fixed outside PR.

PR #4592, opened 2022-12-22
Fix a command-line snippet on GCS documentation.

yotarok:
There's an issue in "gcs.md" on the command-line arguments of "gcloud".
I think the PR checklist is not applicable for this kind of small doc fix.

I'm not an expert of Google Cloud SDK, but I confirmed it worked with the fix.


fineguy:
Hello,

Thank you for spotting this issue! 

You're right, I'll merge this PR.

PR #4593, opened 2022-12-22
Add valid_tags.txt to PyPI releases.

bartvm:
Without this file any custom dataset test will fail since 23e77add4a3c8d2cd86575678f8f991f7a571701. This commit added the `test_valid_tags` and `test_valid_tags_with_comments` tests which require this file.


pierrot0:
The change was applied otherwise, but thank you for your contribution!

PR #4830, opened 2023-03-27
Fixed the broken link in acronym_identification.md

tiruk007:
Fixed the broken link for code in line 5.


fineguy:
Hello,

Thank you for spotting this problem! 

Huggingface self hosts dataset source code instead of using GitHub, e.g. https://huggingface.co/datasets/acronym_identification/blob/main/acronym_identification.py 

I'll do a bulk update for all such datasets, so closing this PR for now.

Issue #4858, opened 2023-04-12
Unable to import tfds with tensorflow 2.12


rivershah:
/!\ PLEASE INCLUDE THE FULL STACKTRACE AND CODE SNIPPET

**Short description**
Unable to import tfds

**Environment information**
* Operating System: <os>
* Python version: <version>
* `tensorflow-datasets`/`tfds-nightly` version: tensorflow-datasets==4.9.1
* `tensorflow`/`tf-nightly` version: tensorflow==2.12.0

* Does the issue still exists with the last `tfds-nightly` package (`pip install --upgrade tfds-nightly`) ?

**Reproduction instructions**

```python
python
Python 3.11.3 (main, Apr  5 2023, 14:15:06) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tensorflow_datasets as tfds
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/tensorflow_datasets/proto/__init__.py", line 20, in <module>
    from waymo_open_dataset import dataset_pb2 as waymo_dataset_pb2
ModuleNotFoundError: No module named 'waymo_open_dataset'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.11/dist-packages/tensorflow_datasets/__init__.py", line 43, in <module>
    import tensorflow_datasets.core.logging as _tfds_logging
  File "/usr/local/lib/python3.11/dist-packages/tensorflow_datasets/core/__init__.py", line 22, in <module>
    from tensorflow_datasets.core import community
  File "/usr/local/lib/python3.11/dist-packages/tensorflow_datasets/core/community/__init__.py", line 18, in <module>
    from tensorflow_datasets.core.community.huggingface_wrapper import mock_builtin_to_use_gfile
  File "/usr/local/lib/python3.11/dist-packages/tensorflow_datasets/core/community/huggingface_wrapper.py", line 31, in <module>
    from tensorflow_datasets.core import dataset_builder
  File "/usr/local/lib/python3.11/dist-packages/tensorflow_datasets/core/dataset_builder.py", line 34, in <module>
    from tensorflow_datasets.core import dataset_info
  File "/usr/local/lib/python3.11/dist-packages/tensorflow_datasets/core/dataset_info.py", line 52, in <module>
    from tensorflow_datasets.core.features import feature as feature_lib
[...]

PR #4865, opened 2023-04-13
Set protobuf version to <=3.20

andreped:
## Description
Suggested fix to resolve issue https://github.com/tensorflow/datasets/issues/4858#issuecomment-1506735950.

I am observing the exact same issues with regular `tensorflow` and setting this restriction resolved the issue.

If this is not a suitable fix, it would be nice to know why as I'm contributing to various projects which have the same issue right now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ REQUIRED_PKGS = [
     'etils[enp,epath]>=0.9.0',
     'numpy',
     'promise',
-    'protobuf>=3.20',
+    'protobuf<=3.20',
     'psutil',
     'requests>=2.19.0',
     'tensorflow-metadata',


andreped:
I noticed that no tests were being ran automatically for a PR, so I'm running the `unittests` CI now in my own fork [here](https://github.com/andreped/datasets/actions/runs/4688415586).

Is there a reason why there is no `pull_request` hook [here](https://github.com/tensorflow/datasets/blob/master/.github/workflows/pytest.yml#L3)?

fineguy:
Hi, thank you for your work!

We published a new release, could you please confirm that it works for you?
`pip install tensorflow_datasets==4.9.2 --upgrade`

> If this is not a suitable fix, it would be nice to know why as I'm contributing to various projects which have the same issue right now.
The `tensorflow_datasets` package is not working properly because the versions of the Python proto files that it generates do not match the version of the `protobuf` library that is installed on the system. Downgrading `protobuf` is indeed a way to solve this issue, but we want to have support for the newer versions as well, hence we updated the proto files: https://github.com/tensorflow/datasets/commit/877e481438e7677e28290e826ae732c9a98c45cf
That being said, the other projects might have issues for a different reason. 

> Is there a reason why there is no `pull_request` hook [here](https://github.com/tensorflow/datasets/blob/master/.github/workflows/pytest.yml#L3)?
Good catch, I think we'll add it soon.

PR #4977, opened 2023-06-20
Fix superfluous tf_dtype conversion

jpauwels:
## Description

Now that NumPy DTypes are preferred, remove the line that casts NumPy DTypes to TF DTypes. Otherwise a warning (dtype_utils.py, l.44) gets triggered even when a NumPy DType gets correctly passed.


fineguy:
Hi Johan

I'll check it and merge.

PR #5084, opened 2023-09-13
fix cycle_gan

louwjac:
This update replaces the broken download link with a working alternate. It also disables the "cityscapes" dataset option because it is no longer available.

### DIFF
--- a/tensorflow_datasets/image_classification/cycle_gan.py
+++ b/tensorflow_datasets/image_classification/cycle_gan.py
@@ -41,9 +41,10 @@ _CITATION = """\
 }
 """
 
-_DL_URL = "https://people.eecs.berkeley.edu/~taesung_park/CycleGAN/datasets/"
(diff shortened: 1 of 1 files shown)


fineguy:
Hi @louwjac, thanks you for your contribution! 

I'll be happy to merge this.

lunar00eclipse:
not useful at all, still cannot access to the dataset horse2zebra, although I was already using the latest tensorflow and tensorflow-dataset

louwjac:
> not useful at all, still cannot access to the dataset horse2zebra, although I was already using the latest tensorflow and tensorflow-dataset

This fix has not made it into the python package yet because there has not been another release since the code was merged. The good news is that you can still use it by installing the nightly build while you wait for the next release. Try:

```
pip install --upgrade tfds-nightly
```

Woods1970:
still broken usingb4.9.2+nightly

PR #5109, opened 2023-10-14
WAR bad array_record wheel on arm64

nouiz:
This WAR a bug in array_record:
https://github.com/google/array_record/issues/71

Mostly, all old pip wheel only work on x86_64, but get installed on all platform as the wheel is marked as platform independent.

A fix was merged in array_record, but no new wheel is made yet.
This PR make sure that we will crash the installation until a new release is done on all non-x86_64 platform. When the release is done, it will make sure to not install older version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,10 @@ DOCLINES = __doc__.split('\n')
 
 REQUIRED_PKGS = [
     'absl-py',
-    'array_record',
+# Min version of 0.5.0 as old array_record wheel are bugged on all
+# platform except 'x86_64'. See
+# https://github.com/google/array_record/issues/71
+    'array_record>=0.5.0',
     'click',
     'dm-tree',
     'etils[enp,epath,etree]>=0.9.0',


fineguy:
Hi @nouiz, thanks for your contribution! 

I see that there's `array-record 0.5.0`: https://pypi.org/project/array-record/#history
Could you please confirm whether it has fixed your issue?

nouiz:
The release of array_record has a fix, but this doesn't fix the old packages that still install on ARM while they don't work on ARM.
So this PR is still needed.

What would remove the need forthis PR is if array_record publish an ARM container, or remove all old wheel. But they didn't.
Merging this PR will be much quicker than getting ARM container for array_record.

nouiz:
We could also bump the min version for all platform to 0.5 if you prefer. This will fix the issue I'm targetting.

fineguy:
Yes, I think bumping the min version is better. Could you please update your PR and I'll merge it?

Issue #2188, opened 2020-07-13
Corrupt files in the `dogs_vs_cats` dataset


tomerh2001:
**Short description**
I encountered this bug during my TensorFlow certification exam, when trying to work with images from the dataset you constantly get the message `Corrupt JPEG data: 228 extraneous bytes before marker 0xd9` again and again, and it takes forever to iterate over the data once with that, I couldn't complete my exam because of that.

**Environment information**
* Operating System: window 10
* Python version: 3.7.4
* `tensorflow-datasets`/`tfds-nightly` version: `tensorflow-datasets` version 3.1.0
* `tensorflow`/`tensorflow-gpu`/`tf-nightly`/`tf-nightly-gpu` version: `tensorflow-gpu` version 2.2.0

**Reproduction instructions**
A very simple way to reproduce the bug:
```
dataset_name = 'cats_vs_dogs'
dataset, info = tfds.load(name=dataset_name, 
                          split=tfds.Split.TRAIN,
                          with_info=True)

for i in dataset:
    print(i)
```

**Expected behavior**
I except to be able to iterate over all the images without getting errors and without it taking forever to complete a single iteration.

vijayphoenix:
@tomergt45 I am unable to reproduce the bug.

As far as I can see, all corrupt images are removed already.

https://github.com/tensorflow/datasets/blob/921c0f86b8eeba863ce0af6523f34ac75d3d7529/tensorflow_datasets/image_classification/cats_vs_dogs.py#L104

And printing is an I/O operation, so time it will a lot of time to print of array of 20000+ images.

tomergt45:
The problem happens when you just iterate over the data (without printing):
```
>>> import tensorflow_datasets as tfds
>>> tfds.__version__
'3.1.0'
>>> dataset = tfds.load('cats_vs_dogs')
>>> dataset = [i for i in dataset['train']]
Corrupt JPEG data: 214 extraneous bytes before marker 0xd9
Corrupt JPEG data: 228 extraneous bytes before marker 0xd9
Corrupt JPEG data: 396 extraneous bytes before marker 0xd9
Corrupt JPEG data: 65 extraneous bytes before marker 0xd9
Corrupt JPEG data: 1403 extraneous bytes before marker 0xd9
Corrupt JPEG data: 128 extraneous bytes before marker 0xd9
Warning: unknown JFIF revision number 0.00
Corrupt JPEG data: 252 extraneous bytes before marker 0xd9
[...]

PR #5124, opened 2023-10-31
cats_vs_dogs: recoding image on generate samples to fix corrupted images

SirLefti:
This is a proposal to fix https://github.com/tensorflow/datasets/issues/2188.
@eclarson suggested to use `decode_image` and `recode_jpeg` in https://github.com/tensorflow/datasets/issues/2188#issuecomment-1403126938 to rewrite the images, which fixes them (12 out of 25000 are corrupted, one additional has a defective JFIF revision number). Kudos to him, I am just putting it into a PR.

I added those lines into the `_generate_examples` function. It is executed only once when extracting the archive, which means:
* the additional computation time for recoding the images is only needed once
* users might have to delete their local copy of the dataset to generate it again with the fixes
* the errors are still displayed, but when generating the dataset instead of during training or inference

Note: I am not familiar with contributing to python projects like tensorflow, but tried to follow the contribution guide. Please let me know if I am missing something.

### DIFF
--- a/tensorflow_datasets/image_classification/cats_vs_dogs.py
+++ b/tensorflow_datasets/image_classification/cats_vs_dogs.py
@@ -97,13 +97,20 @@ class CatsVsDogs(tfds.core.GeneratorBasedBuilder):
       if tf.compat.as_bytes("JFIF") not in fobj.peek(10):
         num_skipped += 1
         continue
+
+      # some images caused 'Corrupt JPEG data...' messages during training or any other iteration
+      # recoding them once fixes the issue (discussion: https://github.com/tensorflow/datasets/issues/2188)
+      # those messages are now displayed when generating the dataset instead
       img_data = fobj.read()
       img_tensor = tf.image.decode_image(img_data)
       img_recoded = tf.io.encode_jpeg(img_tensor)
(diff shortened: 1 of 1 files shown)


fineguy:
Hi @SirLefti, thanks for your contribution! 

A couple thoughts:

- Are you sure that `tf.image.decode_image` doesn't fail for any images? Maybe you could wrap it in try..catch.
- It's a bit hard to understands what's going on without reading the comment that you reference. Could you please comment your code to make it clear? Maybe you could move this image processing to a separate function.

SirLefti:
Documentation added, thank you for pointing that out.
`decode_image` should never fail if the archive itself is not corrupted. Non-jpeg and corrupted files are skipped in the previous lines. One might add a try..catch, but only to raise it with a custom message again, because skipping the sample would result in a different dataset (like raising the Error in the end of the function if the number of known "really corrupted" samples does not match).

fineguy:
Thanks, almost ready for merging! 

Could you please update the dataset version according to https://www.tensorflow.org/datasets/datasets_versioning#semantic? I assume that the data doesn't change, so probably `4.0.1`?

PR #5297, opened 2024-03-04
[Fix] Full update of data_dir in DatasetInfo with update_data_dir()

youliangtan:
## Description

The previous `DatasetInfo.update_data_dir()` only updates the datadir in each split, but not its DatasetIdentity, making the update incomplete. The datadir property doesn't have a setter too, thus, this adds a single liner to update the identity when the update_data_dir() is called.

### DIFF
--- a/tensorflow_datasets/core/dataset_info.py
+++ b/tensorflow_datasets/core/dataset_info.py
@@ -538,7 +538,7 @@ class DatasetInfo(object):
         self.as_proto.splits.add().CopyFrom(split_info.to_proto())
 
   def update_data_dir(self, data_dir: str) -> None:
-    """Updates the data dir for each split."""
+    """Updates the data dir and data dir for each split."""
     new_split_infos = []
     for split_info in self._splits.values():
       if isinstance(split_info, splits_lib.MultiSplitInfo):
@@ -553,6 +553,7 @@ class DatasetInfo(object):
       new_split_info = split_info.replace(filename_template=filename_template)
       new_split_infos.append(new_split_info)
     self.set_splits(splits_lib.SplitDict(new_split_infos))
+    self._identity.data_dir = target_dir
 
   @property
   def initialized(self) -> bool:


fineguy:
Thanks @youliangtan for your interest in TFDS!

Do you have an example when the update is incomplete causing issues? 

We use this function only once in: https://github.com/tensorflow/datasets/blob/47f7a2e653d1f8c5ae9e41749dfbbf20312c422c/tensorflow_datasets/core/dataset_builder.py#L702-L703 to fix `data_dir` in splits, but the identity should already have the right `data_dir` after dataset builder was initialized: https://github.com/tensorflow/datasets/blob/47f7a2e653d1f8c5ae9e41749dfbbf20312c422c/tensorflow_datasets/core/dataset_builder.py#L281-L284

youliangtan:
Thanks for the reply, I'm using this method to update a new data_dir, after `DatasetInfo` is constructed

The code example is [here](https://github.com/rail-berkeley/oxe_envlogger/blob/7c7c71b768c5f551a73d5471f055c16c9a137eae/reshard_rlds.py#L40-L44). This script is mainly to merge or reshard the rlds data, maybe there's  better way of doing this too. :man_shrugging:

fineguy:
I think `DatasetInfo.update_data_dir` should be a private method and only used internally. 

For your code I suggest to consider these options : 
- Use other tools (for example https://github.com/google/seqio) to combine multiple datasets.
- Dynamically define a dataset builder that gets examples from multiple datasets in `_generate_examples()`.
- Construct your `dataset_info` based using https://github.com/tensorflow/datasets/blob/47f7a2e653d1f8c5ae9e41749dfbbf20312c422c/tensorflow_datasets/core/dataset_info.py#L101

fineguy:
I agree, that would be a nice change.

tomvdw:
Have you looked into builder_from_directories? (see https://github.com/tensorflow/datasets/blob/303814edf2437e607574d6c128d85ac6cfa6c30a/tensorflow_datasets/core/read_only_builder.py#L152) This merges multiple datasets.

PR #5634, opened 2024-10-03
adding explicit repr to PythonDataSource

vyeevani:
Need to avoid explicit serialization of flatExamples in the repr


vyeevani:
https://github.com/tensorflow/datasets/issues/5633

PR #11070, opened 2025-06-16
Fix shard computation in `NoShuffleBeamWriter` with overlapping split names

lgeiger:
Currently `NoShuffleBeamWriter` would compute the wrong number of shards for cases where splits have overlapping names.

E.g. given a dataset with two splits `foo` and `foo_bar` the generated tfrecord files would look something like this:
```
builder-foo.tfrecord-00000-of-00003
builder-foo.tfrecord-00001-of-00003
builder-foo.tfrecord-00002-of-00003
builder-foo_bar.tfrecord-00000-of-00002
builder-foo_bar.tfrecord-00001-of-00002
````
The current regex for the `foo` split would be `builder-foo*` which also matches the files of the `foo_bar` split. This results in the number of shards for `foo` to also include shards from the other split. This fixes it by changing the regex to `builder-foo.*`.

### DIFF
--- a/tensorflow_datasets/core/writer_test.py
+++ b/tensorflow_datasets/core/writer_test.py
@@ -592,39 +592,40 @@ class NoShuffleBeamWriterTest(parameterized.TestCase):
 
     with tempfile.TemporaryDirectory() as tmp_dir:
       tmp_dir = epath.Path(tmp_dir)
-      filename_template = naming.ShardedFileTemplate(
-          dataset_name='foo',
-          split='train',
-          filetype_suffix=file_format.file_suffix,
-          data_dir=tmp_dir,
-      )
-      writer = writer_lib.NoShuffleBeamWriter(
-          serializer=testing.DummySerializer('dummy specs'),
-          filename_template=filename_template,
-          file_format=file_format,
-      )
-      to_write = [(i, str(i).encode('utf-8')) for i in range(10)]
-      # Here we need to disable type check as `beam.Create` is not capable of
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by lgeiger** [comment]
@tomvdw Do you mind having a look at this?

**Comment by fineguy** [comment]
Hi @lgeiger, thanks for the PR! 

Could you please add tests? Then I'll help you merge it.

**Comment by lgeiger** [comment]
@fineguy I extended the existing test in 6f96eaa26bcc431122584f5bfe9675538ca4fb55 which fails on main but is fixed with this PT. Best to reviewed with "hide whitespace" diff.

File: tensorflow_datasets/core/writer_test.py [hunk 1]
```diff
@@ -592,39 +592,40 @@ def test_write_beam(self, file_format: file_adapters.FileFormat):
...
+        )
+        with beam.Pipeline(options=options, runner=_get_runner()) as pipeline:
+
+          @beam.ptransform_fn
+          def _build_pcollection(pipeline):
+            pcollection = pipeline | 'Start' >> beam.Create(to_write)
+            return writer.write_from_pcollection(pcollection)
+
+          _ = pipeline | 'test' >> _build_pcollection()  # pylint: disable=no-value-for-parameter
+        shard_lengths, total_size = writer.finalize()
```
fineguy, line 621: You should run the pipeline with both writers first, and only then check `.finalize()` for each one of them. Otherwise it's not clear whether the test is passing because everything works properly, or because some shards didn't exist at the time when you called `.finalize()`.

